Issue #258, opened 2024-12-16
Can't pipe text selection from helix editor into `q chat`


Rahel-A:
### Checks

- [X] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [X] I have run `q doctor` in the affected terminal session
- [X] I have run `q restart` and replicated the issue again

### Operating system

Ubuntu 22.04 (Amazon image)

### Expected behaviour

I can pipe some text into `q chat` through the CLI (`echo "say hi"|q chat`), but I can't do the same through the helix editor.
See this for how `smartcat` works:
https://github.com/efugier/smartcat/blob/main/assets/qatohtml.gif

### Actual behaviour

For this text selection: "say hi", no output text gets saved into where the selection was.

### Steps to reproduce

Open a new buffer in helix editor.
Type "say hi"
Select text "say hi" and then press ":" to enter command mode.
Type "pipe q chat", the editor freezes for a few seconds, and "say hi" prompt disappears.

grant0417:
The PR has been merged, thanks for the contribution.

PR #259, opened 2024-12-16
fix: accept text through pipes in chat

Rahel-A:
For the `q chat` command, also allow piping text into the q chat command.

*Issue #, if available:*
#258 
*Description of changes:*
Similarly to `smartcat`, check if we're piping text, if so, then the chat input will be fetched from the stdin.

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -20,9 +20,9 @@ use crossterm::style::{
 };
 use crossterm::{
     cursor,
-    style,
+    execute,
     queue,
-    execute
+    style,
 };
 use eyre::{
     Result,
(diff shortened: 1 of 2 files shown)


Rahel-A:
Tested through CLI:
```
echo "say hi" | rustup run stable-x86_64-unknown-linux-gnu cargo run -- chat
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.46s
     Running `/home/ANT.AMAZON.COM/rahelgah/work/amazon-q-developer-cli/target/debug/q_cli chat`

Hi! How can I help you with your development needs today?
```

Rahel-A:
[27e104f](https://github.com/aws/amazon-q-developer-cli/pull/259/commits/27e104f0b4473d7802cce2068ea4bcdd4eedbfa7) works but at the end there's still some more ANSI characters.

grant0417:
Also can you edit the description the PR to include:

> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Rahel-A:
Do you have some default prompts auto embedded for the first question?
This is the prompt used in `smartcat`:
```
                You are an extremely skilled programmer with a keen eye for detail and an emphasis on readable code. \
                You have been tasked with acting as a smart version of the cat unix program. You take text and a prompt in and write text out. \
                For that reason, it is of crucial importance to just write the desired output. Do not under any circumstance write any comment or thought \
                as your output will be piped into other programs. Do not write the markdown delimiters for code as well. \
                Sometimes you will be asked to implement or extend some input code. Same thing goes here, write only what was asked because what you write will \
                be directly added to the user's editor. \
                Never ever write ``` around the code. \
                Make sure to keep the indentation and formatting. \
```

grant0417 (approved):

PR #391, opened 2025-01-08
feat: Add Linux updating for AppImage builds

brandonskiser:
*Description of changes:*
- Adding the ability for Linux AppImage builds to update itself. This is done by distributing metadata with the release (referred to as `BundleMetadata`) so that desktop builds can distinguish between AppImage and deb.


crates/fig_install/src/linux.rs
```diff
@@ -165,6 +186,96 @@ pub(crate) async fn update(
...
+
+    debug!(?download_path, ?current_appimage_path, "Replacing the current AppImage");
+    ctx.fs().rename(&download_path, &current_appimage_path).await?;
+    debug!("Successfully swapped the AppImage");
+
+    tx.send(UpdateStatus::Message("Relaunching...".into())).await.ok();
+
+    let pid = ctx.process_info().current_pid().as_u32().to_string();
+    std::process::Command::new(current_appimage_path)
+        .args(["--kill-old-pid", &pid])
```
grant0417, line 260: Why do we need `--kill-old-pid` if we also exit at the end?

grant0417 (approved):

crates/fig_install/src/linux.rs
brandonskiser: This was intended to catch an edge case where the new app launches before the old one finishes exiting. In my testing it looks like it never actually happens where `--kill-old-pid` is required so I'll probably just remove it. Worst case, the user just has to relaunch the AppImage.

grant0417: Its fine to keep, just wanted to check if there was anything I was missing.

grant0417:
Not that this PR needs it, but some more testing around update logic would be very nice :D

PR #392, opened 2025-01-08
fix: chat infinite loop and missing lines

### LINKED ISSUE #390: `q chat` issues
The modifications in #259 and #293 must have caused a regression with the interactive TUI where lines are trimmed:

Screenshot 2025-01-08 at 12 24 09 PM

Additionally piping into q chat causes an infinite loop:

Screenshot 2025-01-08 at 12 25 05 PM

We need to resolve these before the 1.6.0 release.

grant0417:
Fixes #390

*Issue #, if available:*

*Description of changes:*


crates/q_cli/src/cli/chat/parse.rs
```diff
@@ -499,13 +502,17 @@ fn queue_newline_or_advance<'a, 'b>(
     state: &'b mut ParseState,
     width: usize,
 ) -> Result<(), ErrMode<Error<'a>>> {
-    if state.column > 0 && state.column + width > state.terminal_width {
-        state.column = width;
-        queue(&mut o, style::Print('\n'))?;
-    } else {
-        state.column += width;
+    if let Some(terminal_width) = state.terminal_width {
```
chaynabors, line 505: Wait is this right? Shouldn't we use a fallback of like 80 or 256 if it's not set or no?

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -259,32 +264,26 @@ You can include additional context by adding the following to your prompt:
...
                             )?;
                         }
 
-                        if !state.citations.is_empty() {
-                            execute!(output, Print("\n"))?;
-                        }
-                    }
-
-                    if let (Some(conversation_id), Some(message_id)) = (&conversation_id, &message_id) {
-                        fig_telemetry::send_chat_added_message(conversation_id.to_owned(), message_id.to_owned()).await;
```
chaynabors, line 287: Did we remove telem or move it somewhere?

grant0417: Moved

chaynabors (approved):

crates/q_cli/src/cli/chat/parse.rs
grant0417: This is a pretty rare case where we can not query the terminal width which we previously failed in. I think its fine if the [`window_size`](https://docs.rs/crossterm/latest/crossterm/terminal/fn.window_size.html) function failed then we don't make any assumptions here since it may render way worse.

PR #410, opened 2025-01-10
feat: Add updating for Linux AppImage

brandonskiser:
*Description of changes:*
- Adding the ability for Linux AppImage builds to update itself. This is done by distributing metadata with the release (referred to as `BundleMetadata`) so that desktop builds can distinguish between AppImage and deb.
- Adding a new crate `fig_test_utils` so we can start adding more robust stubbing/mocking in our tests in a shared place.


crates/fig_test_utils/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+pub mod server;
+
+use hex::ToHex;
+pub use http;
+pub use server::*;
+
+/// Encodes the byte slice as a lower-case sha256 hex string.
+pub fn sha256(data: &[u8]) -> String {
```
grant0417, line 8: Any reason this in this create? Seems like something that is pretty generic and we already have the `ring` and `hex` dependencies in `fig_install` where this is used.

grant0417 (approved):
LGTM!

crates/fig_test_utils/src/lib.rs
brandonskiser: It was just a personal preference for readability sake to write `sha256` instead of `digest(&ring::digest::SHA256, data).encode_hex()`, I can remove it in a followup since I didn't add any doc comment to the TestServer either

grant0417: No need, just asking.

PR #439, opened 2025-01-14
feat: finish multiplexer protocol and functionality

grant0417:
*Description of changes:*

This PR is relatively large as it is a lot of work that has happened over the past several months. These are the rust changes that have been cherry-picked from https://github.com/aws/amazon-q-developer-cli/tree/grant/packet.

Changes on this branch:
- Add `Packet` to `mux.proto` allowing for compression and packet versioning of wire protocol
- Add `Base64LineCodec` to `fig_ipc`, this is the wire encoding/decodeing of the `mux.proto`
- Move `fig_remote_ipc` to use Uuid for session ids, this is how the majority of the code 
already functioned and cleans up the implementation.
- Remove `AccountInfo`, `StartExchangeCredentials`, and `ConfirmExchangeCredentials` from `remote.proto` these were unused and increased the size of the multiplexer implementation.
- Remove `gamma` features from crates, this was unused
- Cleaned up unused clients in `fig_request`
- Add `Q_TERM_DISABLED` setting to disable qterm launching.
- Misc other dependency clean up that happened to get added to that branch


crates/fig_test_utils/src/lib.rs
```diff
@@ -1,4 +1,10 @@
 pub mod server;
 
+use hex::ToHex;
 pub use http;
 pub use server::*;
+
+/// Encodes the byte slice as a lower-case sha256 hex string.
+pub fn sha256(data: &[u8]) -> String {
```
brandonskiser, line 8: Can you rebase main? This looks like it reverts #440

crates/fig_ipc/src/codec/base64line.rs
```diff
@@ -0,0 +1,186 @@
...
+use tokio_util::codec::{
+    AnyDelimiterCodec,
+    AnyDelimiterCodecError,
+    Decoder,
+    Encoder,
+};
+
+#[derive(Debug, Clone)]
+pub struct Base64LineCodec<T: Message> {
+    line_delinited: AnyDelimiterCodec,
```
brandonskiser, line 20: nit - `line_delimited`

crates/fig_proto/src/mux.rs
```diff
@@ -0,0 +1,107 @@
...
+use rand::{
+    Rng,
+    RngCore,
+};
+use thiserror::Error;
+
+pub use crate::proto::mux::*;
+
+// The current version of the packet, to track backwards incompatible changes
+pub const PACKET_VERSION: u32 = 0;
```
brandonskiser, line 14: just wondering, could this be shortened to just `u8`? Although not sure how chatty/long the messages are in comparison so probably doesn't matter

crates/fig_remote_ipc/src/figterm.rs
```diff
@@ -175,7 +178,9 @@ impl From<InterceptMode> for bool {
 
 #[derive(Debug, Serialize)]
 pub struct FigtermSession {
-    pub id: FigtermSessionId,
+    pub id: Uuid,
+    // This is the ID provided by the figterm session
+    pub external_id: String,
```
brandonskiser, line 183: what's the difference between id and external_id? don't really understand the comment

crates/q_cli/src/util/pid_file.rs
```diff
@@ -0,0 +1,162 @@
...
+    debug!(%pid, %exists, "Checked if process exists");
+    exists
+}
+
+#[instrument(level = "debug")]
+async fn kill_process(pid: i32) -> Result<()> {
+    if !process_exists(pid) {
+        error!(%pid, "Process not found");
+        return Err(Error::new(
+            ErrorKind::AlreadyExists,
```
brandonskiser, line 128: shouldn't this be `NotExists`?

crates/q_cli/src/util/pid_file.rs
```diff
@@ -0,0 +1,162 @@
...
+            Ok(lock) => lock,
+            Err((file, err)) => {
+                debug!(%err, "Failed to acquire lock, trying to handle existing process");
+
+                // Read existing PID
+                match read_to_string(&pid_path).await {
+                    Ok(content) => match content.trim().parse::<i32>() {
+                        Ok(pid) => {
+                            debug!(%pid, "Found existing process ID");
+                            if let Err(err) = kill_process(pid).await {
```
brandonskiser, line 64: If we kill whatever is holding the lock, then maybe add a doc comment  on `PidLock` explaining the behavior since it seems a bit counterintuitive imo

crates/fig_remote_ipc/src/figterm.rs
grant0417: I'll just remove `external_id`, it is client provided and thus could spoof another id if they knew another client. I think its best if we just dont trust/include it.

crates/fig_test_utils/src/lib.rs
grant0417: Not sure why it reverted it, I rebased but its the same. I tried to manually pick up the changes from main.

brandonskiser (approved):

PR #572, opened 2025-02-10
fix: validate FileContext limits

grant0417:
*Description of changes:*
This PR adds validation of the max length of the FileContext fields.

### DIFF
--- a/crates/fig_api_client/src/clients/client.rs
+++ b/crates/fig_api_client/src/clients/client.rs
@@ -43,6 +43,11 @@ use crate::{
     Error,
 };
 
+// Limits for FileContext
+pub const FILE_CONTEXT_LEFT_FILE_CONTENT_MAX_LEN: usize = 10240;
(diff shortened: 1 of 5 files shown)


crates/fig_api_client/src/clients/client.rs
```diff
@@ -97,6 +102,33 @@ impl Client {
     }
 
     pub async fn generate_recommendations(&self, input: RecommendationsInput) -> Result<RecommendationsOutput, Error> {
+        let validate_field_length = |field: &'static str, value: &str, max_length: usize| -> Result<(), Error> {
+            if value.len() > max_length {
+                return Err(Error::LengthValidation {
```
leigaol, line 107: It is recommended to just truncate the string by max_length in this case

grant0417: Fixed.

leigaol (approved):

PR #677, opened 2025-02-25
feat: add /clear command, fix use_aws parameters bug, add tilde expansion, minor fixes

brandonskiser:
*Description of changes:*
- Adding the `/clear` command to clear conversation history. This is useful in the case where the state gets in a bad state where only InternalServerExceptions are created. This is done by looping in the main `prompt_and_send_request` function which caused a larger diff than usual.
- Add tilde expansion for path handling in `fs_read` and `fs_write`
- Fix a bug with `use_aws` where parameters are sometimes returned capitalized by the model. In this case, we always lowercase parameters.
- Don't prompt for consent on readonly commands. This is currently for all `fs_read` and `use_aws` tools.
- Other minor fixes.

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -255,7 +255,7 @@ Hi, I'm <g>Amazon Q</g>. Ask me anything.
                             style::SetAttribute(Attribute::Bold),
                             style::SetForegroundColor(Color::Red),
                             style::Print(format!("Amazon Q is having trouble responding right now: {:?}\n", e)),
-                            style::Print("Please try again later.\n"),
+                            style::Print("Please try again later.\n\n"),
                             style::SetForegroundColor(Color::Reset),
                             style::SetAttribute(Attribute::Reset),
                         )?;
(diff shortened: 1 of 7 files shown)


crates/q_cli/src/cli/chat/tools/mod.rs
```diff
@@ -67,6 +67,16 @@ impl Tool {
         }
     }
 
+    /// Whether or not the tool should prompt the user for consent before [Self::invoke] is called.
+    pub fn requires_consent(&self, _ctx: &Context) -> bool {
+        match self {
+            Tool::FsRead(_) => false,
```
dingfeli, line 73: Do we want to just wild cards the ones that are true.

dingfeli (approved):

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -420,314 +420,329 @@ Hi, I'm <g>Amazon Q</g>. Ask me anything.
...
+                            Some(line) => line,
+                            None => return Ok(None),
+                        }
+                    },
+                },
+            };
+
+            match user_input.trim() {
+                "exit" | "quit" => return Ok(None),
+                "/clear" => {
```
dingfeli, line 589: If I understand this correctly this returns nothing but ends the current loop which causes it to loop again. 
Would a recursive call be simpler here? Or was there any lifetime issues with that.

PR #764, opened 2025-03-07
✨ feat(cli): Add append functionality to fs_write tool

jsamuel1:
***Description of changes:***

This commit enhances the fs_write tool by adding a new "append" command that allows users to append content to the end of files without having to read and rewrite the entire file.

## Features
- 📝 Add new `Append` variant to the `FsWrite` enum
- 🔄 Implement functionality to append content to existing files
- 🆕 Auto-create files when appending to non-existent files
- 📁 Auto-create parent directories when needed
- 🧪 Add comprehensive test coverage for the new functionality

## Implementation Details
- Added validation to ensure path and content are not empty
- Enhanced user feedback during append operations
- Implemented syntax highlighting for appended content
- Maintained consistent code patterns with existing commands
- Added proper newline handling to ensure clean appends
- Updated tool_index.json schema to expose the new command

## Use Cases
- Adding new lines to configuration files
- Appending log entries
- Adding new content to the end of text files
- Creating files incrementally

This enhancement improves the user experience by providing a more convenient way to add content to files without having to read the entire file first and then use the "create" command to overwrite it.

### DIFF
--- a/crates/q_cli/src/cli/chat/tools/fs_write.rs
+++ b/crates/q_cli/src/cli/chat/tools/fs_write.rs
@@ -45,6 +45,11 @@ pub enum FsWrite {
         insert_line: usize,
         new_str: String,
     },
+    #[serde(rename = "append")]
+    Append {
+        path: String,
+        content: String,
+    },
 }
 
 impl FsWrite {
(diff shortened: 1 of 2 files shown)


crates/q_cli/src/cli/chat/tools/fs_write.rs
```diff
@@ -121,6 +123,44 @@ impl FsWrite {
...
+                    style::Print("Appending to: "),
+                    style::SetForegroundColor(Color::Green),
+                    style::Print(format_path(cwd, &path)),
+                    style::ResetColor,
+                    style::Print("\n"),
+                )?;
+
+                // Read existing content if file exists
+                let mut file_content = if fs.exists(&path) {
+                    fs.read_to_string(&path).await.unwrap_or_default()
```
brandonskiser, line 147: Can we return the error here if we fail reading the file for some reason? Rather append stick to what it's intended to do instead of potentially replacing all of the contents.

brandonskiser (changes requested):
Thanks for the contribution! One small request but otherwise lgtm.

crates/q_cli/src/cli/chat/tools/fs_write.rs
jsamuel1: Sure.  
Tool definition will need to change too for that, as I let the tool know it would create if it doesn't exist.

jsamuel1: Done.

PR #765, opened 2025-03-07
feat: add Compact command for compressing conversation history

klxdev:
*Issue #, if available:*
N/A

*Description of changes:*
## Overview
The Compact command (/compact) is a new feature for Amazon Q Developer CLI that intelligently compresses conversation history while preserving context. This feature addresses the challenge of managing long-running conversations that can become unwieldy 
and hit system limits.

## Problem Statement
As conversations with Amazon Q grow longer, they can:
• Reach token limits that degrade response quality
• Slow down response times due to processing large context windows
• Make it difficult to focus on current tasks when earlier parts of the conversation are no longer relevant

## Solution
The Compact command provides a streamlined way to compress conversation history while maintaining essential context through an intelligent summarization process.

## How It Works
1. When a user types /compact, the CLI initiates a conversation compression workflow
2. The system automatically generates a prompt requesting a concise summary of the conversation
3. This summary follows a structured format:
   • **Original User Ask**: Preserves the initial request that started the conversation
   • **Completed Items**: Documents what has been accomplished, including modified resources and important context
   • **Todo Items**: Lists remaining tasks with required details and next steps
4. After receiving the summary, the system:
   • Retains only the last two messages in the conversation history
   • Replaces the extensive conversation history with the concise summary
   • Maintains continuity for ongoing tasks

## Technical Implementation
• Added a new Compact variant to the Command enum
• Implemented the compact_conversation_history() method in ConversationState that intelligently prunes the conversation history
• Added comprehensive test coverage to ensure reliability
• Integrated with the existing conversation state management system

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -90,6 +90,7 @@ const WELCOME_TEXT: &str = color_print::cstr! {"
 • Help me understand my git status
 
 <em>/acceptall</em>    <black!>Toggles acceptance prompting for the session.</black!>
+<em>/compact</em>      <black!>Compacts conversation history with summary</black!>
 <em>/help</em>         <black!>Show the help dialogue</black!>
 <em>/quit</em>         <black!>Quit the application</black!>
 
@@ -101,6 +102,7 @@ const HELP_TEXT: &str = color_print::cstr! {"
 <magenta,em>q</magenta,em> (Amazon Q Chat)
 
 <em>/clear</em>        <black!>Clear the conversation history</black!>
+<em>/compact</em>      <black!>Compacts conversation history with summary</black!>
 <em>/acceptall</em>    <black!>Toggles acceptance prompting for the session.</black!>
 <em>/help</em>         <black!>Show this help dialogue</black!>
 <em>/quit</em>         <black!>Quit the application</black!>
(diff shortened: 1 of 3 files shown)


crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -741,6 +780,19 @@ where
                 Err(err) => return Err(err.into()),
             }
 
+            // If this was a response to a /compact command, compact the conversation history
+            if let Some(Command::Compact) = last_command {
```
brandonskiser, line 784: This code is never executed since `self.conversation_state.next_message` is always `None`, so the history is actually never compacted. You probably need to pass this context as an optional parameter to this state.

brandonskiser (changes requested):
Thanks for the contribution! Can you also run `cargo +nightly fmt`?

chaynabors:
Kris, I would love to accept this and I love the enthusiasm to contribute to the project.

Unfortunately, the quality of the code generated by the model isn't sufficient for production and requires a human with an understanding of the codebase in the loop. The CLI codebase is quite difficult to onboard onto, and uses Rust which changes often enough that the training data our models use is often out of date, incomplete, or simply incorrect.

Additionally, we should take the time to have a discussion if features like /compact should be on our roadmap. QDC uses a subscription based licensing model so a feature which minimizes the quantity of input tokens may not be sufficiently valuable for our customers as of today. Feel free to reach out to me or Matt Schrage internally if you'd like to get this on the roadmap.

Let me know if you aren't satisfied with this response and I can point out individual issues in a code review. The model would be able to correct these issues if I point them out, but our team doesn't have the bandwidth to review incomplete PR's as a rule of thumb.

PR #799, opened 2025-03-11
feat: improve file diff for fs write

brandonskiser:
*Description of changes:*
- Adding a file diff visualization for `fs_write` that displays line numbers. This implementation needs to be improved for truecolor support - currently, we don't highlight the entire document which can result in lines being highlighted incorrectly, consequently impacting the diff.
- Removing the `clean_file_content` function since it was resulting in a lot of errors with the model trying to use str_replace. Essentially, we were writing different content than the model was generating, which thus required it to constantly reread the file to see what we actually wrote.
- Moving the logic for appending a newline to a separate function shared by all `fs_write` commands.
- Fixed bug with append not including a newline at the end of the file.

Screenshot 2025-03-11 at 2 55 08 PM
Screenshot 2025-03-11 at 2 56 20 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6557,6 +6557,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "shell-color",
  "shlex",
  "similar",
  "spinners",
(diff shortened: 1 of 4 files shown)


chaynabors (approved):

hsbakshi:
Is there a way we could remove the trailing whitespace that the model generates? It results in several checkstyle/lint errors as well as some people have editors that highlight them as errors.

Issue #815, opened 2025-03-13
report "libc.so.6: Permission denied" when over ssh


wilikeit:
curl --proto '=https' --tlsv1.2 -sSf "https://desktop-release.codewhisperer.us-east-1.amazonaws.com/latest/q-x86_64-linux.zip" -o "q.zip"

tu@tu-VirtualBox:/tmp$ ./q/install.sh

./q/install.sh: 72: /usr/lib/x86_64-linux-gnu/libc.so.6: Permission denied

[ERROR] Could not determine glibc version.

[ERROR] This release built for a GNU system with glibc 2.34 or newer, try installing the musl version of the CLI.

my glibc is 2.35 ，why report this problem？

tu@tu-VirtualBox:/tmp$ ldd --version

ldd (Ubuntu GLIBC 2.35-0ubuntu3.1) 2.35

jsamuel1:
Was this an x86 or arm64 linux distribution?

wilikeit:
> 这是 x86 还是 arm64 linux 发行版？您能否显示`ls -l /usr/lib/ | grep gnu``uname -a`
> 
> 如果是 aarch64，[则 #816](https://github.com/aws/amazon-q-developer-cli/pull/816) 中的修复应该可以解决此问题。

ls -l /usr/lib/ | grep gnuuname -a
tu@tu-VirtualBox:/tmp$   #no result


./q/install.sh
``bash
#glibc_version=$("$LIBC_PATH" | sed -n 's/^GNU C Library (.*) stable release version \([0-9]*\)\.\([0-9]*\).*$/\1.\2/p')
        glibc_version=$(ldd --version | awk 'NR==1 {print $NF}')  # ◆ If I change this line, I can get the version. 

    if [ -z "$glibc_version" ]; then
        log_error "Could not determine glibc version."
        return 1
```

PR #816, opened 2025-03-13
✨ Add ARM64 support to Linux zip installer and fix issue #815 for x86 Ubuntu 22.

jsamuel1:
Add support for detecting glibc on ARM64/aarch64 Linux systems by checking for libc.so.6 in the aarch64-linux-gnu directory. This enables the installer to properly validate system requirements on ARM-based Linux distributions.


brandonskiser:
Hey @jsamuel1 thanks so much for the contribution.

I want to step back and try to understand what the exact problem it is we're trying to solve - this PR is adding hundreds of lines of extremely verbose test code for a very marginal benefit. The original issue we wanted to solve was to have an automated way to test installation on a particular Linux distribution. I'd be happy to approve just the portion that improve the libc detection on other systems, and manually verify the installation through a container, but otherwise we need more thought into how this test framework should look and fit into our current build process.

PR #896, opened 2025-03-23
Add MCP SSE Transport Support

jamesward:
Adds the MCP SSE Transport and a test that uses an existing SSE Server implementation.

Note: I'm pretty new to Rust so I'm leaving this as a draft for now to get feedback.


snese:
@mschrage does the PR meet the #1691 requirement?

jamesward:
This PR is very out-of-date and only supports SSE, not the new Streamable HTTP.

chaynabors:
I can leave open given that it's draft, but let me know how you're thinking about this PR going forward. Please reach out over slack if you'd still like to work on this so we can ensure it's on the roadmap, and done with the right context.

jamesward:
I won't be able to get back to this for a few weeks due to some vacation time. But let's keep it open in case someone who has time finds this and wants to reference or refactor it.

dingfeli:
We now support SSE. Closing this PR.

PR #920, opened 2025-03-24
[BUGFIX] History overflow bug fix

mr-lee:
1. Added History Overflow Error Handling:
   • Created a new HistoryOverflowError struct to handle conversation history overflow
   • Modified fix_history() to return a boolean indicating if overflow handling is needed
   • Added a new force_valid_history() method to handle history without clearing it

2. Modified Return Types:
   • Changed as_sendable_conversation_state() to return Result<FigConversationState, HistoryOverflowError>
   • Updated all call sites to handle the new Result type with ? operator

3. Improved History Management:
   • Added logic to find valid starting messages in conversation history
   • Implemented smarter history truncation to preserve context
   • Replaced automatic history clearing with controlled overflow handling

4. Test Improvements:
   • Limited test iterations to avoid overflow errors
   • Updated test assertions to handle Result return types

## Files Modified
• crates/q_cli/src/cli/chat/conversation_state.rs
• crates/q_cli/src/cli/chat/mod.rs


crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -113,6 +117,70 @@ impl ConversationState {
         self.history.clear();
     }
 
+    /// Forces the history to be valid without clearing it
+    /// This is used when the user chooses to continue with a large history
+    pub fn force_valid_history(&mut self) {
```
kkashilk, line 122: `force_valid_history` -> `trim_history_until_valid` ?

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -113,6 +117,70 @@ impl ConversationState {
...
+        // Find the first valid user message to keep
```
kkashilk, line 123: could the definition of `valid` change over time? Might be a good idea to extract this into a function

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -113,6 +117,70 @@ impl ConversationState {
...
+        if let Some(i) = self
+            .history
+            .iter()
+            .enumerate()
```
kkashilk, line 127: Is it possible to mark a user-message as valid while the conversation is ongoing, instead of having to scan it retroactively?

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -113,6 +117,70 @@ impl ConversationState {
...
+                .iter()
+                .enumerate()
+                .skip(1) // Skip the first message which should be from the user
+                .find(|(_, m)| -> bool {
+                    match m {
+                        ChatMessage::UserInputMessage(m) => {
+                            matches!(
+                                m.user_input_message_context.as_ref(),
+                                Some(ctx) if ctx.tool_results.as_ref().is_none_or(|v| v.is_empty())
+                            ) && !m.content.is_empty()
```
kkashilk, line 162: Useful to add a helper `is_valid_user_message` fn here

crates/q_cli/src/cli/chat/conversation_state.rs
kkashilk: For example, we wanna make sure we don't consider this as a valid message - 
https://github.com/aws/amazon-q-developer-cli/blob/main/crates/q_cli/src/cli/chat/mod.rs#L1130

mr-lee:
After discussing with the team - it'll be easier for us to simply increase history length than trying to solve this way.  We'll be closing the PR.

PR #927, opened 2025-03-24
feat(cli) Bash pipe command safety

kiiadi:
## Description

This PR enhances the `execute_bash` function to support safe piped commands without requiring user confirmation. Previously, piped commands were supported but always triggered a confirmation prompt, which interrupted the user experience. This change allows common read-only piped commands like `find | grep` or `ls | grep` to execute directly when all commands in the pipe chain are validated as safe.

## Changes

- Modified the command validation logic to check each command in a pipe chain
- Added support for automatically approving piped commands when all components are in the READONLY_COMMANDS list
- Maintained the security prompt for any piped commands containing potentially unsafe operations
- Improved user experience by reducing unnecessary confirmation prompts

## Testing

- Verified safe piped commands like `find | grep` execute without prompts
- Confirmed unsafe piped commands still trigger the security confirmation
- Tested various combinations of allowed read-only commands in pipes
- Ensured backward compatibility with existing command handling

## Security

This change maintains security by:
- Only auto-approving known read-only commands in pipes
- Validating each command in the pipe chain individually
- Preserving confirmation prompts for any potentially unsafe operations

### DIFF
--- a/crates/q_cli/src/cli/chat/tools/execute_bash.rs
+++ b/crates/q_cli/src/cli/chat/tools/execute_bash.rs
@@ -37,7 +37,7 @@ impl ExecuteBash {
             return true;
         };
 
-        const DANGEROUS_PATTERNS: &[&str] = &["|", "<(", "$(", "`", ">", "&&", "||"];
+        const DANGEROUS_PATTERNS: &[&str] = &["<(", "$(", "`", ">", "&&", "||"];
         if args
             .iter()
             .any(|arg| DANGEROUS_PATTERNS.iter().any(|p| arg.contains(p)))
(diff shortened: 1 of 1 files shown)


mr-lee (approved):

crates/q_cli/src/cli/chat/tools/execute_bash.rs
```diff
@@ -24,7 +24,7 @@ use super::{
 };
 use crate::cli::chat::truncate_safe;
 
-const READONLY_COMMANDS: &[&str] = &["ls", "cat", "echo", "pwd", "which", "head", "tail"];
+const READONLY_COMMANDS: &[&str] = &["ls", "cat", "echo", "pwd", "which", "head", "tail", "find", "grep"];
```
jsamuel1, line 27: @mr-lee What is the best way for us to curate a growing list of READONLY_COMMANDS here?

I foresee this list growing significantly.
eg. I hit another couple today:
wc   # wordcount
jq   # json query
yq # but not if the -i Inplace flag is given!
test
realpath
sleep
df
du

PR #930, opened 2025-03-25
Add file/folder name auto-completion in q chat

kkashilk:
*Issue #, if available:*

1. File/Folder name auto-complete does not currently work in the q-chat window

*Description of changes:*

Extended the `ChatCompleter` to suggest file/folder name auto-completions as fallback

Tested by running - 
`cargo run --bin q_cli -- chat `

See screenshot below for info on what the output looks like -

Screenshot 2025-03-24 at 11 11 29 PM


mr-lee (approved):
Changes look good - just fix the tests and formatter and we can merge!

PR #954, opened 2025-03-27
feat(chat): Add RFC for conversation compaction feature

kkashilk:
Add RFC for conversation compaction feature

### DIFF
--- a/rfcs/0002-compact-command.md
+++ b/rfcs/0002-compact-command.md
@@ -1,0 +1,298 @@
+- Feature Name: compact_command
+- Start Date: 2025-03-26
+
+# Summary
+
+[summary]: #summary
+
+Add a `/compact` command to the Amazon Q Developer CLI chat interface that allows users to compact their conversation history to save context space while preserving the essential information.
+
+# Motivation
+
+[motivation]: #motivation
+
+Large language models like those powering Amazon Q have context limits that restrict how much conversation history can be maintained. As conversations grow longer, they consume more of this limited context space, which can lead to:
+
+1. Degraded response quality as the model loses access to earlier parts of the conversation
+2. Increased token usage and associated costs
+3. Slower response times due to processing larger context windows
+
+The `/compact` command addresses these issues by providing users with a way to reduce the size of their conversation history while preserving the most important information. This allows for longer, more productive conversations without hitting context limits.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Basic Usage
+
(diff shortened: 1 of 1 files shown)


rfcs/0002-compact-command.md
```diff
@@ -0,0 +1,298 @@
...
+- `/compact --conservative [--summary]` - Preserves more detail with lighter summarization
+- `/compact show` - Shows current context usage statistics
+- `/compact [--summary]` - Shows what would be compacted without actually performing the operation
+
+## Understanding Context Usage
+
+When users run `/compact show`, they will see:
+- Current context usage (tokens/percentage)
+- Available context space
+- Estimated conversation turns remaining
```
mschrage, line 57: This seems like potentially overloading the `/compact` command versus just adding another command like `/usage`

rfcs/0002-compact-command.md
```diff
@@ -0,0 +1,298 @@
...
+Amazon Q: I've compacted our conversation history, preserving the key points while reducing the context size by approximately 65%. We can continue our discussion with more context space available.
+```
+
+## Options
+
+The `/compact` command supports optional parameters:
+
+- `/compact` - Default behavior, automatically determines what to compact and how
+- `/compact --aggressive [--summary]` - Maximizes context savings with more aggressive summarization
+- `/compact --conservative [--summary]` - Preserves more detail with lighter summarization
```
mschrage, line 48: Is this additional flexibility (aggressive/conservative) valuable or does it add unnecessary complexity?

rfcs/0002-compact-command.md
kkashilk: Yeah, that makes sense, alternatively we can show this information at the end of each turn

rfcs/0002-compact-command.md
kkashilk: IMO, the flexibility helps to some extent, but we could allow the user to edit the compact-summary for more control. 

This could be done by pushing the summary to a file that the user can read/edit on their own and then call something like `/compact commit`

Wyd?

kkashilk: this flexibility is definitely geared towards power-users, most people might be okay with a single `/compact` command

kkashilk: The way I imagine this to work is, the user has two files open side-by-side, one that shows the entire conversation history, and another that shows the compacted version. 
The user can then decide to copy-paste certain things from the history into the summary, and remove things they don't think is useful/needed anymore

chaynabors:
Closing in favor of new RFC repo. This change was made and is in prod I believe.

Issue #873, opened 2025-03-20
Add tool to create GitHub issues from within q chat


mschrage:
Feature request: Add capability to report issues directly from within q chat

**Proposed functionality:**
- Add a new tool that creates GitHub issues with:
   - Chat context
   - Request IDs
   - Transcript of the conversation

This would streamline the feedback process for users by allowing them to report issues without leaving the CLI interface.

hayemaxi:
Similar to the q issue command, this would open up the pre-filled template link for the user to review and submit rather than directly using creating the issue. This would require the `gh` cli to be locally installed and configured, and the user will likely want to review the information prior to submitting. 
The tool can fill in these details along with a title and description of the issue that the user mentioned (or deduced from chat history), which would be generated by Q.

hayemaxi:
Merged an initial tool for this. Follow up work before we should close this out:
- Record commands, e.g. /context /acceptall
- Record how often context files were used in requests and tool uses (should we record in the transcript what context files were used?)
- [this](https://github.com/aws/amazon-q-developer-cli/pull/974#discussion_r2017302788)

PR #974, opened 2025-03-27
feat(tools): add gh_issue

hayemaxi:
**To be squashed**

*Issue #, if available:*
https://github.com/aws/amazon-q-developer-cli/issues/873

*Description of changes:*
Add the `gh_issue` tool that allows Q to open pre-filled issue templates with additional information. Similar to `q issue`, when asked to report an issue, a browser will open to the template (or URL if the browser cannot be opened).

Features:
- Triggers when asked to create or report an issue, or to open a feature request.
- Adds additional chat information automatically.
- Q could fill in expected/actual behavior, steps to reproduce, title.
- Includes any request IDs that resulted in an error.
- Includes chat context (from `/context show --expand`). This also includes their sizes.
- Includes the last **5** user-assistant requests from the chat transcript. This also includes any tool uses or errors that were displayed.

https://github.com/user-attachments/assets/b3f70548-bba2-490d-86eb-6e514f299157

### DIFF
--- a/crates/q_cli/src/cli/chat/conversation_state.rs
+++ b/crates/q_cli/src/cli/chat/conversation_state.rs
@@ -59,9 +59,10 @@ pub struct ConversationState {
     /// calling [Self::as_sendable_conversation_state].
     pub next_message: Option<UserInputMessage>,
     history: VecDeque<ChatMessage>,
-    /// Similar to history in that stores user and assistant responses, except that it is not used in message requests. 
-    /// Instead, the responses are expected to be in human-readable format, e.g user messages prefixed with '> '.
-    /// Should also be used to store errors posted in the chat.
+    /// Similar to history in that stores user and assistant responses, except that it is not used
+    /// in message requests. Instead, the responses are expected to be in human-readable format,
+    /// e.g user messages prefixed with '> '. Should also be used to store errors posted in the
+    /// chat.
     pub transcript: VecDeque<String>,
     tools: Vec<Tool>,
     /// Context manager for handling sticky context files
@@ -158,7 +159,7 @@ impl ConversationState {
 
         // Record message before adding context.
         self.append_user_transcript(&input);
-        
+
         // Combine context files with user input if available
         let content = if let Some(context) = context_files {
             format!("{}\n{}", context, input)
@@ -472,18 +473,14 @@ impl ConversationState {
         self.context_message_length
     }
 
-    pub fn append_user_transcript(&mut self, message: &String) {
+    pub fn append_user_transcript(&mut self, message: &str) {
         self.append_transcript(format!("> {}", message.replace("\n", "> \n")));
     }
 
     pub fn append_assistant_transcript(&mut self, message: &AssistantResponseMessage) {
-        let tool_uses = message.tool_uses.as_deref()
-            .map_or("none".to_string(), |tools| {
-                tools.iter()
-                    .map(|tool| tool.name.clone())
-                    .collect::<Vec<_>>()
-                    .join(",")
-            });
+        let tool_uses = message.tool_uses.as_deref().map_or("none".to_string(), |tools| {
+            tools.iter().map(|tool| tool.name.clone()).collect::<Vec<_>>().join(",")
+        });
         self.append_transcript(format!("{}\n[Tool uses: {tool_uses}]", message.content.clone()));
     }
 
--- a/crates/q_cli/src/cli/chat/conversation_state.rs
+++ b/crates/q_cli/src/cli/chat/conversation_state.rs
@@ -31,7 +31,6 @@ use tracing::{
     info,
     warn,
 };
-use regex::Regex;
 
 use super::context::ContextManager;
 use super::tools::ToolSpec;
@@ -60,6 +59,10 @@ pub struct ConversationState {
     /// calling [Self::as_sendable_conversation_state].
     pub next_message: Option<UserInputMessage>,
     history: VecDeque<ChatMessage>,
+    /// Similar to history in that stores user and assistant responses, except that it is not used in message requests. 
+    /// Instead, the responses are expected to be in human-readable format, e.g user messages prefixed with '> '.
+    /// Should also be used to store errors posted in the chat.
+    pub transcript: Vec<String>,
     tools: Vec<Tool>,
     /// Context manager for handling sticky context files
     pub context_manager: Option<ContextManager>,
@@ -93,6 +96,7 @@ impl ConversationState {
             conversation_id,
             next_message: None,
             history: VecDeque::new(),
+            transcript: Vec::new(),
             tools: tool_config
                 .into_values()
                 .map(|v| {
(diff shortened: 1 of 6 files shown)


crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -92,6 +97,7 @@ impl ConversationState {
             conversation_id,
             next_message: None,
             history: VecDeque::new(),
+            transcript: VecDeque::with_capacity(MAX_CONVERSATION_STATE_HISTORY_LEN),
```
kkashilk, line 100: Do we need to store `MAX_CONVERSATION_STATE_HISTORY_LEN` if we only plan to log the last 5 messages?

crates/q_cli/src/cli/chat/tools/gh_issue.rs
```diff
@@ -0,0 +1,161 @@
...
+                "profile=\n{}\n\n",
+                &ctx_manager.profile_config.paths.join("\n")
+            ));
+        }
+
+        // Handle context files
+        match ctx_manager.get_context_files(false).await {
+            Ok(context_files) if !context_files.is_empty() => {
+                ctx_str.push_str("files=\n");
+                let total_size: usize = context_files
```
kkashilk, line 132: might be useful to also record how many turns it has been in this conversation overall.

crates/q_cli/src/cli/chat/conversation_state.rs
mr-lee: that aside - would it be useful to see all the messages, or make that possible to toggle?

I think users may want the option to share the transcript or not.

crates/q_cli/src/cli/chat/tools/gh_issue.rs
mr-lee: Also how many with tool use!

saisrinivas26: also does this store mode of conversation like /acceptall or default behaviour?

crates/q_cli/src/cli/chat/conversation_state.rs
kkashilk: Users can edit the messages they wanna include in Github Issues UI, which would be better for editing/removing data they don't wanna share.

crates/q_cli/src/cli/chat/tools/mod.rs
```diff
@@ -68,16 +76,24 @@ impl Tool {
...
 
     /// Invokes the tool asynchronously
-    pub async fn invoke(&self, context: &Context, updates: &mut impl Write) -> Result<InvokeOutput> {
+    // TODO: Need to rework this to avoid passing in args meant for a single tool.
+    pub async fn invoke(
+        &self,
+        context: &Context,
+        updates: &mut impl Write,
+        gh_issue_context: GhIssueContext<'_>,
+    ) -> Result<InvokeOutput> {
```
dingfeli, line 90: The current flow is that the construction of `Tool` is where the necessary arguments are passed. i.e. The `Tool`s are stateful. 
Is there a way contextualize `GhIssue` so that these information are just part of the `Tool` instance?

crates/q_cli/src/cli/chat/conversation_state.rs
hayemaxi: > Do we need to store MAX_CONVERSATION_STATE_HISTORY_LEN if we only plan to log the last 5 messages?
No, but 5 was arbitrary so this keeps it disjoint in case we changed it or made it dynamic.

> would it be useful to see all the messages
We can increase the value, maybe 10 or 15? I am concerned about overloading the amount of information in the issue post and having the user trim it down. It is easier to trim down a few messages rather than 100 (or more).

kkashilk: >No, but 5 was arbitrary so this keeps it disjoint in case we changed it or made it dynamic.

Right, what I mean is to use a common constant that dictates how many messages we store and report

hayemaxi: Yes, you need some size bigger than the max amount to return. The transcript is not necessarily (1 user, 1 chat), (1 user, 1 chat). It could be (1 user, 5 chat), (1 user, 2 chat), etc.
Things like tool uses, errors, etc. will create more assistant chats.

crates/q_cli/src/cli/chat/tools/mod.rs
hayemaxi: The construction of tool happens only from model input, but the GhIssueContext comes from internal state not present from model input. As it currently sits, I will still have to pass it in somewhere that is out of place.

I will look into a way to avoid this in a follow up.

dingfeli: Perhaps we can do it from validate? 
https://github.com/aws/amazon-q-developer-cli/blob/7307b539eb9cbda430ed737a3ac3e001c6544623/crates/q_cli/src/cli/chat/mod.rs#L1272-L1285

We can try to contextualize the `Tooluse` some more here. Something like: 
```
    async fn validate_tools(&mut self, tool_uses: Vec<ToolUse>) -> Result<ChatState, ChatError> {
        let conv_id = self.conversation_state.conversation_id().to_owned();
        debug!(?tool_uses, "Validating tool uses");
        let mut queued_tools: Vec<QueuedTool> = Vec::new();
        let mut tool_results = Vec::new();
        for tool_use in tool_uses {
            let tool_use_id = tool_use.id.clone();
            let mut tool_telemetry = ToolUseEventBuilder::new(conv_id.clone(), tool_use.id.clone())
                .set_tool_use_id(tool_use_id.clone())
                .set_tool_name(tool_use.name.clone())
                .utterance_id(self.conversation_state.message_id().map(|s| s.to_string()));
            // we're wtihin `ChatContext` here and the values you needed are stored as states
            // contextualize will be a method that takes `ToolUse` and enrich it with the needed context
            let tool_use = self.contextualize(tool_use);
            match Tool::try_from(tool_use) {
                Ok(mut tool) => {
                    match tool.validate(&self.ctx).await {
                        Ok(()) => {
                            tool_telemetry.is_valid = Some(true);
                            queued_tools.push((tool_use_id.clone(), tool));
                        },
                        Err(err) => {
                            tool_telemetry.is_valid = Some(false);
                            tool_results.push(ToolResult {
                                tool_use_id: tool_use_id.clone(),
                                content: vec![ToolResultContentBlock::Text(format!(
                                    "Failed to validate tool parameters: {err}"
                                ))],
                                status: ToolResultStatus::Error,
                            });
                        },
                    };
                },
```

If we are trying to make the next release we can always do this in a follow up though.

dingfeli (approved):

crates/q_cli/src/cli/chat/tools/mod.rs
hayemaxi: Looking into this as a follow up.

Issue #957, opened 2025-03-27
[README] - Make it pretty


GoodluckH:
Add logo, emojis, badges, etc

PR #976, opened 2025-03-27
Prettify README + setup script

GoodluckH:
*Issue #, if available:*
#957 
*Description of changes:*
* Added Installation instructions
* Added logos, emojis, etc
* Added Features banner image


scripts/setup.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "Detected Arch"
+    sudo pacman -Syu --noconfirm
+    sudo pacman -S --noconfirm --needed webkit2gtk base-devel curl wget openssl appmenu-gtk-module gtk3 libappindicator-gtk3 librsvg libvips cmake jq pkgconf
+  elif [ -f /etc/fedora-release ]; then
+    echo "Detected Fedora"
+    sudo dnf check-update
+    sudo dnf install -y webkit2gtk3-devel.x86_64 openssl-devel curl wget libappindicator-gtk3 librsvg2-devel jq
+    sudo dnf group install -y "C Development Tools and Libraries"
+  else
+    echo "Unsupported Linux distribution. Check the docs for manual installation instructions."
```
mr-lee, line 22: tsk how dare you leave gentoo off the list.  shame

scripts/setup.sh
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env bash
+
+set -e
+
+OS="$(uname -s)"
+
+install_linux_deps() {
```
mr-lee, line 7: feel like we should have amazon linux here because well, this is technically an AWS thing lol

mr-lee (approved):
This looks fantastic!

scripts/setup.sh
GoodluckH: Can add for Amazon and Gentoo. But would need someone knowledge to supply the bash scripts lol

I have no idea what those two platforms need for dependencies

mr-lee: I'm kidding mostly about gentoo - but amazon yeah let's add it.  It's mostly a fedora clone, so yum will work.

Issue #955, opened 2025-03-27
Store conversation history in durable format


mschrage:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.3.0 (24D60)

### Expected behaviour

> it would be great to have the CLI store the command / prompt history in a durable form, .bash_history -style.

### Actual behaviour

CLI does not have this functionality right now.

### Environment

```yaml
[q-details]
version = "1.7.2"
hash = "b6bf5f1675dd7395009e8bc65c1145820671f211"
date = "2025-03-21T01:36:16.66764Z (6d ago)"
variant = "full"

[system-info]
os = "macOS 15.3.0 (24D60)"
chip = "Apple M1"
total-cores = 8
memory = "16.00 GB"

[environment]
cwd = "/Users/USER/experiments/amazon-q-developer-cli"
cli-path = "/Users/USER/experiments/amazon-q-developer-cli"
os = "Mac"
shell-path = "/bin/zsh"
shell-version = "5.9"
terminal = "macOS"
install-method = "unknown"

[env-vars]
[...]

PR #978, opened 2025-03-27
[RFC] Persistent conversations

hsbakshi:
This RFC proposes adding persistent conversation support to Amazon Q CLI, enabling users to save and resume conversations across sessions. The feature introduces automatic conversation continuation based on working directories, named conversations, and local storage support using SQLite, with future extensibility for cloud storage options. This enhancement aims to improve user experience for multi-day projects, team collaboration, and automation scenarios.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/rfcs/persistent_conversations_q_cli
+++ b/rfcs/persistent_conversations_q_cli
@@ -1,0 +1,90 @@
+- Feature Name: persistent_conversations_q_cli
+- Start Date: 2024-03-27
+
+# Summary
+
+Enhance Amazon Q CLI Chat with persistent conversation management capabilities, enabling users to save, load, and manage conversations across sessions. This feature will initially support local storage with automatic conversation continuation based on the working directory, with future extensibility for cloud storage options.
+
+# Motivation
+
+The current ephemeral nature of Amazon Q CLI conversations presents several challenges. When a terminal closes, all conversation history is lost, forcing users to start from scratch. This lack of continuity is particularly problematic for people intermittently working on projects, where the context from previous hours or days is crucial. Moreover, it hampers the sharing of complex troubleshooting or design sessions with colleagues, especially across different time zones. The current setup also limits the potential for automation through scripting and impedes efficient knowledge transfer among team members. By introducing conversation persistence, we can address these issues, providing continuity, enabling collaboration, and supporting automation scenarios.
+
+
+The addition of conversation persistence will address these issues by providing continuity, enabling collaboration, and supporting automation scenarios.
+
+# Guide-level explanation
+
+The persistent conversations feature introduces several new capabilities to Amazon Q CLI:
+
+1. Automatic Continuation:
+```bash
+$ q
+Welcome to Amazon Q CLI!
+
+Would like to continue your previous conversation? (y/n)
+```
+    
+
+    
+2. Named Conversations:
+    
+```
+# Create/load a named conversation
+$ q --conversation my_project
+
+# List available conversations
+$ q --list-conversations
+Available conversations:
+1. my_project (Last modified: 2024-12-15 14:30)
+2. debug_lambda (Last modified: 2024-12-14 09:45)
+
+# Delete a conversation
+$ q --delete-conversation debug_lambda
+
+```
+
+    
+3. In-Chat Commands:
+* /conversation <name>: Create or load a conversation
+* /list-conversations: Show available conversations
+* /delete-conversation <name>: Remove a saved conversation
+
+# Reference-level explanation
+
+## Storage Implementation:
+
+We will use SQLite for conversation storage, located in the ~/.aws/amazonq/conversations/ directory. Each conversation will be stored in a file named using the SHA-1 hash of the working directory for automatic continuation. The schema will include the conversation history (with line limits), profile reference, context reference, and metadata such as timestamps.
+
+## Conversation Management:
+
+For directory-based automatic continuation, we'll create a auto-generated conversation ID for the working directory. When starting Q in the same directory, it will ask whether to load the previous conversation, or to start a fresh session. If the user had multiple conversations in the same directory, then we will provide an option to select from the last 3, sorted based on the last modified timestamp. Named conversations will be managed explicitly through CLI arguments, persisting across directory changes and supporting multiple concurrent conversations.
+
+## Telemetry:
+
+We will track the usage of new CLI arguments, monitor the number of total and simultaneous conversations, and track conversation file sizes to help us understand user behavior and optimize the feature.
+
+# Drawbacks
+
+Implementing persistent conversations introduces increased complexity in conversation management. It will require additional storage on the user's system and may have a slight performance impact due to persistent storage operations. We'll also need to consider security measures for stored conversation data.
+
+# Rationale and alternatives
+
+We've proposed to use SQLite for robust storage with good performance. The directory-based automatic continuation offers an intuitive user experience, and starting with local storage reduces complexity for the initial implementation.
+
+We considered alternatives such as a cloud-first storage approach, which was rejected due to increased complexity, and plain text storage, which was deemed inadequate for performance and maintainability reasons. 
+
+# Unresolved questions
+
+Several questions remain to be resolved during the implementation phase. We need to determine maximum conversation history limits and whether compression is required for storage optimization. Cleanup policies for old conversations and security measures for sensitive content need to be established. Additionally, we need to develop a migration strategy for future cloud storage support.
+
+# Future possibilities
+
+Looking ahead, we see several exciting possibilities for expanding this feature. We could integrate cloud storage options, including S3 and HTTPS-based storage, enabling cross-device synchronization. Enhanced collaboration features like conversation sharing, team workspace support, and access control mechanisms could be implemented.
+
+We also envision advanced management capabilities such as conversation tagging, powerful search functionality, and export/import features. For automation, we could develop an API for programmatic conversation management, integrate with CI/CD pipelines, and support batch processing capabilities.
+
+These future enhancements would further improve the utility and flexibility of persistent conversations in Amazon Q CLI, making it an even more powerful tool for developers and teams.
+
+
+
+


rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+# Guide-level explanation
+
+The persistent conversations feature introduces several new capabilities to Amazon Q CLI:
+
+1. Automatic Continuation:
+```bash
+$ q
+Welcome to Amazon Q CLI!
+
+Would like to continue your previous conversation? (y/n)
```
hayemaxi, line 24: `[y/n/never]`

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+# Delete a conversation
+$ q --delete-conversation debug_lambda
+
+```
+
+    
+3. In-Chat Commands:
+* /conversation <name>: Create or load a conversation
+* /list-conversations: Show available conversations
+* /delete-conversation <name>: Remove a saved conversation
```
hayemaxi, line 50: conversation is verbose, maybe
```
/session create
/session load
/session list
/session delete
```

Further ideas:
- Creating a session (conversation) should record everything that has been done so far as well as any future interactions for the session. Prompts/responses should be stored frequently
- Creating a session within a session cuts off the previous session saving at that point. The new session has all the previous session data as well as whatever is going to come
- there is a default (anonymous) session that is always running/available. it is overwritten each time the user starts a new default session.

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+
+# Reference-level explanation
+
+## Storage Implementation:
+
+We will use SQLite for conversation storage, located in the ~/.aws/amazonq/conversations/ directory. Each conversation will be stored in a file named using the SHA-1 hash of the working directory for automatic continuation. The schema will include the conversation history (with line limits), profile reference, context reference, and metadata such as timestamps.
+
+## Conversation Management:
+
+For directory-based automatic continuation, we'll create a auto-generated conversation ID for the working directory. When starting Q in the same directory, it will ask whether to load the previous conversation, or to start a fresh session. If the user had multiple conversations in the same directory, then we will provide an option to select from the last 3, sorted based on the last modified timestamp. Named conversations will be managed explicitly through CLI arguments, persisting across directory changes and supporting multiple concurrent conversations.
```
hayemaxi, line 60: Is directory based necessary? Would like to see other thoughts on if we can just have it be global instead.

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+
+## Telemetry:
+
+We will track the usage of new CLI arguments, monitor the number of total and simultaneous conversations, and track conversation file sizes to help us understand user behavior and optimize the feature.
+
+# Drawbacks
+
+Implementing persistent conversations introduces increased complexity in conversation management. It will require additional storage on the user's system and may have a slight performance impact due to persistent storage operations. We'll also need to consider security measures for stored conversation data.
```
hayemaxi, line 68: What are the security implications? How can we mitigate them?

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+
+# Rationale and alternatives
+
+We've proposed to use SQLite for robust storage with good performance. The directory-based automatic continuation offers an intuitive user experience, and starting with local storage reduces complexity for the initial implementation.
```
hayemaxi, line 72: Is this the proper storage mechanism? Maybe we can leverage https://github.com/aws/amazon-q-developer-cli/issues/955

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+
+We considered alternatives such as a cloud-first storage approach, which was rejected due to increased complexity, and plain text storage, which was deemed inadequate for performance and maintainability reasons. 
+
+# Unresolved questions
+
+Several questions remain to be resolved during the implementation phase. We need to determine maximum conversation history limits and whether compression is required for storage optimization. Cleanup policies for old conversations and security measures for sensitive content need to be established. Additionally, we need to develop a migration strategy for future cloud storage support.
```
hayemaxi, line 78: We should define limits here. E.g. if appending next prompt/response breaches 1mb, remove oldest prompt/response

rfcs/persistent_conversations_q_cli
hayemaxi: Instead,

`Enter 'y' to continue your previous conversation. Otherwise, start chatting.`

rfcs/persistent_conversations_q_cli
brandonskiser, line 78: SQLite benefit is binary storage won't require any compression logic on our part.

A couple unresolved questions to consider:
1. File size limit on the SQLite db size (once limit is reached, continually delete old messages until under the limit)
2. Cleanup policies for old conversation
3. Encryption (maybe symmetric encryption by using some consistent state of the user's machine, e.g. hardware id)

rfcs/persistent_conversations_q_cli
brandonskiser: SQLite would provide straightforward ways to:
1. Have a minimal file size - no need to compress on our part.
2. Incrementally update the conversation history without having to write the whole conversation as e.g. a JSON file.
3. Provide intelligent search capabilities, for whatever reason.

rfcs/persistent_conversations_q_cli
brandonskiser, line 68: We'd need a mechanism to track both file size of an individual conversation, and the total size stored on the user machine. If we intend to have these conversations shared across IDE + CLI in the future, then the ownership can get weird here though - both implementations would need to have consistent behavior for controlling file limits. For this initial version, I think it's fine to go with the simplest approach though.

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+3. In-Chat Commands:
+* /conversation <name>: Create or load a conversation
+* /list-conversations: Show available conversations
+* /delete-conversation <name>: Remove a saved conversation
+
+# Reference-level explanation
+
+## Storage Implementation:
+
+We will use SQLite for conversation storage, located in the ~/.aws/amazonq/conversations/ directory. Each conversation will be stored in a file named using the SHA-1 hash of the working directory for automatic continuation. The schema will include the conversation history (with line limits), profile reference, context reference, and metadata such as timestamps.
```
brandonskiser, line 56: This should probably be namespaced to the CLI until we have a clear vision for sharing conversations across the IDE and CLI? so instead have it be e.g.
- qcli_SHA1(directory) - for directory conversations
- qcli_SHA1(name) - for named conversations

I'm guessing the IDE and CLI will have different implementations out of the gate for handling conversation saving and state, unless we can hard decide on a schema + contract now (unlikely?)

rfcs/persistent_conversations_q_cli
brandonskiser: > Prompts/responses should be stored frequently

Probably at every user prompt unless we have a good reason not to.

> Creating a session within a session cuts off the previous session saving at that point. The new session has all the previous session data as well as whatever is going to come

Yeah, so the user can run some conversation clear command if they want to have a blank conversation instead.

> there is a default (anonymous) session that is always running/available if a named session is not in use.

Is this referring to a "global" session that isn't a directory or named session?

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+
+    
+2. Named Conversations:
+    
+```
+# Create/load a named conversation
+$ q --conversation my_project
+
+# List available conversations
+$ q --list-conversations
```
brandonskiser, line 36: How are directory conversations named? If we store a conversation with a hash of the directory, how could multiple conversations for the same directory be supported?

rfcs/persistent_conversations_q_cli
brandonskiser, line 60: > If the user had multiple conversations in the same directory, then we will provide an option to select from the last 3

If I understand correctly, we have two types of conversations:
1. Named conversations - explicitly created by the user.
2. Directory-based conversations - implicitly created on every new `q chat` session (without the `--conversation` flag).

So we're saying that we want to support multiple unnamed conversations per directory, in which case we need some way to uniquely identify each by name.

brandonskiser, line 60: How do we handle multiple terminal sessions for the same directory?

If multiple terminal sessions are writing to the same named conversation or directory conversation, then confusion will ensue - we should have some locking mechanism (related example: vim swapfiles) so that the user is at least warned about another session potentially editing the same conversation.

rfcs/persistent_conversations_q_cli
```diff
@@ -0,0 +1,90 @@
...
+* /list-conversations: Show available conversations
+* /delete-conversation <name>: Remove a saved conversation
+
+# Reference-level explanation
+
+## Storage Implementation:
+
+We will use SQLite for conversation storage, located in the ~/.aws/amazonq/conversations/ directory. Each conversation will be stored in a file named using the SHA-1 hash of the working directory for automatic continuation. The schema will include the conversation history (with line limits), profile reference, context reference, and metadata such as timestamps.
+
+## Conversation Management:
```
ANorwell, line 58: How do conversations interact with projects?

As a user, 90% of what I personally want here is just a single ongoing conversation and set of context files per dir. The only commands I want to use most of the time would be:

```
q  # the dir determines the project and what context I'm using
/clear  # explicitly start a new conversation. The default is to preserve the conversation
/history # See this entire conversation
```

It would be nice to explicitly select projects and spin up multiple projects/conversations in same directory on occasion, but it's the minority case.


(n fact, I did build it more or less this way for https://github.com/ANorwell/claippy)

PR #1035, opened 2025-04-01
feat(chat): Add /ed command to compose prompts in text editor

### LINKED ISSUE #1029: Ability to use $EDITOR to compose a prompt
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.3.2 (24D81)

### Expected behaviour

I would love to use an editor to compose my prompt.
For example I would like to have the chance to open a vim instance to put my prompt, to have better multiline and markdown support.

### Actual behaviour

I can only use the raw cli to input prompts.

### Steps to reproduce

N/A

### Environment

```yaml
[q-details]
version = "1.7.2"
hash = "b6bf5f1675dd7395009e8bc65c1145820671f211"
date = "2025-03-21T01:36:16.66764Z (11d ago)"
variant = "full"

[system-info]
os = "macOS 15.3.2 (24D81)"
chip = "Apple M3 Pro"
total-cores = 12
memory = "36.00 GB"

[environment]
cwd = "/Users/USER"
cli-path = "/Users/USER"
os = "Mac"
shell-path = "/bin/zsh"
shell-version = "5.9"
terminal = "Wezterm"
install-method = "unknown"

[env-vars]
QTERM_SESSION_ID = "ee773d43f1f34431865ab14086f47623"
Q_SET_PARENT_CHECK = "1"
Q_TERM = "1.7.2"
SHELL = "/bin/zsh"
TERM = "xterm-256color"
__CFBundleIdentifier = "com.github.wez.wezterm"
```

carlesso:
Implement the /ed command that opens the user's preferred editor to compose a prompt. The content is automatically submitted when the editor closes, improving the user experience for longer prompts.

Fixes #1029

*Issue #, if available:* #1029

*Description of changes:*


crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -96,6 +104,7 @@ const WELCOME_TEXT: &str = color_print::cstr! {"
 • Help me understand my git status
 
 <em>/acceptall</em>    <black!>Toggles acceptance prompting for the session.</black!>
+<em>/ed</em>           <black!>Open your $EDITOR to compose a prompt</black!>
```
mschrage, line 107: Can you remove /ed from the masthead text, but keep it in the /help text?

mschrage (changes requested):
Thanks so much for your contribution! One small nit to avoid overloading users with commands when they are getting started.

crates/q_cli/src/cli/chat/mod.rs
carlesso: Absolutely! Updated!

GoodluckH:
Thanks for contributing! Was running into errors when trying it out.

## Reproduce
1. In q chat, `/ed`
2. Write anything
3. Exit using `:wq`

Screenshot 2025-04-01 at 4 25 52 PM

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -111,6 +119,7 @@ const HELP_TEXT: &str = color_print::cstr! {"
 
 <em>/clear</em>        <black!>Clear the conversation history</black!>
 <em>/acceptall</em>    <black!>Toggles acceptance prompting for the session.</black!>
+<em>/ed</em>           <black!>Open your $EDITOR to compose a prompt</black!>
```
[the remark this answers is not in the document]
kkashilk: Also, good to have a descriptive command name like `/edit_prompt` with a short-form that you can use in most cases.

Plus, using the <TAB> key will auto-complete commands, so you ideally won't have to type out the whole thing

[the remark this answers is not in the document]
carlesso: That make sense. What if instead of `/edit_prompt` we use `/editor`? As it's more that what it does.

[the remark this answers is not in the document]
carlesso: Discussed offline.

crates/q_cli/src/cli/chat/prompt.rs
```diff
@@ -30,6 +30,7 @@ const COMMANDS: &[&str] = &[
     "/clear",
     "/help",
     "/acceptall",
+    "/ed",
```
kkashilk, line 33: need to change this to `/editor` as well

carlesso: Done!

mschrage:
This seems good. One open question for you:

Should there be a /editor command at all or should this just be a setting?

carlesso:
@mschrage imho it should be a command. I don't want to do it every time, but every now and then I have a more complex prompt (in an existing q chat session) where I want to format the text and/or copypaste snippet of codes.

For example once I was using q to prepare some document, and I wanted to create some "correct-looking" snippets of python, so I used the `/editor` to say something like:
---

This is a sample json of data:
```
json sample
```

Can you generate some python code to build objects following the suggestion?
```
copy/paste of some other readme
```

mschrage:
Can you fix the lint issues? Then we are good to merge.

carlesso:
Sure, waiting for the validation to run.

mschrage (approved):

PR #1040, opened 2025-04-01
rfc(preprocessors): rfc for MCP preprocessor tools

mikewrighton:
This RFC introduces **Preprocessor Tools** in the Model Context Protocol (MCP). Preprocessor tools are executed deterministically for every prompt before the user input is processed by the language model (LLM). Their outputs are injected into the request context but do not persist in the conversation history. Preprocessor tools are not available as general tools and are excluded from the tool specifications sent to the LLM.

### DIFF
--- a/rfcs/0002-preprocessor-tools.md
+++ b/rfcs/0002-preprocessor-tools.md
@@ -1,0 +1,112 @@
+- Feature Name: preprocessor_tools
+- Start Date: 2025-03-31
+
+# Summary
+
+[summary]: #summary
+
+This RFC introduces **Preprocessor Tools** in the Model Context Protocol (MCP). Preprocessor tools are executed deterministically for every prompt before the user input is processed by the language model (LLM). Their outputs are injected into the request context but do not persist in the conversation history. Preprocessor tools are not available as general tools and are excluded from the tool specifications sent to the LLM.
+
+# Motivation
+
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


mschrage:
Curious to hear your thoughts on RFC #1050 ?

These definitely can coexist. Effectively your `preprocessor` flag would signal to the Q CLI to add a new context hook.

However, I'm curious if we should extend the MCP protocol at all or just use the context hooks mechanism as proposed.

mikewrighton:
> Curious to hear your thoughts on RFC #1050 ?
> 
> These definitely can coexist. Effectively your `preprocessor` flag would signal to the Q CLI to add a new context hook.
> 
> However, I'm curious if we should extend the MCP protocol at all or just use the context hooks mechanism as proposed.

RFC #1050 makes sense to me and I agree they can both exist, where the control lives in the context hook mechanism (more flexibility for users) rather than just being built-in preprocessor-handling logic. 

So then the question of whether we need the new MCP field is valid. There's one thing it does solve - in the case of `RetrieveMemories` (and probably other context-providing tools, too) we don't want it to also be available as a standard tool because it would usually end up getting called twice: once as a preprocessor, and again by the LLM (depending on the tool description). If there's a field in the MCP tool spec saying 'this is a special preprocessor tool' then we can exclude it from the regular tool list we send to the LLM. I guess an alternative is to say any tool which is added as a preprompt context hook will be excluded from the tool list. But this might be less obvious behavior. What do you think?

Issue #1030, opened 2025-04-01
Feature Request: Enable Amazon Q Chat to Use User's Shell Environment


prerna-p:
## Summary
Amazon Q Chat should execute commands within the user's actual shell environment (bash, zsh, etc.) to access user-defined aliases, functions, environment variables, and other shell customizations.

## Problem Statement
Currently, Amazon Q Chat executes commands in a separate bash shell that doesn't have access to the user's shell configuration. This means:

1. Custom aliases defined in .zshrc, etc. are not available
2. User-defined shell functions cannot be used
3. Custom environment variables are not accessible
4. PATH modifications and other shell customizations are not applied

This creates a disconnect between how users work in their terminal and how Amazon Q Chat executes commands, requiring users to provide full command paths or avoid using their custom shortcuts.

## Proposed Solution
Modify Amazon Q Chat to:

1. Detect the user's default shell (bash, zsh, fish, etc.)
2. Execute commands through that shell with the user's configuration loaded
3. Preserve the shell's environment between command executions within a session

For example, when a user with zsh as their default shell asks Amazon Q to run a command like "bb" (which might be an alias for "brazil-build"), Amazon Q should:
• Execute the command through zsh
• Load the user's .zshrc configuration
• Run the command with all aliases and functions available

## Benefits
• Seamless integration with the user's existing workflow
• Reduced friction when asking Amazon Q to execute commands
• Ability to use shorthand commands and aliases familiar to the user
• Consistent environment between terminal and Amazon Q Chat

## Use Cases
1. Running aliased commands like "gt" instead of "gradle tasks"
2. Using custom functions defined in shell configuration
3. Accessing environment variables set in shell startup files
4. Utilizing modified PATH or other environment customizations

## Technical Considerations
• Security implications of executing commands in the user's shell environment
• Handling different shell types (bash, zsh, fish, etc.)
• Maintaining shell state between command executions
• Potential performance implications

## Example

When a user with an alias "describereg" for 'aws cloudformation describe-type-registration --profile default --registration-token' in their .zshrc asks:

`run describereg foobar` 

Amazon Q should execute this in the user's zsh environment with their .zshrc loaded, effectively running:

`aws cloudformation describe-type-registration --profile default --registration-token foobar`

jsamuel1:
[...]

PR #1043, opened 2025-04-01
feat(chat): support user shell env in shell command tool

GoodluckH:
*Issue #, if available:*
#1030 

*Description of changes:*
* Renamed `execute_bash` tool to `execute_shell_commands`
* Fetches all aliases based on Shell type (currently support `zsh` and `bash`) from user config files

### DIFF
--- a/crates/q_cli/src/cli/chat/tools/execute_shell_commands.rs
+++ b/crates/q_cli/src/cli/chat/tools/execute_shell_commands.rs
@@ -1,7 +1,45 @@
 use std::collections::VecDeque;
-use std::fs;
-use std::io::Write;
-use std::process::Stdio;
+use std::io::{
+    self,
+    Write,
+};
+use std::os::fd::{
+    AsFd,
+    AsRawFd,
+    FromRawFd,
+    RawFd,
+};
+use std::path::Path;
+
+use console::strip_ansi_codes;
+use fig_os_shim::Context;
+use filedescriptor::FileDescriptor;
+use nix::fcntl::{
+    FcntlArg,
+    FdFlag,
+    OFlag,
+    fcntl,
+    open,
+};
+use nix::libc;
+use nix::pty::{
+    Winsize,
+    grantpt,
+    posix_openpt,
+    ptsname,
+    unlockpt,
+};
+use nix::sys::signal::{
+    SigHandler,
+    Signal,
+    signal,
+};
+use nix::sys::stat::Mode;
+use portable_pty::unix::close_random_fds;
+use tokio::io::unix::AsyncFd;
+use tokio::select;
+use tokio::sync::mpsc::channel;
+nix::ioctl_write_ptr_bad!(ioctl_tiocswinsz, libc::TIOCSWINSZ, Winsize);
 
 use crossterm::queue;
 use crossterm::style::{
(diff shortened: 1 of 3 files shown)


GoodluckH:
Yes, I initially tried using `interactive` mode to source config files but it will result in tty suspension error that will kill the Q CLI session.

I'm still investigating how to avoid that.

mschrage:
Some interesting discussion here:

> This will fix it:
>
>stty -tostop
>From the man page:
>
>tostop (-tostop)
>Send (do not send) SIGTTOU for background output. This causes background jobs to stop if they attempt terminal output.
>This tostop is normally the default setting, as it's usually undesirable to mix the output of

https://stackoverflow.com/questions/24056102/why-do-i-get-suspended-tty-output-in-one-terminal-but-not-in-others

GoodluckH:
Yes I tried all those but didn't work. 

Instead, I will launch a PTY.

jsamuel1:
zsh, bash, fish.
What other ones are supported in other parts of q cli?

How will automated tests work for this?  Docker based, or can you do it with chroot/mocks?

crates/q_cli/src/cli/chat/tools/execute_shell_commands.rs
```diff
@@ -82,66 +128,164 @@ impl ExecuteBash {
...
 
-        let stdout = child.stdout.take().unwrap();
-        let stdout = tokio::io::BufReader::new(stdout);
-        let mut stdout = stdout.lines();
+        unsafe { command.pre_exec(pre_exec_fn) };
 
-        let stderr = child.stderr.take().unwrap();
-        let stderr = tokio::io::BufReader::new(stderr);
-        let mut stderr = stderr.lines();
+        let mut child = command.spawn()?;
```
brandonskiser, line 219: This method seems to break our (admittedly already poor support of) interactive commands, e.g. `npx create-react-app helloworld`

My thought was that we would spawn the child with a pty, and set our terminal to raw mode. This way, we can copy all bytes directly to the child which can then handle sigint etc. which should support more interactive commands.

crates/q_cli/src/cli/chat/tools/execute_shell_commands.rs
```diff
@@ -154,17 +298,10 @@ impl ExecuteBash {
                     ""
                 }
             ),
-            "stderr": format!(
```
brandonskiser, line 157: removing stderr?

crates/q_cli/src/cli/chat/tools/execute_shell_commands.rs
```diff
@@ -82,66 +128,164 @@ impl ExecuteBash {
...
+                }
             };
         }
         .wrap_err_with(|| format!("No exit status for '{}'", &self.command))?;
 
-        updates.flush()?;
-
-        let stdout = stdout_buf.into_iter().collect::<Vec<_>>().join("\n");
-        let stderr = stderr_buf.into_iter().collect::<Vec<_>>().join("\n");
+        let stdout = stdout_lines.into_iter().collect::<String>();
```
brandonskiser, line 288: I was also thinking about how we can handle interactive commands since some (e.g. `vim`) shouldn't have their stdout returned back to the model, still an open question to solve

brandonskiser (changes requested):

crates/q_cli/src/cli/chat/tools/execute_shell_commands.rs
GoodluckH: With the PTY mode it's hard to differentiate `stderr`, but we can already infer from status code if the resulting message is error or not. The output of this function is to give some context to the LLM to inform it of the execution result.

Issue #1049, opened 2025-04-02
[Feature Request] Implement “context hooks” which allow users or MCP servers to pass context either at the start of a conversation or alongside each prompt


jsamuel1:
Possible syntax:

```
/context add [init | preprompt ] <statement>
/context add preprompt !q-personalization --list-memories
```

Where statement can be anything type-able on the prompt.
For performance we would recommend one of:
`!   - execute a shell command`
`/   - execute an internal command`
`#  - execute a tool with exact parameters.`   <-- I don't think we have this at the moment.  Pick an appropriate character

Using this syntax, further files could be included in the context, or custom context inclusions, such as environment specific data.

jsamuel1:
See #1061 too, as a usecase for this.

PR #1050, opened 2025-04-02
Context hooks

jsamuel1:
This RFC proposes implementing context hooks that allow users or MCP servers to pass context either at the start of a conversation or alongside each prompt. The feature enables automated context injection without requiring manual /context commands.

*Issue #, if available:*
#1049 

*Description of changes:*

### DIFF
--- a/rfcs/0003-context-hooks.md
+++ b/rfcs/0003-context-hooks.md
@@ -142,6 +142,20 @@ This simplified syntax makes it easier to read and write hook configurations, es
 
 ## Hook Types
 
+### Inline Hooks
+
+Inline hooks execute a shell command directly:
+
+```
+env | grep AWS_
+
+# Output:
+AWS_PROFILE=development
+AWS_REGION=us-west-2
+```
+
+> **Note**: Inline hooks will be implemented first, and we will evaluate whether the other hook types (Script and HTTP) are required based on user feedback and use cases.
+
 ### Script Hooks
 
 Script hooks execute a local script that outputs context information:
@@ -171,13 +185,6 @@ Response:
 - Implement least privilege access
 ```
 
-### Inline Hooks
-
-Inline hooks execute a shell command directly:
-
-```
-env | grep AWS_
-
 # Output:
 AWS_PROFILE=development
 AWS_REGION=us-west-2
@@ -190,27 +197,54 @@ Once configured, context hooks run automatically:
 1. When starting a new conversation with `q chat`, conversation start hooks run
 2. When sending a prompt, per-prompt hooks run
 
-Users can see which hooks are active with the `/hooks` command:
+Users can see which hooks are active with the `/context hooks` command:
 
 ```
-> /hooks
+> /context hooks
 Active context hooks:
 - project-info (conversation start)
 - git-status (per prompt)
 ```
 
-Users can temporarily disable hooks with the `/hooks disable` command:
+Users can add inline hooks directly from the chat interface:
 
 ```
-> /hooks disable git-status
+> /context hooks add git-status --type=per-prompt --command="git status --short"
+Added hook: git-status (per-prompt)
+
+> /context hooks add project-info --type=conversation-start --command="echo 'Project: $(basename $(pwd))'"
+Added hook: project-info (conversation-start)
+```
+
+Users can temporarily disable hooks with the `/context hooks disable` command:
+
+```
+> /context hooks disable git-status
 Disabled hook: git-status
 ```
 
-And re-enable them with `/hooks enable`:
+And re-enable them with `/context hooks enable`:
 
 ```
-> /hooks enable git-status
+> /context hooks enable git-status
 Enabled hook: git-status
+```
+
+Users can also enable or disable all hooks at once:
+
+```
+> /context hooks disable-all
+Disabled all hooks
+
+> /context hooks enable-all
+Enabled all hooks
+```
+
+When hooks are running, the UI shows feedback about their execution:
+
+```
+> Running hook: git-status...
+> Hook git-status completed in 0.03s
 ```
 
 ### Context Visibility
@@ -247,6 +281,13 @@ Additional context visibility commands include:
 
 This visibility helps users understand what context is being provided to the LLM, making the system more transparent and easier to debug.
 
+### Hook Persistence and Behavior
+
+- **Conversation Start Hooks**: Evaluated once at the beginning of a conversation. Their context persists at the top of the conversation, ahead of the first prompt, even if earlier messages are compacted or cleared away.
+- **Per-Prompt Hooks**: Evaluated on every user prompt and attached immediately before the human text.
+- **Hook Disabling**: When hooks are disabled, they remain in the config file with a disabled flag and stay disabled until explicitly re-enabled.
+- **Clear Command**: Running `/clear` will re-evaluate conversation start hooks, refreshing their context.
+
 # Reference-level explanation
 
 [reference-level-explanation]: #reference-level-explanation
@@ -272,6 +313,8 @@ pub struct HookConfig {
     // Common fields
     pub timeout_ms: Option<u64>,
     pub max_output_size: Option<usize>,
+    pub criticality: Criticality,
+    pub cache_ttl_seconds: Option<u64>,
     // Type-specific fields
     pub path: Option<String>,        // For script hooks
     pub command: Option<String>,     // For inline hooks
@@ -285,8 +328,71 @@ pub enum HookType {
     Inline,
 }
 
+pub enum Criticality {
+    Fail,    // Hook failure will prevent prompt from being sent
+    Warn,    // Hook failure will log a warning but allow prompt to be sent
+    Ignore,  // Hook failure will be silently ignored
+}
+
 pub struct HookConfigManager {
     config_path: PathBuf,
+    conversation_start_hooks: Vec<HookConfig>,
+    per_prompt_hooks: Vec<HookConfig>,
+}
+
+impl HookConfigManager {
+    pub fn new() -> Result<Self> {
+        // Find and load config file
+        // Parse YAML into HookConfig structs
+        // Validate configurations
+    }
+    
+    pub fn get_conversation_start_hooks(&self) -> &[HookConfig] {
+        &self.conversation_start_hooks
+    }
+    
+    pub fn get_per_prompt_hooks(&self) -> &[HookConfig] {
+        &self.per_prompt_hooks
+    }
+    
+    pub fn add_inline_hook(&self, name: &str, hook_type: &str, command: &str) -> Result<()> {
+        // Create a new HookConfig for an inline hook
+        let hook_type = match hook_type {
+            "conversation-start" => HookType::Inline,
+            "per-prompt" => HookType::Inline,
+            _ => return Err(anyhow!("Invalid hook type: {}", hook_type)),
+        };
+        
+        let hook = HookConfig {
+            name: name.to_string(),
+            hook_type,
+            enabled: true,
+            timeout_ms: Some(5000), // 5 second default timeout
+            max_output_size: Some(10240), // 10KB default max output
+            criticality: Criticality::Warn, // Default to warn on failure
+            cache_ttl_seconds: Some(60), // Default 1 minute cache
+            path: None,
+            command: Some(command.to_string()),
+            url: None,
+            headers: None,
+        };
+        
+        // Add to appropriate list based on type
+        if hook_type == "conversation-start" {
+            self.conversation_start_hooks.push(hook);
+        } else {
+            self.per_prompt_hooks.push(hook);
+        }
+        
+        // Save configuration to file
+        self.save_config()
+    }
+    
+    fn save_config(&self) -> Result<()> {
+        // Save configuration to file
+        // This will be called after adding, enabling, or disabling hooks
+    }
+}
     conversation_start_hooks: Vec<HookConfig>,
     per_prompt_hooks: Vec<HookConfig>,
 }
@@ -313,9 +419,15 @@ impl HookConfigManager {
 The Hook Executor runs hooks and captures their output:
 
 ```rust
+struct CachedResult {
+    output: String,
+    timestamp: Instant,
+}
+
 pub struct HookExecutor {
     config_manager: Arc<HookConfigManager>,
     hook_registry: Arc<HookRegistry>,
+    cache: DashMap<String, CachedResult>,
 }
 
 impl HookExecutor {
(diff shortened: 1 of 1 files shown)


atxtechbro (approved):

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+Users can see which hooks are active with the `/hooks` command:
+
+```
+> /hooks
+Active context hooks:
+- project-info (conversation start)
+- git-status (per prompt)
+```
+
+Users can temporarily disable hooks with the `/hooks disable` command:
```
hayemaxi, line 202: Does this mean they are disabled for the session only?

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+```
+
+## Using Context Hooks
+
+Once configured, context hooks run automatically:
+
+1. When starting a new conversation with `q chat`, conversation start hooks run
+2. When sending a prompt, per-prompt hooks run
+
+Users can see which hooks are active with the `/hooks` command:
```
hayemaxi, line 193: Instead of `/hooks` command, why not just make an extension of `/context`?
E.g. `/context hooks`, `/context hooks disable ...`, etc.

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+Disabled hook: git-status
+```
+
+And re-enable them with `/hooks enable`:
+
+```
+> /hooks enable git-status
+Enabled hook: git-status
+```
+
```
hayemaxi, line 215: More commands:
- `/context hooks enable-all`
- `/context hooks disable-all`

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+1. **Script Execution**: Scripts run with the same permissions as the Amazon Q CLI process
+2. **HTTP Requests**: Sensitive information in headers (like tokens) should be stored securely
+3. **Output Validation**: Hook output should be validated to prevent injection attacks
+4. **Resource Limits**: Hooks should have timeouts and output size limits
+
+## Performance Considerations
+
+Context hooks can impact performance in several ways:
+
+1. **Startup Time**: Conversation start hooks run before the first prompt, potentially increasing startup time
```
hayemaxi, line 500: We should display any _conversation start_ hooks we are actively running during startup. Otherwise, long running or internet request hooks will definitely make it seem like Q is just frozen.

Side note: also attach how long it took to run the hook?

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+2. **Prompt Latency**: Per-prompt hooks run before each prompt, potentially increasing response time
```
hayemaxi, line 501: For these prompts I think displaying each hook that we run is noisy (unless they used `/context show --always`). For these, maybe we can display a warning if any hook exceeded some arbitrary run time, e.g. 10 seconds.

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+3. **Background Execution**: To mitigate performance impact, hooks can run in the background:
+   - Conversation start hooks run asynchronously during chat initialization
+   - Per-prompt hooks run in parallel with a configurable timeout
+4. **Caching**: Hook outputs can be cached with configurable TTL to avoid redundant executions
```
hayemaxi, line 505: Would this configuration live in the hooks json config?

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+4. **Resource Limits**: Hooks should have timeouts and output size limits
+
+## Performance Considerations
+
+Context hooks can impact performance in several ways:
+
+1. **Startup Time**: Conversation start hooks run before the first prompt, potentially increasing startup time
+2. **Prompt Latency**: Per-prompt hooks run before each prompt, potentially increasing response time
+3. **Background Execution**: To mitigate performance impact, hooks can run in the background:
+   - Conversation start hooks run asynchronously during chat initialization
```
hayemaxi, line 503: This means the first (or more!) prompts from the user might not include the expected context.

As a middle ground between this and the above comments (and this might be what you are suggesting), we can start the program with the hooks running in the background, but await on the hooks once a user submits a prompt.

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+3. **Protocol Extension**: This provides a standardized way for MCP servers to inject context without protocol changes
+4. **Organizational Policies**: Companies can implement centralized context policies through MCP-provided hooks
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+1. **Security Risks**: Executing arbitrary scripts and making HTTP requests introduces security risks
+2. **Performance Impact**: Running hooks with each prompt could slow down the conversation flow
+3. **Complexity**: Adds another configuration system that users need to understand
```
hayemaxi, line 522: In a future version, we could add the ability to create and update hooks via commands, e.g. `/context hooks add ...`

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+1. Users will continue to manually add context, leading to inconsistent usage
+2. Integration with external tools and systems will remain limited
+3. Automated workflows involving Amazon Q CLI will be harder to implement
+4. The CLI will be less powerful for advanced users and organizations
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+1. **Hook Ordering**: Should hooks run in a specific order? Should users be able to specify priority?
```
hayemaxi, line 588: Let's just choose an order now: the order they appear in the config file. We can make it clear to users that this is the case. Is there a need to have a different order?

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+2. **Error Handling**: How should errors in hook execution be presented to users?
```
hayemaxi, line 589: By default it would be useful to display them to users as they happen, similar to tool errors, but less verbose.

I wonder if we can add to the config to either silence hook errors or make them fatal, but this might add too much complexity.

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+```
+
+Where:
+- `!` prefix indicates a shell command
+- `/` prefix indicates an internal command
+- `#` prefix indicates a tool call
+
+This simplified syntax makes it easier to read and write hook configurations, especially for common use cases.
+
+## Hook Types
```
hayemaxi, line 143: Why do we need separate hook types? It seems we can achieve everything with just inline hooks, e.g

```
inline hook: sh ~/scripts/api_call

# api_call
result = curl ....
echo result
```

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,646 @@
...
+AWS_PROFILE=development
+AWS_REGION=us-west-2
+```
+
+## Using Context Hooks
+
+Once configured, context hooks run automatically:
+
+1. When starting a new conversation with `q chat`, conversation start hooks run
+2. When sending a prompt, per-prompt hooks run
```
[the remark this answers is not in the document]
jsamuel1: Almost.  

Conversation hook context is evaluated once, and persists at the top of the conversation, ahead of the first prompt.  (It will not scroll out of context)

Prompt hook context is evaluated on every human prompt and attached immediately before the human text.

rfcs/0003-context-hooks.md
jsamuel1: That was the thought.  "just in case" things break with the hook.

rfcs/0003-context-hooks.md
jsamuel1: Agree.  Updating.

rfcs/0003-context-hooks.md
jsamuel1: * Hook Completed in 0.03µS

rfcs/0003-context-hooks.md
jsamuel1: To align with how other tools do this sort of thing, we might be better if this "disables until enabled", but doesn't remove it from the config file.  (ie. across sessions).   

Keen for a view here.

jsamuel1:
I'll make the requested changes to the RFC document. These changes focus on:

1. Changing the command structure from /hooks to /context hooks
2. Adding /context hooks enable-all and /context hooks disable-all commands
3. Adding UI feedback for hook execution, including timing information
4. Clarifying hook persistence behavior and interaction with /clear
5. Specifying where hook context appears in the conversation

rfcs/0003-context-hooks.md
jsamuel1: Lets propose we implement inline first, and evaluate?

rfcs/0003-context-hooks.md
jsamuel1: In what way can they be less verbose?

rfcs/0003-context-hooks.md
jsamuel1: OK.  Will update to say that they are evaluated in order of:
1) Per-Conversation hooks, in order of appearance in file (once)
then
2) Per-Prompt hooks, in order of appearance in file (every prompt)

rfcs/0003-context-hooks.md
jsamuel1: Yes.  I'll add the `add`.

rfcs/0003-context-hooks.md
jsamuel1: That was my intention with the statement.  I'll update.

rfcs/0003-context-hooks.md
jsamuel1: Configuration should be with the hooks in the json config (context.json / global_context.json ).

rfcs/0003-context-hooks.md
jsamuel1: Unless we can use a single-line view of the execution?

eg.
`hooks: . . . . . completed 5 hooks in 1.2s`

rfcs/0003-context-hooks.md
hayemaxi: For the session makes sense to me. Users can change the config to disable it until re-enabled.

rfcs/0003-context-hooks.md
hayemaxi: Extending off your UI suggestion:

less verbose:
```
hooks: . . . . . completed 5 hooks in 1.2s, 2 hooks failed: company_rules, project_status

> 
```

after `/context show --always`
```
hooks: . . . . . completed 5 hooks in 1.2s, 2 hooks failed:

company_rules:
<Err data>

project_status:
<Err data>

>
```

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,997 @@
...
+1. Users will continue to manually add context, leading to inconsistent usage
+2. Integration with external tools and systems will remain limited
+3. Automated workflows involving Amazon Q CLI will be harder to implement
+4. The CLI will be less powerful for advanced users and organizations
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+1. **Hook Ordering**: Hooks are evaluated in the following order:
```
hayemaxi, line 937: Since hooks run in parallel, they won't complete in a defined order. Should we say that hook results will be added in the following order instead?

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,997 @@
...
+      }
+    ]
+  }
+}
+```
+
+Where:
+- `!` prefix indicates a shell command
+- `/` prefix indicates an internal command
+- `#` prefix indicates a tool call
```
hayemaxi, line 139: What about just making hook types for these instead of trying to interpret special characters?
```
- InlineShellHook
- CommandHook
- ToolHook
```

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,997 @@
...
+        "enabled": true
+      }
+    ]
+  }
+}
+```
+
+### Simplified Command Syntax
+
+For ease of use, hooks can also be configured using a simplified prefix notation in the command field:
```
hayemaxi, line 105: Let's avoid adding complexity for both the user's understanding and the code by just enforcing a "type" field.

rfcs/0003-context-hooks.md
```diff
@@ -0,0 +1,997 @@
...
+```json
+{
+  "context": {
+    "hooks": {
+      "conversation_start": [
+        {
+          "name": "project-info",
+          "type": "script",
+          "path": "~/.aws/amazonq/hooks/project-info.sh",
+          "enabled": true
```
mortizbey, line 60: I recommend changing this to `"disabled": false` to be more in line with how MFC Clients config json are doing it:

Example:
```
"awslabs.nova-canvas-mcp-server": {
      "command": "uvx",
      "args": [
        "awslabs.nova-canvas-mcp-server@latest"
      ],
      "env": {
        "SHELL": "/usr/bin/zsh",
        "FASTMCP_LOG_LEVEL": "ERROR",
        "AWS_PROFILE": "MCP_NOVA_CANVAS_PROFILE"
      },
      "disabled": false,
      "autoApprove": [
        "generate_image"
      ]
    }
    ```

File: rfcs/0003-context-hooks.md [hunk 17]
```diff
@@ -0,0 +1,997 @@
...
+            "Authorization": "Bearer ${CONTEXT_API_TOKEN}"
+          },
+          "enabled": true
+        }
+      ],
+      "per_prompt": [
+        {
+          "name": "environment-vars",
+          "type": "inline",
+          "command": "env | grep AWS_",
```
mortizbey, line 76: Here I propose changing this (or maybe adding a new `command` type is better) to the following for the same reason as before (MCP convention).

```
"per_prompt": [
 {
   "name": "environment-vars",
   "type": "command",
   "command": "env",
   "args": ["|", "grep", "AWS_"],
   "enabled": false
  }
  ```

File: rfcs/0003-context-hooks.md (hunk 10 above)
Comment (mortizbey) [reply rfcs/0003-context-hooks.md:940]: Here I recommend to not reinvent the wheel and just add a new subsection in the new `/context show` UI proposed here: https://github.com/aws/amazon-q-developer-cli/pull/1147.

**Comment by hayemaxi** [comment]
We will need to enforce uniqueness on the hook name across `per_prompt` and `conversation_start` categories. E.g. if a user calls `/context hooks disable test`, and there are multiple hooks named `test` in either of the categories, they are all disabled and a warning message is displayed about duplicate hooks.

Otherwise, the user will have to specify the category each time, which would be very verbose and complicate the logic. E.g.
`/context hooks disable test --global --type=per-prompt`. We would also need separate caches for each of the different categories, and we would also need to let the user know that with any duplicate named hooks, only the result of the first one is returned.

File: rfcs/0003-context-hooks.md (hunk 13 above)
Comment (jsamuel1) [reply rfcs/0003-context-hooks.md:937]: Yes, it is stable ordering of the results that matter.   We should make it clear to people in our documentation that hooks should not have dependencies on each other.

File: rfcs/0003-context-hooks.md (hunk 14 above)
Comment (jsamuel1) [reply rfcs/0003-context-hooks.md:139]: Note -- the above special characters (other than #)  are already in the CLI parser -- using those just means you pass it to the parser and execute -> so less code.   If we added # for a tool call (or something like that) -> it should be for regular prompt too.

File: rfcs/0003-context-hooks.md (hunk 16 above)
Comment (jsamuel1) [reply rfcs/0003-context-hooks.md:60]: Thanks.  Good pickup.

File: rfcs/0003-context-hooks.md (hunk 17 above)
Comment (jsamuel1) [reply rfcs/0003-context-hooks.md:76]: Do we support either/or, or is it just the one.

Note -- `env | grep AWS_` is different to `env "|" "grep" "AWS_"` 
So the second version may not work if we use execve for the command, rather than shell -c .

**Comment by jsamuel1** [comment]
> We will need to enforce uniqueness on the hook name across `per_prompt` and `conversation_start` categories. E.g. if a user calls `/context hooks disable test`, and there are multiple hooks named `test` in either the same category or between the 2 cartegories, they are all disabled and a warning message is displayed about duplicate hooks.
> 
> Otherwise, the user will have to specify the category each time, which would be very verbose and complicate the logic. E.g. to disable in the global hooks: `/context hooks disable test --global --type=per-prompt`. We would also need separate caches for each of the different categories, and we would also need to let the user know that with any duplicate named hooks, only the result of the first one is returned.

Agree.   Or we use a convention for the naming when disambiguation is needed, eg.

`/context hooks disable global.test` would only disable a "test" if it is "global" scope, whereas without the qualifier, it wouldn't care.

What are the downsides to enforcing uniqueness?
If this all in the one config file, we can warn on startup, and tell them which one is disabled.  OR we can warn on startup and rename one of them.   "test (1)"

**Comment by hayemaxi** [comment]
> What are the downsides to enforcing uniqueness? If this all in the one config file, we can warn on startup, and tell them which one is disabled. OR we can warn on startup and rename one of them. "test (1)"

This increased complexity, but it is better than not enforcing uniqueness. If we don't then it will be hard for users to specify which hook to remove/disable/enable.

As an aside, we can simplify both the code and the UX by adjusting the schema. What do you think about getting rid of `conversation_start` and `per_prompt` and instead adding a `when` field:
```
{
    "hooks": [
         {
             "name": "environment-vars",
             "type": "command",
             "command": "env",
             "args": ["|", "grep", "AWS_"],
             "disalbed": false,
             "when": "conversation_start"
             // or
              "when": "per_prompt"
            }
     ]
}
```

**Comment by hayemaxi** [comment]
The initial implementation has been merged in the referenced PRs above. These changes from the RFC that were implemented should be incorporated into this RFC so we can merge it, and all un-implemented content that remains should be moved to a new RFC.

Implementation differences:

### Overview
The implementation fulfills the desire of the RFC to allow dynamic context to be created via automatically executed shell commands. This allows the user to call APIs, perform calculations, or run commands like "git status" as the context for Amazon Q.

### Configuration
Triggers, which are for the most part consistent with the RFC:

- Conversation Start Hooks: Run once and the results are attached directly to the top of the conversation history along with other context (e.g. summary context or file context)
- Per-Prompt Hooks: Run on each user prompt (unless a cache time is set in the JSON config), and are included directly with the user's message sent to Q. Unlike the user message it is attached to, it is not recorded in conversation history.

Consistent with the RFC, context hooks are defined in the same configurations that hold context files. Context hooks are instead defined in a single JSON key rather than separate categories `conversation_star: []` and `per_prompt: []`. Additionally, there is no `criticality` field as mentioned in the RFC. These changes were made to reduce complexity during implementation. The name of the hook must be unique.

```
{
  "paths": [],
  "hooks": {
    "pirate_talk": {                             // The name must be unique
      "trigger": "per_prompt",                   // This field determines per_prompt or conversation_start
      "type": "inline",                          // There is only 1 type for now
      "disabled": true,                          // To disable or not
      "timeout_ms": 30000,                       // When to stop running the hook if its taking too long (in milliseconds)
      "max_output_size": 10240,                  // Max amount of output characters (bytes) to include in the context. 
                                                 // Anything more is truncated
      "cache_ttl_seconds": 0,                    // How long to keep the hooks results in memory before running it again. Does 
                                                 // not change when the results are added to the context
      "command": "echo \"Talk like a pirate\""   // The shell command to execute
    }
  }
}
```

### Simplified Command Syntax
Not implemented due to unnecessary complexity right now. This may make it too difficult to understand for users, so it should be re-evaluated before implementation is considered.


### Hook Types
Consistent with the RFC, `inline` hooks are the initial type and the only type right now.


### Commands
Commands are implemented according to the RFC with a few minor changes. The output of `/context hooks help` can explain best:
```
(Beta) Context Hooks

Use context hooks to specify shell commands to run. The output from these
commands will be appended to the prompt to Amazon Q. Hooks can be defined
in global or local profiles.

Usage: /context hooks [SUBCOMMAND]

Description
  Show existing global or profile-specific hooks.
  Alternatively, specify a subcommand to modify the hooks.

Available subcommands
  hooks help                         Show an explanation for context hooks commands

  hooks add [--global] <name>        Add a new command context hook
                                         --global: Add to global hooks
         --trigger <trigger>           When to trigger the hook, valid options: `per_prompt` or `conversation_start`
         --command <command>             Shell command to execute

  hooks rm [--global] <name>         Remove an existing context hook
                                         --global: Remove from global hooks

  hooks enable [--global] <name>     Enable an existing context hook
                                         --global: Enable in global hooks

  hooks disable [--global] <name>    Disable an existing context hook
                                         --global: Disable in global hooks

  hooks enable-all [--global]        Enable all existing context hooks
                                         --global: Enable all in global hooks

  hooks disable-all [--global]       Disable all existing context hooks
                                         --global: Disable all in global hooks

Notes
• Hooks are executed in parallel
• 'conversation_start' hooks run on the first user prompt and are attached once to the conversation history sent to Amazon Q
• 'per_prompt' hooks run on each user prompt and are attached to the prompt, but are not stored in conversation history
```

Notes:
1.
- `/context show --hooks` - Show only hook contributions
- `/context show --previous` - Show context from previous prompt
- `/context show --always` - Always show context after each prompt

are **not** implemented. To view hooks use `/context hooks`. These were omitted to reduce complexity of the initial implementation, but would absolutely be useful as a follow up.

2. The user can configure the hook name, command, and trigger on CLI only. They will have to edit the context configuration JSON file to add a cache time or max output size.

jsamuel1:
What do we do with RFC PR's for features when done?

chaynabors:
Context hooks were added. We have a new RFC repo to formalize this process and ensure RFC's are getting merged before implementation work happens. Closing in favor of the new repo. It will be public soon.

PR #1054, opened 2025-04-02
feat(commands): granular permissions with /tools

hayemaxi:
Implements https://github.com/aws/amazon-q-developer-cli/pull/921 which describes trusting and untrusting specific tools for the current session.

- Start with default permissions. Users can change them from command line or within chat via /tools
- `/acceptall` and `--accept-all` were deprecated in favor of `--trust-all-tools` and `/tools trustall`. They will continue to work, but display a notice and activation functionality of the new commands.
  - Command::AcceptAll no longer exists
- UI has been reword a little, see screenshots.
- `report_issue` will include trust override settings in the report.
- Chat supports multiple tool use requests from Q at once. I have refactored the tool flow to allow asking permissions for individual tool requests at a time
  - Now, ExecuteTools (checks for acceptance one tool at a time) -> PromptUser (ask for acceptance) -> HandleInput (handle acceptance) -> ExecuteTools (find next tool that needs acceptance OR trigger execution if none remaining).
  - I was not able to find a prompt that made Q send multiple tool_uses in a single request coming from Q. It seems to do them iteratively even when asked to do them in parallel.
  - In any case, that original logic to support parallel requests is preserved with this new structure.

Risks: Likely conflicts with https://github.com/aws/amazon-q-developer-cli/pull/999

image

image

image
image
image
image

### DIFF
--- a/crates/q_cli/src/cli/chat/conversation_state.rs
+++ b/crates/q_cli/src/cli/chat/conversation_state.rs
@@ -33,7 +33,10 @@ use tracing::{
 };
 
 use super::context::ContextManager;
-use super::tools::ToolSpec;
+use super::tools::{
+    QueuedTool,
+    ToolSpec,
+};
 use super::truncate_safe;
 use crate::cli::chat::tools::{
     InputSchema,
@@ -64,7 +67,7 @@ pub struct ConversationState {
     /// e.g user messages prefixed with '> '. Should also be used to store errors posted in the
     /// chat.
     pub transcript: VecDeque<String>,
-    tools: Vec<Tool>,
+    pub tools: Vec<Tool>,
     /// Context manager for handling sticky context files
     pub context_manager: Option<ContextManager>,
     /// Cached value representing the length of the user context message.
@@ -350,12 +353,12 @@ impl ConversationState {
     }
 
     /// Sets the next user message with "cancelled" tool results.
-    pub fn abandon_tool_use(&mut self, tools_to_be_abandoned: Vec<(String, super::tools::Tool)>, deny_input: String) {
+    pub fn abandon_tool_use(&mut self, tools_to_be_abandoned: Vec<QueuedTool>, deny_input: String) {
         debug_assert!(self.next_message.is_none());
         let tool_results = tools_to_be_abandoned
             .into_iter()
-            .map(|(tool_use_id, _)| ToolResult {
-                tool_use_id,
+            .map(|tool| ToolResult {
+                tool_use_id: tool.id,
                 content: vec![ToolResultContentBlock::Text(
                     "Tool use was cancelled by the user".to_string(),
                 )],
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -7,7 +7,10 @@ mod parser;
 mod prompt;
 mod tools;
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::{
+    HashMap,
+    HashSet,
+};
 use std::io::{
     IsTerminal,
     Read,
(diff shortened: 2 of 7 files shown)


crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -350,12 +353,12 @@ impl ConversationState {
     }
 
     /// Sets the next user message with "cancelled" tool results.
-    pub fn abandon_tool_use(&mut self, tools_to_be_abandoned: Vec<(String, super::tools::Tool)>, deny_input: String) {
+    pub fn abandon_tool_use(&mut self, tools_to_be_abandoned: Vec<QueuedTool>, deny_input: String) {
```
dingfeli, line 356: If the user accepts a partial list of tools, and without running it, interrupts it the workflow. 
Do we now accept the permission granted in this interaction? Or do we revoke the permissions granted? 
Right now it looks like what's granted would be remembered without running as well.

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1007,30 +1073,177 @@ where
...
+                            style::SetForegroundColor(Color::Reset),
+                        )?;
+                    },
+                    Some(ToolsSubcommand::TrustAll) => {
+                        self.conversation_state
+                            .tools
+                            .iter()
+                            .for_each(|FigTool::ToolSpecification(spec)| {
+                                self.tool_permissions.trust_tool(spec.name.as_str());
+                            });
```
dingfeli, line 1111: Is there an advantage to explicitly marking every tool as trusted as opposed to checking a flag. 
i.e. is there a source of truth that will allow us to know if trustall has been enabled? 
If someone is to add a new tool during a session, is the expectation here that the new tool is trusted? Or do we need to prompt the user again?

Not a big deal now though. We can always change this later.

dingfeli (approved):
Seems mostly aligned with the RFC referenced. 
Implementation seems sound to me. 
Feel free to request reviews from people who are involved in the RFC discussion.

crates/q_cli/src/cli/chat/tools/mod.rs
```diff
@@ -142,6 +143,78 @@ impl TryFrom<ToolUse> for Tool {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ToolPermission {
+    pub trusted: bool,
+}
```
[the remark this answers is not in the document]
hayemaxi: Yes, instead of blanket trust bool, we can expand this to account for things like sub commands and directories.

crates/q_cli/src/cli/chat/conversation_state.rs
hayemaxi: Yes, for `trust` only. Future calls on accepted tools will still request it. IMO this is desirable because the user is declaring that they trust the tool regardless of future attempts. Maybe they trust fs_writes, but once they see that it will run execute_bash, they cancel. That doesn't mean that that they don't trust fs_writes afterall.

dingfeli: That makes sense. Just wanted to make sure. The tool level permission should be largely context free.

crates/q_cli/src/cli/chat/mod.rs
hayemaxi: > Is there an advantage to explicitly marking every tool as trusted as opposed to checking a flag.
i.e. is there a source of truth that will allow us to know if trustall has been enabled?

We still need it for non-trust-all instances, but yes we could skip iteration of we had a trust-all flag. I didn't add this because:

Most if not all cases the there will only be one tool to iterate for. We can remove complexity of maintaining a trust-all flag since we already have the iteration code.
It doesn't seem that we can add tools mid-session at this time, but if we could in the future, we may want to re-evaluate trustall for that tool.

> If someone is to add a new tool during a session, is the expectation here that the new tool is trusted? Or do we need to prompt the user again?

Currently the expectation is that the user needs to trust the new tool once it is added. I am open to discussing this, but it makes sense to me. Especially if the tool is added via some automation.

Issue #1028, opened 2025-03-31
Container based execution environment documentation


jsamuel1:
For customers that require containers/docker for their execution environment, document steps to build and configure a docker container for use with your source repository and/or aws environment.

PR #1063, opened 2025-04-03
feat(docker) Added docs/docker with instructions for running q cli in a container

jsamuel1:
*Issue #, if available:*
#1028 
*Description of changes:*
Added documentation and a script/alias for running q cli i a container.
The script will add an alias to the user .bashrc/.zshrc for a new alias `q-dev`, which launches a container with the current folder mapped to ~/src/ as read-write and with aws, ssh and git config mounted readonly.


JoeShi:
Can we merge this? Many customers would like to know how to run it in the docker container.

chaynabors:
I don't think we want to maintain a dockerfile for this. Should be simple enough for someone to run a simple template with standalone chat installed soon. I'm casting a large net as I close PR's for the refactor so DM me if you think we should still do this.

Issue #1071, opened 2025-04-03
[UX] Impossible to cancel Thinking with ^C


alanmcco:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.3.2 (24D81)

### Expected behaviour

When Q is "Thinking...." interrupting it with CONTROL+C will cancel its thinking immediately and allow me to send a different prompt. 

### Actual behaviour

I prompted Q to format some text for me into markdown, but I forgot to paste the text. Q started "Thinking...". I immediately tried to cancel with CONTROL+C and I witnessed the ^C interrupt get printed to the shell. However, Q repeated "Thinking..." and was impossible to interrupt for minutes. 

<img width="1307" alt="Image" src="https://github.com/user-attachments/assets/7552bef7-d94e-415a-b10c-b9c8a87c2eff" />

<img width="1307" alt="Image" src="https://github.com/user-attachments/assets/1982b4cc-21df-4785-a6dc-92a4dc21e043" />

### Steps to reproduce

Yes. I was able to reproduce the problem in the same session by sending in another prompt. I started a new chat and I was able to reproduce it, but it canceled it more quickly.. in the new session it took a few seconds before it canceled (I pressed ^C about 5 times fairly quickly) in the other session which was longer (perhaps more context) it took a very long time to cancel (I believe more than a minute). 

### Environment

```yaml
[q-details]
version = "1.7.2"
hash = "b6bf5f1675dd7395009e8bc65c1145820671f211"
date = "2025-03-21T01:36:16.66764Z (14d ago)"
variant = "full"

[system-info]
os = "macOS 15.3.2 (24D81)"
chip = "Apple M3 Pro"
total-cores = 12
memory = "36.00 GB"

[environment]
cwd = "/Users/USER/Documents/Obsidian Vault"
cli-path = "/Users/USER/Documents/Obsidian Vault"
os = "Mac"
shell-path = "/bin/zsh"
shell-version = "5.9"
terminal = "macOS"
install-method = "unknown"

[env-vars]
[...]

PR #1072, opened 2025-04-03
fix(chat ux): exit thinking state using ctrl + c

GoodluckH:
*Issue #, if available:*
#1071 

https://github.com/user-attachments/assets/e702d6a6-5c23-473f-8f95-d58e5a835d77

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -417,7 +417,13 @@ where
                     }
                     self.prompt_user(tool_uses, skip_printing_tools).await
                 },
-                ChatState::HandleInput { input, tool_uses } => self.handle_input(input, tool_uses).await,
+                ChatState::HandleInput { input, tool_uses } => {
+                    let tool_uses_clone = tool_uses.clone().unwrap();
+                    tokio::select! {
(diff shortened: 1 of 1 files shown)


mschrage (approved):

dingfeli:
This causes a panic with initial input that is not empty.

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -417,7 +417,13 @@ where
                     }
                     self.prompt_user(tool_uses, skip_printing_tools).await
                 },
-                ChatState::HandleInput { input, tool_uses } => self.handle_input(input, tool_uses).await,
+                ChatState::HandleInput { input, tool_uses } => {
+                    let tool_uses_clone = tool_uses.clone().unwrap();
```
dingfeli, line 421: There is no need for this unwrap here. 
We can just pass in the `tool_uses` clone directly to tool_uses in the `Interrupt` variant of the enum, which already accepts an option.

GoodluckH: Great catch! Fixed here: https://github.com/aws/amazon-q-developer-cli/pull/1092

Issue #1087, opened 2025-04-04
[bug][parent] Unable to show Regex in plain text mode


GoodluckH:
This is the parent issue to organize all issues reporting a bug where REGEX chars won't display if they're not wrapped in with backquotes.

It's possibly due to terminal intercepting and escaping those as ANSI codes 

<img width="660" alt="Image" src="https://github.com/user-attachments/assets/bcc2a43b-ccf6-4036-959a-3a79c2037fff" />

PR #971, opened 2025-03-27
fix(chat): Fix display of square brackets in chat output

ymhiroki:
*Issue #, if available:* #889 

*Description of changes:*

Modify URL parser to properly handle square brackets in text by saving the input position and resetting it when the complete URL pattern does not match. This ensures square brackets are correctly displayed in chat output when they are not part of a markdown link.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1090, opened 2025-04-04
fix(chat): diplay special chars properly and blockquotes formatting

GoodluckH:
*Issue #, if available:*
#1087 

Before:

Screenshot 2025-04-04 at 1 34 04 PM

After:
Screenshot 2025-04-04 at 1 33 44 PM


crates/q_cli/src/cli/chat/parse.rs
```diff
@@ -138,9 +138,6 @@ pub fn interpret_markdown<'a, 'b>(
     match state.in_codeblock {
         false => {
             stateful_alt!(
-                // This pattern acts as a short circuit for alphanumeric plaintext
```
brandonskiser, line 141: You have a better idea of what this comment is referring to? Considering it's just being deleted and moved around

crates/q_cli/src/cli/chat/parse.rs
```diff
@@ -743,7 +742,7 @@ mod tests {
     validate!(bulleted_item_1, "- bullet", [style::Print("• bullet")]);
     validate!(bulleted_item_2, "* bullet", [style::Print("• bullet")]);
     validate!(numbered_item_1, "1. number", [style::Print("1. number")]);
-    validate!(blockquote_1, "&gt; hello", [
```
brandonskiser, line 746: Is this being updated because the model response never has `&gt;` and instead `>` directly? I guess if you haven't noticed anything break in your manual testing then it's probably fine

brandonskiser (approved):

crates/q_cli/src/cli/chat/parse.rs
GoodluckH: Yes, it appears that our model is responding back in MKD format. Even in rare case it sends back `&gt;`, it won't break anything, it'll just show ">" like what it's currently formatted for blockquotes.

crates/q_cli/src/cli/chat/parse.rs
GoodluckH: Actually I just moved it back. I think someone else already pushed a fix for displaying square brackets. https://github.com/aws/amazon-q-developer-cli/pull/971

crates/q_cli/src/cli/chat/parse.rs
brandonskiser: It looks like we have code that handles `&gt;` specifically, in which case I would prefer we don't remove this test case

GoodluckH: Yes there's a test case for it:

```
validate!(greater_than_1, ".&gt;.", [style::Print(".>.")]);
```

Issue #1086, opened 2025-04-04
Add ability to reset permissions for single tools


hayemaxi:
`/tools reset` will reset all tools to default permission levels, which leaves it up to the tool to determine what the default permission is. E.g. execute_bash will only allow read-only commands.

There is no way to reset single tools to default permissions if their trust level was modified by the user manually, e.g. with `/tools trust` or `/tools untrust`

Instead, we can support:
```
/tools reset <tool_name>   # reset a single tool to default permission level
/tools resetall            # reset all tools to default permission levels
```

or if `/tools reset` has been ingrained in user workflows by the time this is implemented, simply
```
/tools reset <tool_name>   # reset a single tool to default permission level
/tools reset               # reset all tools
```

kzhou003:
Hi @hayemaxi @GoodluckH 

If this issue is still open, can I get review/approval for my PR here: https://github.com/aws/amazon-q-developer-cli/pull/1102? 
Please let me know if anything. Thanks.

GoodluckH:
Thanks for contributing! Made a comment

hayemaxi:
We will want to handle these comments from the original PR before closing this out.

> Will this print to the user? IIRC it does not. Instead, you may want to throw an error from this function and catch it when handling ToolsSubcommand::ResetSingle.
...
Should catch here, if not is_ok() then we can print in red: Cannot reset tool '{tool}', it does not exist.

GoodluckH:
I closed this too quick lol. I can pick this up!

PR #1102, opened 2025-04-07
feat: enhance tools reset

kzhou003:
*Issue #, if available:* https://github.com/aws/amazon-q-developer-cli/issues/1086

*Description of changes:* 

1. Added a new `ResetSingle` variant to the `ToolsSubcommand` enum in `crates/q_cli/src/cli/chat/command.rs`
2. Updated the `AVAILABLE_COMMANDS` constant to include documentation for the new reset single tool command
Added handling in the command parser to support both `/tools reset (reset all)` and `/tools reset <tool_name>` (reset single)
Added a new `reset_tool` method to the `ToolPermissions` struct in `crates/q_cli/src/cli/chat/tools/mod.rs`
5. Added command handling for the `ResetSingle` variant in the Chat state machine in `crates/q_cli/src/cli/chat/mod.rs`

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -1244,6 +1244,15 @@ where
                             style::SetForegroundColor(Color::Reset),
                         )?;
                     },
+                    Some(ToolsSubcommand::ResetSingle { tool_name }) => {
+                        self.tool_permissions.reset_tool(&tool_name);
+                        queue!(
+                            self.output,
+                            style::SetForegroundColor(Color::Green),
+                            style::Print(format!("\nReset tool '{}' to the default permission level.", tool_name)),
+                            style::SetForegroundColor(Color::Reset),
+                        )?;
+                    },
                     Some(ToolsSubcommand::Help) => {
                         queue!(
                             self.output,
--- a/crates/q_cli/src/cli/chat/tools/mod.rs
+++ b/crates/q_cli/src/cli/chat/tools/mod.rs
@@ -27,6 +27,7 @@ use fs_read::FsRead;
 use fs_write::FsWrite;
 use gh_issue::GhIssue;
 use serde::Deserialize;
+use tracing::warn;
 use use_aws::UseAws;
 
 use super::parser::ToolUse;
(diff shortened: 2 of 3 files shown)


crates/q_cli/src/cli/chat/tools/mod.rs
```diff
@@ -194,6 +194,10 @@ impl ToolPermissions {
         self.permissions.clear();
     }
 
+    pub fn reset_tool(&mut self, tool_name: &str) {
+        self.permissions.remove(tool_name);
```
GoodluckH, line 198: What happens if it's an invalid tool name?

kzhou003: Thank you for reviewing! 

Added a warning logging message when the tool name is invalid for better user feedback. 

Based on https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove, it seems it is also safe to remove from hashmap when the map does not contain the key.

GoodluckH (approved):

GoodluckH:
Thank you for your contribution!

kzhou003:
@GoodluckH Thank you for approving! 
What is the next step to merge it? I do not see merge button here and I dont find instructions for that in contribution guide.

crates/q_cli/src/cli/chat/tools/mod.rs
```diff
@@ -194,6 +195,14 @@ impl ToolPermissions {
         self.permissions.clear();
     }
 
+    pub fn reset_tool(&mut self, tool_name: &str) {
+        if !self.permissions.contains_key(tool_name) {
+            warn!("No custom permissions set for tool '{tool_name}' to reset");
```
hayemaxi, line 200: Will this print to the user? IIRC it does not. Instead, you may want to throw an error from this function and catch it when handling `ToolsSubcommand::ResetSingle`. See other comment

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1482,6 +1482,15 @@ where
...
                         )?;
                     },
+                    Some(ToolsSubcommand::ResetSingle { tool_name }) => {
+                        self.tool_permissions.reset_tool(&tool_name);
+                        queue!(
+                            self.output,
+                            style::SetForegroundColor(Color::Green),
+                            style::Print(format!("\nReset tool '{}' to the default permission level.", tool_name)),
+                            style::SetForegroundColor(Color::Reset),
+                        )?;
```
hayemaxi, line 1492: (see other comment). Should catch here, if not `is_ok()` then we can print in red: `Cannot reset tool '{tool}', it does not exist.`

kzhou003: @hayemaxi Makes sense to me. What is the best way to amend this? I will add this change in my next PR if it sounds good to you.

PR #1127, opened 2025-04-07
fix(chat): Safeguard against mutative find cli invocations

brandonskiser:
*Issue #, if available:*

*Description of changes:*


mschrage:
Should we consider adding a dedicated `find` tool?

mschrage (approved):

brandonskiser:
> Should we consider adding a dedicated `find` tool?

Probably not for now since this is more easily composable with other execute bash commands, plus all the overhead of having our own find implementation - we could instead improve the fs_read tool search functionality to have more-or-less feature parity

Issue #1067, opened 2025-04-03
Initial implementation of `/compact` command for conversation summarization


aseemxs:
Create the core functionality for the `/compact` command that allows users to explicitly request conversation summarization. This includes command parsing, summary generation logic, and conversation state management. The implementation will focus on a straightforward approach where the history is cleared and replaced with an AI-generated summary of the previous conversation.

jsamuel1:
Ideally, we should auto-compact as we progress, including in the background whilst waiting for the next human prompt.

Compaction should include both LLM and robotic reduction of context size.

Reductions could include:
* Deduplication of read/write/read loops to have one latest copy of an object in context (perhaps with an LLM summarized history?) - files, profile contexts, other object types.
* Latest version only of directory tree information
* Summarization of Debug Loops, at the end of any debug cycle, or when output from a tool is more than X big, always summarize it early, once the first action has been performed on the output.
* Summarization of a series of human prompts into an intent and history of tasks / *Journal* .  <"Hey Q, what have I done so far?" could use the journal information>

jsamuel1:
Manual use of /compact could be the equivalent of "Remember all this and give me a summary".   Most of the time, manual use of this command should be treated as an indicator of a bug.

PR #1128, opened 2025-04-07
feat: Introduce /compact command

aseemxs:
*Description of changes:*
- Add functionality to summarize and clear conversation history
- Allow presenting the summary with --summary flag
- Let /compact take an optional prompt to modify summary request
- Ensure the model acknowledges the summary in future responses
- Modify clear() method to optionally preserve summary
- Warn users about context limit based on number of characters

[see issue : 1067](https://github.com/aws/amazon-q-developer-cli/issues/1067)

### DIFF
--- a/crates/q_cli/src/cli/chat/chat_state.rs
+++ b/crates/q_cli/src/cli/chat/chat_state.rs
@@ -1,4 +1,5 @@
 use std::collections::VecDeque;
+
 use fig_api_client::model::ChatMessage;
 
 /// Character count warning levels for conversation size
@@ -33,7 +34,7 @@ impl SummarizationState {
             show_summary: false,
         }
     }
-    
+
     pub fn with_prompt(prompt: Option<String>) -> Self {
         Self {
             original_history: None,
--- a/crates/q_cli/src/cli/chat/chat_state.rs
+++ b/crates/q_cli/src/cli/chat/chat_state.rs
@@ -1,0 +1,44 @@
+use std::collections::VecDeque;
+use fig_api_client::model::ChatMessage;
+
+/// Character count warning levels for conversation size
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TokenWarningLevel {
+    /// No warning, conversation is within normal limits
+    None,
+    /// Critical level - at single warning threshold (500K characters)
+    Critical,
+}
+
+/// Constants for character-based warning threshold
+pub const MAX_CHARS: usize = 500000; // Character-based warning threshold
+
+/// State for tracking summarization process
+#[derive(Debug, Clone)]
+pub struct SummarizationState {
+    /// The saved original history
+    pub original_history: Option<VecDeque<ChatMessage>>,
+    /// Optional custom prompt used for summarization
+    pub custom_prompt: Option<String>,
+    /// Whether to show the summary after compacting
+    pub show_summary: bool,
+}
+
+impl SummarizationState {
+    #[allow(dead_code)]
+    pub fn new() -> Self {
+        Self {
+            original_history: None,
+            custom_prompt: None,
+            show_summary: false,
+        }
+    }
+    
+    pub fn with_prompt(prompt: Option<String>) -> Self {
+        Self {
+            original_history: None,
+            custom_prompt: prompt,
+            show_summary: false,
+        }
+    }
+}
--- a/crates/q_cli/src/cli/chat/conversation_state.rs
+++ b/crates/q_cli/src/cli/chat/conversation_state.rs
@@ -32,6 +32,7 @@ use tracing::{
     warn,
 };
 
+use super::chat_state::{MAX_CHARS, TokenWarningLevel};
 use super::context::ContextManager;
 use super::tools::{
     QueuedTool,
@@ -72,6 +73,8 @@ pub struct ConversationState {
     pub context_manager: Option<ContextManager>,
     /// Cached value representing the length of the user context message.
     context_message_length: Option<usize>,
+    /// Stores the latest conversation summary created by /compact
+    pub latest_summary: Option<String>,
 }
 
 impl ConversationState {
@@ -113,13 +116,27 @@ impl ConversationState {
                 .collect(),
             context_manager,
             context_message_length: None,
+            latest_summary: None,
         }
     }
 
-    /// Clears the conversation history.
-    pub fn clear(&mut self) {
+    /// Returns a vector representation of the history (for accessors)
+    pub fn history_as_vec(&self) -> Vec<ChatMessage> {
+        self.history.iter().cloned().collect()
+    }
+    
+    /// Returns the length of the conversation history
+    pub fn history_len(&self) -> usize {
+        self.history.len()
+    }
+    
+    /// Clears the conversation history and optionally the summary.
+    pub fn clear(&mut self, preserve_summary: bool) {
         self.next_message = None;
         self.history.clear();
+        if !preserve_summary {
+            self.latest_summary = None;
+        }
     }
 
     pub async fn append_new_user_message(&mut self, input: String) {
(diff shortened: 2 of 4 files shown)


crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -62,6 +68,36 @@ use fig_api_client::model::{
...
 use fig_util::CLI_BINARY_NAME;
+
+/// Help text for the compact command
+fn compact_help_text() -> String {
+    color_print::cformat!(
+        r#"
+<magenta,em>Conversation Compaction</magenta,em>
+
+The <em>/compact</em> command summarizes the conversation history to free up context space
+while preserving the essential information. This is useful for long-running conversations
```
brandonskiser, line 79: "while preserving essential information"

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -472,6 +502,67 @@ impl ConversationState {
...
+                            for result in results {
+                                for content in &result.content {
+                                    match content {
+                                        ToolResultContentBlock::Text(text) => total_chars += text.len(),
+                                        ToolResultContentBlock::Json(doc) => {
+                                            if let Some(s) = doc.as_string() {
+                                                total_chars += s.len();
+                                            } else {
+                                                // Approximate JSON size
+                                                total_chars += 100;
```
mschrage, line 527: When would this happen?

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -6,6 +7,11 @@ mod parse;
 mod parser;
 mod prompt;
 mod tools;
+
+// Re-export types for use in this module
```
brandonskiser, line 11: Remove this comment

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -472,6 +502,67 @@ impl ConversationState {
...
+                            }
+                        }
+                    }
+                },
+                ChatMessage::AssistantResponseMessage(msg) => {
+                    total_chars += msg.content.len();
+                    // Add tool uses if any
+                    if let Some(tool_uses) = &msg.tool_uses {
+                        // Approximation for tool uses
+                        total_chars += tool_uses.len() * 200;
```
mschrage, line 541: why are we approximating like this?

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -133,6 +170,10 @@ const HELP_TEXT: &str = color_print::cstr! {"
 <em>/editor</em>       <black!>Open $EDITOR (defaults to vi) to compose a prompt</black!>
 <em>/help</em>         <black!>Show this help dialogue</black!>
 <em>/quit</em>         <black!>Quit the application</black!>
+<em>/compact</em>      <black!>Summarize conversation to free up context space</black!>
+  <em>help</em>        <black!>Show help for the compact command</black!>
+  <em>[prompt]</em>    <black!>Optional custom prompt to guide summarization</black!>
+  <em>--summary</em>   <black!>Display the summary after compacting</black!>
```
mschrage, line 176: This flags should not be added at the top level help text.

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -789,21 +841,132 @@ where
...
+                    style::Print("\nCompacting conversation history...\n"),
+                    style::SetForegroundColor(Color::Reset)
+                )?;
+                
+                // Create a summary request based on user input or default
+                let summary_request = match prompt {
+                    Some(custom_prompt) => {
+                        // Make the custom instructions much more prominent and directive
+                        format!(
+                            "[SYSTEM NOTE: This is an automated summarization request, not from the user]\n\n\
```
mschrage, line 915: Can we pull prompts out into a separate file?

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1611,7 +1781,8 @@ where
                     )
                     .await;
                 }
-                if self.interactive {
+                // Handle citations for non-summarization responses
+                if self.interactive && !is_summarization {
                     queue!(self.output, style::ResetColor, style::SetAttribute(Attribute::Reset))?;
```
mschrage, line 1786: Why are we special casing this - shouldn't the markdown be rendered the same?

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1632,6 +1803,94 @@ where
...
+            
+            // Add the message
+            self.conversation_state.push_assistant_message(special_message);
+            
+            // Display confirmation message
+                
+            execute!(
+                self.output,
+                style::SetForegroundColor(Color::Green),
+                style::Print("\n✅ Conversation has been successfully summarized and cleared!\n\n"),
```
mschrage, line 1836:
```suggestion
                style::Print("\nConversation history has been compacted successfully!\n\n"),
```

Can we use a checkmark character with the color green rather than an emoji?

mschrage (commented):
Overall, looks good! I will defer to @brandonskiser but had a few comments and nits. Consider them guidance/clarification rather than required changes.

crates/q_cli/src/cli/chat/chat_state.rs
```diff
@@ -0,0 +1,44 @@
+use std::collections::VecDeque;
```
brandonskiser, line 1: This file should be called `summarization_state` instead

crates/q_cli/src/cli/chat/chat_state.rs
```diff
@@ -0,0 +1,44 @@
...
+    /// The saved original history
+    pub original_history: Option<VecDeque<ChatMessage>>,
+    /// Optional custom prompt used for summarization
+    pub custom_prompt: Option<String>,
+    /// Whether to show the summary after compacting
+    pub show_summary: bool,
+}
+
+impl SummarizationState {
+    #[allow(dead_code)]
```
brandonskiser, line 28: why dead code?

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -133,6 +170,10 @@ const HELP_TEXT: &str = color_print::cstr! {"
 <em>/editor</em>       <black!>Open $EDITOR (defaults to vi) to compose a prompt</black!>
 <em>/help</em>         <black!>Show this help dialogue</black!>
 <em>/quit</em>         <black!>Quit the application</black!>
+<em>/compact</em>      <black!>Summarize conversation to free up context space</black!>
```
brandonskiser, line 173: "Summarize the conversation"

crates/q_cli/src/cli/chat/mod.rs
brandonskiser: Agreed, something like a prompts.rs file that just has consts

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1442,6 +1606,9 @@ where
 
         let mut tool_uses = Vec::new();
         let mut tool_name_being_recvd: Option<String> = None;
+        
+        // Flag to track if we're processing a summarization response
+        let is_summarization = self.summarization_state.is_some();
```
brandonskiser, line 1611: Just a thought, this function has gotten way too complicated handling different types of responses and so on. Wonder if it's cleaner to define a separate state for summaries

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -789,21 +841,132 @@ where
...
+                let saved_history = VecDeque::from(self.conversation_state.history_as_vec().clone());
+                
+                // Set up summarization state with history, custom prompt, and show_summary flag
+                let mut summarization_state = SummarizationState::with_prompt(prompt.clone());
+                summarization_state.original_history = Some(saved_history);
+                summarization_state.show_summary = show_summary; // Store the show_summary flag
+                self.summarization_state = Some(summarization_state);
+                
+                // Tell user we're working on the summary
+                execute!(
```
brandonskiser, line 903: Don't think we need this, instead the spinner should be enough.

(also, would be better to have an extra newline here between this message and the spinner)

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1569,25 +1736,28 @@ where
...
-                        self.output.flush()?;
-                        state.newline = state.set_newline;
-                        state.set_newline = false;
-                    },
-                    Err(err) => match err.into_inner() {
-                        Some(err) => return Err(ChatError::Custom(err.to_string().into())),
-                        None => break, // Data was incomplete
-                    },
+            // For summarization, we capture the summary but don't print it
+            if !is_summarization {
```
brandonskiser, line 1740: Really thinking summarization should really be in another state (maybe like ChatState::SummarizeConversation), and not have extra logic added here.

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -113,13 +116,27 @@ impl ConversationState {
...
             context_manager,
             context_message_length: None,
+            latest_summary: None,
         }
     }
 
-    /// Clears the conversation history.
-    pub fn clear(&mut self) {
+    /// Returns a vector representation of the history (for accessors)
+    pub fn history_as_vec(&self) -> Vec<ChatMessage> {
```
brandonskiser, line 124: Why have this separate function? the only two references either immediately cast it back to a VecDeque, or just get the last element

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1569,25 +1736,28 @@ where
...
-                        offset += parsed.offset_from(&input);
-                        self.output.flush()?;
-                        state.newline = state.set_newline;
-                        state.set_newline = false;
-                    },
-                    Err(err) => match err.into_inner() {
-                        Some(err) => return Err(ChatError::Custom(err.to_string().into())),
-                        None => break, // Data was incomplete
-                    },
+            // For summarization, we capture the summary but don't print it
```
brandonskiser, line 1739: The above if check is dropping the spinner, which causes a long pause. You would need to add an extra special case to not drop the spinner for summarization until we have the entire message.

crates/q_cli/src/cli/chat/conversation_state.rs
brandonskiser: This looks buggy - you need to have smarter logic on how to get the size of a `Document` type, basically recurse through all object fields and get the character size of each field. This is critical for `execute_bash` which returns an object of `stdout` and `stderr` which could be huge.

crates/q_cli/src/cli/chat/conversation_state.rs
brandonskiser: Agreed, don't estimate. Tool uses here can be huge, e.g. for large file writes

brandonskiser (changes requested):

brandonskiser (approved):

crates/q_cli/src/cli/chat/mod.rs
kkashilk: Is the only difference between the two prompts (default/custom) the addition of custom_prompt at the end?

I'd expect custom-prompt to be something the user has complete control over

crates/q_cli/src/cli/chat/conversation_state.rs
```diff
@@ -472,6 +502,67 @@ impl ConversationState {
         self.context_message_length
     }
 
+    /// Calculate the total character count in the conversation
+    pub fn calculate_char_count(&self) -> usize {
```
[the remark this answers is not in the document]
brandonskiser: you'd have to touch every single function that modifies the history, sounds incredibly complicated when doing it this way is simple and fast enough

Issue #1117, opened 2025-04-07
Add "tips" to the CLI masthead


mschrage:
When a user starts a new chat session, cycle though a set of tips that communicate features that are accessible from within q chat.

- "Use /context to add files and folders to your conversation"
- "Use /editor to edit the current prompt in your $EDITOR"

PR #1135, opened 2025-04-08
Feature/terminal notifications

saisrinivas26:
*Issue #, if available:*

*Description of changes:*


brandonskiser:
Should this disableable through a setting, e.g. a `chat.disableNotifications`? (ie the `Settings` type passed with `ChatContext`).

I would personally be kind of annoyed hearing this bell on every response, would be nice to have an escape hatch to disable this.

brandonskiser (approved):

mschrage (changes requested):
Agreed @brandonskiser. This seems like it should probably be behind a setting.

saisrinivas26:
> Agreed @brandonskiser. This seems like it should probably be behind a setting.

I was contemplating about it give lot of users requested this. Sure will update and will set default to true.

GoodluckH:
If it plays sound when terminal is the active application, this is going to be an annoying UX, especially when there's no easy way to figure out how to turn it off.

This would makes sense if you can detect that the terminal running Q is not the active window, and then play the sound. Otherwise, I think it's best to default it to false and let ppl know of this capability through tips (#1117).

> given lot of users requested this

I'm pretty sure they only want to hear the sound when doing other tasks and just wanted to be reminded that the Q job is done. Do we have any evidence showing users want to hear this sound even when they're actively interacting with the CLI?

GoodluckH:
Some future iteration: we can emit notifications: https://developer.apple.com/documentation/usernotifications/sending-push-notifications-using-command-line-tools

saisrinivas26:
> Some future iteration: we can emit push notifications: https://developer.apple.com/documentation/usernotifications/sending-push-notifications-using-command-line-tools

yes this is what will be the follow up

saisrinivas26:
> I'd be EXTREMELY careful with shipping this.
> 
> As a work productivity tool, sound effects are intrusive and should be used to communicate important and urgent stuff (like Slack when you get a message). We risk giving a lousy product impression (like those loud and cheap toys in Walmart) to users if we don't play this card right.
> 
> If Q plays sound even when the terminal is the foreground app, it's going to be an annoying UX that leads to uninstalls.
> 
> This would only make sense if you can detect that the terminal running Q is not the active window, and then play the sound. Even then, it's best to default it to false and let ppl know of this capability through tips (#1117). [Default Bias](https://growth.design/psychology#default-bias) is a powerful double-edge sword and for this case, we're going to see more complaints coming than loves.
> 
> > given lot of users requested this
> 
> I'm pretty sure they only want that reminder when they're _away_ from the active Q terminal that's running some jobs. Do we have any evidence showing users want to hear this sound even when they're actively interacting with the CLI?

Agreed I updated to the default to false. Might be too intrusive

GoodluckH:
> Agreed I updated to the default to false. Might be too intrusive

Thanks! We can also talk more about this in tmrw's standup to see we can figure out something addresses the problem effectively without diluting our UX

mschrage (approved):

PR #1172, opened 2025-04-10
[RFC] Agent hooks feature - Q CLI

akashanand98:
The initial content of this RFC has been auto-generated by Q CLI. We encourage all stakeholders to:

Review and provide feedback
Suggest modifications
Participate in discussions
Build upon the proposed concepts
This document serves as a starting point for collaborative development and improvement of the Agent hooks feature.


agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Defining Agent Hooks
+
+Agent hooks are defined using natural language statements that describe a cause and an effect. The basic syntax is:
+
+```
+q hook create "When [cause], then [effect]"
```
jsamuel1, line 36: The template has made it a commandline command, rather than a q cli /command. :)

Should it be:
`/hook create "When [cause], then [effect]"`

jsamuel1: (q cli should help update all the other bits similarly)

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+2. Match it against registered hooks
+3. Execute the effect action (updating README.md)
+4. Notify the user that the hook was triggered
+
+## Hook Notification and Control
+
+Users can control how hooks are executed:
+
+```
+q hook config --notification=always|important-only|never
```
jsamuel1, line 69: per the first part -- these should be "/ commands", rather than commandline commands.

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+    effect: Action,
+    enabled: bool,
+    created_at: DateTime,
+    last_triggered: Option<DateTime>,
+}
+
+enum Condition {
+    FileEvent(FileEventCondition),
+    CommandEvent(CommandEventCondition),
+    TimeEvent(TimeEventCondition),
```
jsamuel1, line 128: Nice.  :)   We should definitely have TimeEvent/Cron type events.

We may be able to simplify the conditions here to a Tool, Tool command/subcommand and a Context.   eg.  "File_read, <any command>, ./**/*.md"

This should be able to cover MCP tools, internal tools, internal commands (via the use_q_command tool -- in another RFC).

Does Time then become another condition on an event?
eg.  Trigger on "Any Tool, Any Context" at most "Every hour"

jsamuel1: Though we may want something for external inputs too - eg. If triggered via X.  (SNS event,  IDE Plugin event, ...)

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+- Action: Run unit tests
+
+### Event Detection
+
+The system monitors relevant events through:
+
+1. File system watchers for file events
+2. Command interception for command events
+3. Periodic checks for time-based events
+
```
jsamuel1, line 161: External event watchers too?  (SNS, Operating System events, Terminal resize event, ...)

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+   a. If auto-execute is enabled, Action Executor performs the effect
+   b. Otherwise, user is prompted for confirmation
+5. Results and notifications are displayed to the user
+
+### Persistence
+
+Hooks are stored in a JSON file in the Q CLI configuration directory:
+
+```
+~/.config/q-cli/hooks.json
```
jsamuel1, line 177: @mr-lee -- Per user, Per-Project or both?
This seems closer to Rules, and therefore we should do Per-Project to match?

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+3. **Command Execution**: For intercepting and executing commands
+4. **NLP Pipeline**: For parsing natural language hook definitions
+5. **Notification System**: For alerting users about hook triggers
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+1. **Complexity**: Adds significant complexity to the Q CLI system
+2. **Performance Impact**: Continuous monitoring for hook triggers could impact performance
```
jsamuel1, line 195: Can the hooks be pre-parsed and attached to data structures, rather than evaluated continuously?
What is the impact here?

agent-hooks-rfc.md
akashanand98: Updating these in the RFC as per the comment now.

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+3. **Project Standards**:
+   ```
+   q hook create "When new JavaScript files are created, then ensure they have proper header comments"
+   ```
+
+4. **Dependency Management**:
+   ```
+   q hook create "When package.json is modified, then check for security vulnerabilities"
+   ```
+
```
jsamuel1, line 96: "When git pull is called, then review any merge conflicts and present them in a table with suggested actions"

"When the Jira MCP is called to change the state of a ticket to Under Development, then Create a new Git branch with the Jira ticket number and description based off latest main brain"

agent-hooks-rfc.md
akashanand98: These CRON based events can't bind to a single q chat session right ?
And time based events seem rare for a developer using a package / repo. Usually SNS /Eventbridge can take care of these more seamlessly..

agent-hooks-rfc.md
```diff
@@ -0,0 +1,250 @@
...
+1. **Hook Marketplace**: Allow users to share and discover useful hooks
+2. **Hook Templates**: Predefined hooks for common workflows
+3. **Hook Chaining**: Allow hooks to trigger other hooks in sequence
+4. **Context-Aware Hooks**: Hooks that consider the broader context of user actions
+5. **Learning Hooks**: Hooks that adapt based on user behavior and feedback
+6. **Team Hooks**: Shared hooks for team workflows and standards
+7. **Integration with CI/CD**: Connect hooks to external CI/CD pipelines
+8. **Visual Hook Editor**: GUI for creating and managing complex hooks
+9. **Hook Analytics**: Track hook usage and effectiveness
```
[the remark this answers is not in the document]
akashanand98: Telemetry-Based Optimization and recommendation is added in the list (future possibilities).

agent-hooks-rfc.md
akashanand98: Little tricky , maybe adding critical/non-critical list of hooks and effectively monitoring them only based on triggers ?

agent-hooks-rfc.md
akashanand98: We can add but it might have dependency on say 'appscript' for MACOS based events..
Example :
 'When system goes into low power mode, then save all current work and suspend background tasks'

agent-hooks-rfc.md
akashanand98: Updating the location to `~/.aws/amazonq/hooks.json`

chaynabors:
I believe we've implemented this but RFCs are moving. We will start fresh there once it is public. Closing.

Issue #1175, opened 2025-04-10
The conversation history character count calculation is inaccurate


brandonskiser:
As noted in #1169 the `calculate_char_count` method on `ConversationState` is inaccurate since it is not being performed on the result of `as_sendable_conversation_state`, which is the actual conversation state being sent to the backend.

To facilitate an accurate character count calculation, we need to decouple our in-app representation of the conversation from the data being sent to the backend.

One approach would be to update `as_sendable_conversation_state` to be idempotent. Any code that already calls `as_sendable_conversation_state` would therefore not need to be updated.

sandikodev:
After reviewing the current source, `calculate_char_count` in `conversation.rs` already calls `backend_conversation_state` (which is the equivalent of `as_sendable_conversation_state`) rather than computing the count on the raw in-memory state. This appears to have been addressed as part of the context tracking refactor. Can a maintainer confirm whether this can be closed?

PR #1177, opened 2025-04-10
feat: /usage

GoodluckH:
https://github.com/user-attachments/assets/8c7c32ef-4d11-4c0f-90e1-13d5a9d2bc76

Currently, I'm basically reconstructing the `FigConverationState` output and that's gonna be hard to scale and maintain. It's a beta feature with rough est. Will follow up with a more accurate representation once #1175 is resolved.

Follow on ticket: #1178

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -73,6 +73,7 @@ use hooks::{
     HookTrigger,
 };
 use summarization_state::{
+    CONTEXT_WINDOW_SIZE,
     SummarizationState,
     TokenWarningLevel,
 };
@@ -205,6 +206,7 @@ const HELP_TEXT: &str = color_print::cstr! {"
   <em>rm</em>          <black!>Remove file(s) from context [--global]</black!>
   <em>clear</em>       <black!>Clear all files from current context [--global]</black!>
   <em>hooks</em>       <black!>View and manage context hooks</black!>
+<em>/usage</em>      <black!>Show current session's context window usage</black!>
 
 <cyan,em>Tips:</cyan,em>
 <em>!{command}</em>            <black!>Quickly execute a command in your current session</black!>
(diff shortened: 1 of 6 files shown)


crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -1535,6 +1538,155 @@ where
...
+                let window_width = self.terminal_width();
+                let progress_bar_width = std::cmp::min(window_width, 80); // set a max width for the progress bar for better aesthetic
+
+                let context_width =
+                    ((context_token_count as f64 / CONTEXT_WINDOW_SIZE as f64) * progress_bar_width as f64) as usize;
+                let assistant_width =
+                    ((assistant_token_count as f64 / CONTEXT_WINDOW_SIZE as f64) * progress_bar_width as f64) as usize;
+                let user_width =
+                    ((user_token_count as f64 / CONTEXT_WINDOW_SIZE as f64) * progress_bar_width as f64) as usize;
+                let left_over_width = progress_bar_width - context_width - assistant_width - user_width;
```
brandonskiser, line 1598: This needs to be safe, get a panic when I use too much context.

Screenshot 2025-04-14 at 10 30 29 AM

brandonskiser (changes requested):

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -205,6 +206,8 @@ const HELP_TEXT: &str = color_print::cstr! {"
 
 const RESPONSE_TIMEOUT_CONTENT: &str = "Response timed out - message took too long to generate";
 
+const CONTEXT_WINDOW_SIZE: usize = 200_000; // tokens
```
brandonskiser, line 209: We should have some shared behavior with `summarization_state` around here too, like have this be the window size and the summarization state `MAX_CHARS` constant be derived from this.

GoodluckH: Which one is the source of the truth? The 200k tokens or 500k chars?

Rn I'm just dividing the char count by 3 to get tokens. With that logic, the new derived MAX_CHARS would be 600k, not 500k

crates/q_cli/src/cli/chat/mod.rs
GoodluckH: addressed

brandonskiser (approved):

PR #1180, opened 2025-04-10
Ask for confirmation before clearing chat history

GoodluckH:
Sometimes users may habitually type `/clear` to clean the console output but in `q chat`, this actually erase chat history. 

Adding a confirmation dialog to prevent mistakes.

https://github.com/user-attachments/assets/76a17b79-9700-4e3d-b295-228d3378d8b5

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -706,7 +706,8 @@ where
             }
         }
 
-        if !skip_printing_tools && pending_tool_index.is_some() {
+        let show_tool_use_confirmation_dialog = !skip_printing_tools && pending_tool_index.is_some();
+        if show_tool_use_confirmation_dialog {
             execute!(
                 self.output,
                 style::SetForegroundColor(Color::DarkGrey),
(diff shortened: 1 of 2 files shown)


crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -850,16 +850,57 @@ where
...
                     style::SetForegroundColor(Color::Green),
-                    style::Print("\nConversation history cleared.\n\n"),
-                    style::SetForegroundColor(Color::Reset)
+                    style::Print("y"),
+                    style::SetForegroundColor(Color::DarkGrey),
+                    style::Print("/"),
+                    style::SetForegroundColor(Color::Green),
+                    style::Print("n"),
+                    style::SetForegroundColor(Color::DarkGrey),
+                    style::Print("]:\n"),
```
brandonskiser, line 866: nit, maybe add an extra newline here for consistency - other points in the conversation seem to always have a blank line between the last text and the next `>` prompt.

brandonskiser (changes requested):

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -850,16 +850,57 @@ where
...
+                            if line.trim().is_empty() {
+                                continue;
+                            }
+                            break line;
+                        },
+                        (None, false) => {
+                            execute!(
+                                self.output,
+                                style::Print(format!(
+                                    "\n(To exit, press Ctrl+C or Ctrl+D again or type {})\n\n",
```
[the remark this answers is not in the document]
GoodluckH: refactored

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -731,36 +731,22 @@ where
...
-                (None, true) => return Ok(ChatState::Exit),
-            }
+        let user_input = match self.read_user_input(&self.generate_tool_trust_prompt()) {
+            Ok(input) => input,
+            Err(ChatError::Interrupted { .. }) => return Ok(ChatState::Exit),
+            Err(e) => {
+                execute!(
+                    self.output,
+                    style::SetForegroundColor(Color::Red),
+                    style::Print(format!("\nError using tools: {}\n\n", e)),
```
brandonskiser, line 741: This error message doesn't make much sense imo, this is just prompting the user regardless of whether or not tools are being presented.

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -2141,6 +2162,41 @@ where
         Ok(())
     }
 
+    /// Helper function to read user input with a prompt and Ctrl+C handling
+    fn read_user_input(&mut self, prompt: &str) -> Result<String, ChatError> {
```
brandonskiser, line 2166: Previously this logic would just bubble up errors or return an exit state, so now we have to handle errors when instead it would be nicer to just have:
```rust
let input = match self.read_user_input(...)? {
Some(input) => input,
None => return ChatState::Exit
}
```

Prevents us from now handling errors verbosely in multiple locations.

Essentially change this to return a `Result<Option<String>, ChatError>` instead

brandonskiser (changes requested):

crates/q_cli/src/cli/chat/mod.rs
GoodluckH: You're right! This is much cleaner now

crates/q_cli/src/cli/chat/mod.rs
GoodluckH: removed

brandonskiser (approved):

PR #1181, opened 2025-04-10
feat: Add fzf integration for command selection and context management

bergjaak:
Here is a skim integration into Q CLI. This will make it a lot more efficient to use slash commands, including adding and removing files from the `/context`.

Here is a video of how this looks in action:

https://github.com/user-attachments/assets/33a83928-6fef-493a-9273-11cc20e9e2c6

- Add Ctrl+K keybinding to open skim (rust version of fzf) command selector
- Implement file selection for context add commands
- Implement context file selection for context rm commands


bergjaak:
I will replace the list of hardcoded strings with something (vibecoded btw)

I will also replace fzf with a Skim dependency

crates/q_cli/src/cli/chat/skim_integration.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// Represents a command with its description
+#[derive(Debug, Clone)]
+pub struct CommandInfo {
+    pub command: String,
+    pub description: String,
+}
+
+/// Get all available commands with their descriptions
+pub fn get_available_commands() -> Vec<CommandInfo> {
```
jsamuel1, line 20: todo.  another branch is adding a CommandManager, that will help here.   Do you want a get_available_commands() call that returns command + description?

(Non-blocking - as an FYI for the reviewers)

crates/q_cli/src/cli/chat/skim_integration.rs
```diff
@@ -0,0 +1,377 @@
...
+    let mut temp_file = NamedTempFile::new()?;
+    temp_file.write_all(items.join("\n").as_bytes())?;
+    
+    // Build skim options
+    let options = SkimOptionsBuilder::default()
+        .height(Some("40%"))
+        .prompt(Some(prompt))
+        .reverse(true)
+        .multi(multi)
+        .color(Some("fg:252,bg:234,hl:67,fg+:252,bg+:235,hl+:81"))
```
jsamuel1, line 88: Check out cli/theme.rs for color theme class.

crates/q_cli/src/cli/chat/skim_integration.rs
```diff
@@ -0,0 +1,377 @@
...
+                            cmd.push_str(&format!(" {}", file));
+                        }
+                        Ok(Some(cmd))
+                    },
+                    _ => Ok(Some(selected_command.clone())), // User cancelled file selection, return just the command
+                }
+            } else if selected_command == "/tools trust" || selected_command == "/tools untrust" {
+                // For tools trust/untrust, we need to select a tool
+                let tools = vec![
+                    "fs_read".to_string(),
```
jsamuel1, line 174: How will we find these lists again to update them?

bergjaak: I think I made this kind of better but it's still kind of janky. I now have it setup where a unit test should fail if the commands aren't found. But there's probably a better way

crates/q_cli/src/cli/chat/skim_integration.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+#[derive(PartialEq)]
+enum CommandType {
+    ContextAdd(String),
+    Tools(&'static str),
+    Profile(&'static str),
+}
+
+impl CommandType {
```
bergjaak, line 194: I am guessing there's got to be a better way than this

bergjaak: jsamuel1 pointed out to me that `My rfc execute_command thingy that is actually use_q_command, has a CommandRegistry, with CommandHandler, which will fix all the hardcoding.` So I will look into that and see how that can fully resolve this jank

crates/q_cli/src/cli/chat/skim_integration.rs
bergjaak: I think that would be great!

crates/q_cli/src/cli/chat/skim_integration.rs
mschrage: +1 - we probably shouldn't be hardcoding colors as this will change depending on the user's terminal theme.

brandonskiser: Screenshot 2025-04-11 at 5 19 35 PM

This is how it looks on light mode - I'd call this blocking. Prefer if we can at least get everything printed as dark mode

Btw `cli/theme.rs` seems unrelated for this task, that's more tied to the autocomplete feature.

crates/q_cli/src/cli/chat/skim_integration.rs
brandonskiser: guess it's fine for now

brandonskiser (changes requested):

crates/q_cli/src/cli/chat/skim_integration.rs
bergjaak: I was able to just remove the color options and now it defaults to the color theme of the user's terminal (or whatever the correct terminology is there, not sure). As evidence, please see the updated video that I posted in the PR description as evidence

crates/q_cli/src/cli/chat/input_source.rs
```diff
@@ -23,9 +37,56 @@ mod inner {
...
+    }
+}
+
+impl ConditionalEventHandler for SkimCommandSelector {
+    fn handle(&self, _evt: &Event, _n: RepeatCount, _positive: bool, _ctx: &EventContext<'_>) -> Option<Cmd> {
+        // Launch skim command selector with the context manager if available
+        match skim_integration::select_command(self.context_manager.as_deref()) {
+            Ok(Some(command)) => {
+                // Return a command to replace the current line with the selected command
+                Some(Cmd::Replace(Movement::WholeBuffer, Some(command)))
```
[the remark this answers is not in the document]
bergjaak: I also couldn't figure this out... I changed it to `insert` but that seems to only have the correct behavior for `vi` mode. And it doesn't seem possible to chain 2 commands together (but I could be wrong)

brandonskiser (approved):

crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -161,6 +162,7 @@ const WELCOME_TEXT: &str = color_print::cstr! {"
 <em>/quit</em>         <black!>Quit the application</black!>
 
 <cyan!>Use Ctrl(^) + j to provide multi-line prompts.</cyan!>
+<cyan!>Use Ctrl(^) + k to fuzzily search commands and context (use tab to select multiple files).</cyan!>
```
ANorwell, line 165: This is overriding the existing rustyline (and every other shell environment) binding for ctrl-K of kill to end of line. I think another binding is needed.

PR #1198, opened 2025-04-11
fix(/context): handle filepaths with white spaces

GoodluckH:
https://github.com/user-attachments/assets/0436c9ef-8058-4cd1-bcf7-c88ce6c71bdc

### DIFF
--- a/crates/q_cli/src/cli/chat/command.rs
+++ b/crates/q_cli/src/cli/chat/command.rs
@@ -485,13 +485,18 @@ impl Command {
                             let mut force = false;
                             let mut paths = Vec::new();
 
-                            for part in &parts[2..] {
-                                if *part == "--global" {
+                            let args = match shlex::split(&parts[2..].join(" ")) {
+                                Some(args) => args,
+                                None => return Err("Failed to parse quoted arguments".to_string()),
+                            };
+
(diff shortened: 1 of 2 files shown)


crates/q_cli/src/cli/chat/command.rs
```diff
@@ -558,6 +566,33 @@ impl Command {
             prompt: input.to_string(),
         })
     }
+
+    /// Extract file paths from the arguments
+    /// Handles quoted paths that include whitespaces
+    /// and stores them in the `paths` vector.
+    fn collect_file_paths(args: &[&str], paths: &mut Vec<String>, i: &mut usize) {
```
hayemaxi, line 573: Can't we just use `shlex` module? We already have the dependecy. I am using this to process quote args in the hooks implementation: https://github.com/hayemaxi/amazon-q-developer-cli/blob/c6c1506b1e47b959b37463413f9b0b8ee1135207/crates/q_cli/src/cli/chat/command.rs#L638

GoodluckH: Wow I didn't know we have that already! Just refactored

hayemaxi (approved):

Issue #1157, opened 2025-04-09
Add startup message for -a, --acceptall, --trust-all-tools options


hayemaxi:
https://github.com/aws/amazon-q-developer-cli/pull/1120#issuecomment-2784903588 for https://github.com/aws/amazon-q-developer-cli/pull/1120

> Could you also add that "All tools are now trusted..." language when user enter q chat with -a or -acceptall flag?

PR #1200, opened 2025-04-11
feat(/tools): update trustall UX, /trust and /untrust capability

hayemaxi:
https://github.com/aws/amazon-q-developer-cli/issues/1157

- Make trustall/acceptall warning more verbose.
- Trustall/acceptall warning will show if you start chat with -a or --acceptall or --trust-all-tools
- /trust and /untrust takes an unlimited amount of tools at a time
- /trust and /untrust will display an error if you specify a tool name that does not exist

image
image

image
image

### DIFF
--- a/crates/q_cli/src/cli/chat/mod.rs
+++ b/crates/q_cli/src/cli/chat/mod.rs
@@ -211,8 +211,8 @@ const HELP_TEXT: &str = color_print::cstr! {"
   <em>--summary</em>   <black!>Display the summary after compacting</black!>
 <em>/tools</em>        <black!>View and manage tools and permissions</black!>
   <em>help</em>        <black!>Show an explanation for the trust command</black!>
-  <em>trust</em>       <black!>Trust a specific tool for the session</black!>
-  <em>untrust</em>     <black!>Revert a tool to per-request confirmation</black!>
+  <em>trust</em>       <black!>Trust a specific tool or tools for the session</black!>
+  <em>untrust</em>     <black!>Revert a tool or tools to per-request confirmation</black!>
   <em>trustall</em>    <black!>Trust all tools (equivalent to deprecated /acceptall)</black!>
   <em>reset</em>       <black!>Reset all tools to default permission levels</black!>
 <em>/profile</em>      <black!>Manage profiles</black!>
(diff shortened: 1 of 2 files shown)


crates/q_cli/src/cli/chat/mod.rs
```diff
@@ -204,6 +203,8 @@ const HELP_TEXT: &str = color_print::cstr! {"
 "};
 
 const RESPONSE_TIMEOUT_CONTENT: &str = "Response timed out - message took too long to generate";
+const TRUST_ALL_TEXT: &str = color_print::cstr! {"\n<green!>All tools are now trusted (<red!>!</red!>). This <bold>disables</bold> acceptance prompting.\
+\nAgents can sometimes do unexpected things so understand the risks.</green!>"};
```
GoodluckH, line 207: Maybe say ".... Q will execute tools without asking for confirmation."

it's shorter

hayemaxi: I received feedback that warning was too short 😔

GoodluckH: Oh then maybe replace "This disables...prompting." with "Q will execute tools without asking for confirmation." Because "acceptance prompting" might be hard to understand

GoodluckH (approved):

GoodluckH:
I wonder if we can remove that text but for the user input indicator, we can show ![all tools trusted]> initially

Then for the subsequent inputs we just show !>

We assume our users are smarter than us, so if they somehow figured out how to trust all tools, they should already know the risk.

We just need some text indicating what ! means

GoodluckH (changes requested):
See comments.

GoodluckH:
> We have gotten some feedback that trust doesn't fully explain the implications of using it

Interesting, that means we are not communicating that properly when they run `/tools trustall`. If they can find that command, I think they pretty much know what they're doing. Maybe we can just add a confirmation dialog when they run `/tools trustall`, and put verbose stuff there instead of on the greeting screen. 

Or if confirmation dialog is too much code change, we can just add that risk disclaimer in a different color. And also the "!". Maybe we can replace ! with 🤖 or something to look not too punishing.

GoodluckH (approved):
Approving for now to unblock release. we can iterate on UI as we go

Issue #1208, opened 2025-04-11
Amazon Q CLI `/editor` command fails when EDITOR contains arguments


scorbiere:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.3.2 (24D81)

### Expected behaviour

The `/editor` command should properly parse the EDITOR environment variable, separating the command from its arguments and executing them correctly.

### Actual behaviour

When the EDITOR environment variable contains a command with arguments (like `code -w`), the `/editor` command fails to open the editor. This happens because the command with arguments is treated as a single executable name rather than being properly parsed.

```
[chat-transcript]
> /editor

Error opening editor: Failed to open editor: No such file or directory (os error 2)
```

Possible workaround is:
```bash
# Set a valid value for the variable FOLDER_IN_PATH
echo '#!/usr/bin/env bash
code -w "$*"' > $FOLDER_IN_PATH/qeditor
chmod +x $FOLDER_IN_PATH/qeditor

export EDITOR=qeditor
```

After setting this up, the `/editor` command in `q chat` will successfully open VS Code and wait for you to close the file before continuing.

### Steps to reproduce

1. Set EDITOR environment variable to include arguments: `export EDITOR="code -w"`
2. Run `q chat`
3. Try to use the `/editor` command
4. Observe that it fails to open the editor

### Environment

```yaml
[q-details]
version = "1.7.3"
hash = "3e4ae79d371315e80ddac772b43fff2cba314104"
date = "2025-04-10T05:48:16.435186Z (2d ago)"
variant = "full"

[system-info]
os = "macOS 15.3.2 (24D81)"
chip = "Apple M1 Pro"
total-cores = 10
memory = "32.00 GB"

[environment]
cwd = "/Users/USER"
cli-path = "/Users/USER"
os = "Mac"
shell-path = "/bin/zsh"
shell-version = "5.9"
terminal = "macOS"
install-method = "unknown"

[env-vars]
QTERM_SESSION_ID = "542556a4b606434bacef617bf135ee8a"
Q_SET_PARENT_CHECK = "1"
Q_TERM = "1.7.1"
Q_USING_ZSH_AUTOSUGGESTIONS = "1"
SHELL = "/bin/zsh"
[...]

PR #1209, opened 2025-04-12
fix(chat): Parse EDITOR env var to support arguments

carlesso:
*Issue #, if available:* #1208

*Description of changes:* The /editor command now properly parses the EDITOR environment variable to handle cases where it contains arguments (like 'code -w'). This fixes issue #1208 where the command would fail when EDITOR contained spaces.


crates/q_cli/Cargo.toml
```diff
@@ -89,6 +89,7 @@ whoami.workspace = true
 winnow.workspace = true
 shell-color.workspace = true
 syntect = { version = "5.2.0", features = [ "default-syntaxes", "default-themes" ]}
+shell-words = "1.1.0"
```
brandonskiser, line 92: Let's not add this dependency, can you try using `shlex` instead?

brandonskiser (changes requested):

crates/q_cli/Cargo.toml
carlesso: Sure, updated!

brandonskiser (approved):

PR #1216, opened 2025-04-13
feat(context): de-duplicate hook names within configs

hayemaxi:
Follow up to https://github.com/aws/amazon-q-developer-cli/pull/1176#discussion_r2040038176

- Seamlessly de-duplicate hook names in a single config instead of making the user deal with it
  - Also checks the config file and deduplicates any names in case they were added outside of chat
- Use separate hook caches for profile vs global hooks:
  - Allows hooks to have the same names across profile/global configs
  - Allows us to clear execute cache easily when switching profiles

### DIFF
--- a/crates/q_cli/src/cli/chat/context.rs
+++ b/crates/q_cli/src/cli/chat/context.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::io::Write;
 use std::path::{
     Path,
@@ -22,7 +23,6 @@ use super::hooks::{
     Hook,
     HookExecutor,
 };
-use crate::cli::chat::hooks::HookConfig;
 
 pub const AMAZONQ_FILENAME: &str = "AmazonQ.md";
 
@@ -32,7 +32,9 @@ pub const AMAZONQ_FILENAME: &str = "AmazonQ.md";
 pub struct ContextConfig {
     /// List of file paths or glob patterns to include in the context.
     pub paths: Vec<String>,
-    pub hooks: HookConfig,
+
+    /// Map of Hook Name to [`Hook`]. The hook name serves as the hook's ID.
+    pub hooks: HashMap<String, Hook>,
 }
 
 #[allow(dead_code)]
@@ -353,6 +355,7 @@ impl ContextManager {
     /// A Result indicating success or an error
     pub async fn switch_profile(&mut self, name: &str) -> Result<()> {
         validate_profile_name(name)?;
+        self.hook_executor.profile_cache.clear();
 
         // Special handling for default profile - it always exists
         if name == "default" {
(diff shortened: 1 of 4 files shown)


crates/q_cli/src/cli/chat/hooks.rs
```diff
@@ -109,13 +112,15 @@ pub struct CachedHook {
 
 #[derive(Debug, Clone)]
 pub struct HookExecutor {
-    pub execution_cache: HashMap<String, CachedHook>,
+    pub global_cache: HashMap<String, CachedHook>,
+    pub profile_cache: HashMap<String, CachedHook>,
```
brandonskiser, line 116: Helpful to add a doc comment stating the key is the hook name

crates/q_cli/src/cli/chat/context.rs
```diff
@@ -545,30 +542,49 @@ impl ContextManager {
...
+                &*h
+            }));
+        }
 
         self.hook_executor.run_hooks(hooks, updates).await
     }
+
+    /// Ensure that all hooks in a single config have a unique name.
+    /// Hooks will be renamed to be unique if there are duplicates, and the config is saved.
+    async fn validate_hook_config(&mut self, global: bool) -> Result<()> {
```
brandonskiser, line 566: I don't like having a lot of extra logic for validating and creating new hook names when this could also be solved by just changing the structure used when saving hooks. We have that `ContextConfig` is already saved per profile, so instead of saving hooks as a list of hooks, why not instead have it be:
```
ContextConfig {
    paths: ...,
    hooks: HookConfig {
        conversation_start: { [key: string]: Hook },
        per_prompt: { [key: string]: Hook },
    }
}
```

This way we could also avoid having to create randomly generated id's for hook names. I feel it's simpler for us and the user to just present an error to the user stating a hook with that name already exists instead of this approach.

hayemaxi: > ```
> ContextConfig {
>     paths: ...,
>     hooks: HookConfig {
>         conversation_start: { [key: string]: Hook },
>         per_prompt: { [key: string]: Hook },
>     }
> }
> ```

This won't work either. Hook names have to be unique across conversation_start AND per_prompt, otherwise this si ambiguous: `/context hooks rm my_hook`. It could be in either of the categories.

> This way we could also avoid having to create randomly generated id's for hook names. I feel it's simpler for us and the user to just present an error to the user stating a hook with that name already exists instead of this approach.

It's not random, and I would argue its easier for the user because there is never a failure point for them. Their hook _always_ gets added. Also, there is a lack of any config validation happening that we desperately need, whether right now or later (currently if there is an issue with the JSON then the context manager just doesn't load without any warning to the user). 
However, the automatic renaming may be confusing, and maybe this issue is a just a symptom of the structure of the config. Maybe we can resolve this by just changing the schema to not have these separations. Thoughts?

```
context.json

{
    paths: [ ... ],
    hooks: {
        my_hook : { // the name, which is now unique amongst all hooks in the config file
             ... normal hook data
             "when": "conversation_start" | "per_prompt", // open to a different name, could be "on" as well
        }
     ...
    }
}
```

brandonskiser: > and I would argue its easier for the user because there is never a failure point for them. Their hook always gets added. Also, there is a lack of any config validation happening that we desperately need, whether right now or later

Definitely true about handling validation and serialization errors - although, the random naming is still just strange to me because hooks should probably have some more semantic meaning to their name instead of just having the user get away with `hook_1, hook_2, hook_3` etc. I'd like to hear others thoughts on this hook naming part before going with this approach.

> Maybe we can resolve this by just changing the schema to not have these separations. Thoughts?
I'd prefer this! Saves having to have hook name as a validation check in any validation logic we might have.

brandonskiser (approved):

PR #1217, opened 2025-04-13
Update README.md with new features in version 1.7.3

aws-mbcohn:
This PR updates the README.md file to include the latest features added in version 1.7.3:

   - Added Non-Interactive Mode feature
   - Added Tool Permissions feature
   - Added Conversation Management feature (compact command)
   - Added Editor Integration feature

   These features were already implemented but were missing from the README feature list. This update ensures users are aware of all the current capabilities of Amazon Q CLI.

   The changes are minimal and only affect the Features section of the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 <p align="center">
   <a href="https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html">
     <picture>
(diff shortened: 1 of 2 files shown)


GoodluckH:
For README features, I think we might want to keep it as minimal as possible since it's more like a landing page for us. We try to not make the page to become another AWS doc. Think of README as a first impression to people never heard of Q CLI, we should convert them as quick as possible.

Detailed and non key features are already included in the release notes and in the upcoming #1117 feature. If you want to highlight the fact that we have more features than listed, I would suggest just add one more line that sounds like "And so much more: editor mode, context history compression ... star the repo to keep updated" 

@mschrage @mr-lee wdyt

GoodluckH (approved):

chaynabors:
Fully rewriting the README soon. Closing in favor of that.

Issue #1227, opened 2025-04-14
[UX] Make `/tools` UI pretty and easy to understand  + more scalable codes


GoodluckH:
Design concept:
![Image](https://github.com/user-attachments/assets/1c279542-02c6-40ac-a3bb-38e7a85c3cb2)

To make this happen, we need to refactor the codebase a bit to clearly indicate read/write access for each tool

PR #1228, opened 2025-04-14
feat(/tools ux): improve UI and refactor codes

GoodluckH:
#1227


hayemaxi (approved):

hayemaxi:
nit: consider adding an `[green]enabled/[gray]disabled` key the prompt, e.g.:
```
Pre-approved tool operations:
enabled/disabled
```
in case the user finds themselves in a state where everything is the same color.

PR #1276, opened 2025-04-17
Feature: Web Search

GoodluckH:
Currently only supports retrieving. We need to have more discussion to align on what search engine to use

https://github.com/user-attachments/assets/a0c8c13b-b78f-43ab-9524-ddfbc13768e5


crates/q_chat/src/tools/mod.rs
```diff
@@ -175,6 +183,7 @@ impl ToolPermissions {
             "execute_bash" => "trust read-only commands".dark_grey(),
             "use_aws" => "trust read-only commands".dark_grey(),
             "report_issue" => "trusted".dark_green().bold(),
+            "web_search" => "trusted".dark_green().bold(),
```
brandonskiser, line 186: Don't think we should have this as trusted by default, the model could make requests to arbitrary websites that can track the user's activity in unwanted ways.

crates/q_chat/src/tools/tool_index.json
```diff
@@ -147,6 +147,33 @@
       ]
     }
   },
+  "web_search": {
+    "name": "web_search",
+    "description": "Search/retrieving the web for the specified query. Currently only supports retrieving.",
```
brandonskiser, line 152: Maybe hold off on including any search related functionality until it's supported? Model might hallucinate

crates/q_chat/src/tools/web_search.rs
```diff
@@ -0,0 +1,192 @@
...
+            updates,
+            style::Print(format!(
+                "{} {}...",
+                if self.mode == WebSearchMode::Search {
+                    "Searching"
+                } else {
+                    "Retrieving"
+                },
+                if self.mode == WebSearchMode::Search {
+                    self.query.as_ref().unwrap_or(&"".to_string()).clone().dark_green()
```
brandonskiser, line 130: nit - `unwrap_or(&"".to_string()).clone()` seems strange, shouldn't need to `clone` here right?

crates/q_chat/src/tools/web_search.rs
```diff
@@ -0,0 +1,192 @@
...
+                let robots_url = format!(
+                    "{}://{}/robots.txt",
+                    parsed_url.scheme(),
+                    parsed_url
+                        .host_str()
+                        .ok_or_else(|| eyre::eyre!("Invalid host in URL"))?
+                );
+
+                let user_agent = "AmazonQCLI/1.0";
+                let client = reqwest::Client::new();
```
brandonskiser, line 66: Use `fig_request::client()` instead, should be fine to add as a workspace dependency for this. It's used as part of checking for updates anyways. The user agent will be set appropriately

crates/q_chat/src/tools/web_search.rs
```diff
@@ -0,0 +1,192 @@
...
+                    return Err(eyre::eyre!("Request failed with status: {}", response.status()));
+                }
+                // Get the response body as text
+                let html_string = response
+                    .text()
+                    .await
+                    .map_err(|e| eyre::eyre!("Failed to read response body: {}", e))?;
+
+                // Convert HTML to Markdown
+                let converter = HtmlToMarkdown::builder().skip_tags(vec!["script", "style"]).build();
```
brandonskiser, line 105: What's the reasoning behind this? Does the model not understand html well enough, or is this intended to primarily save on token usage?

Just asking since I would debate whether or not it's worth adding this dependency

crates/q_chat/src/tools/web_search.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+        if self.mode == WebSearchMode::Retrieve && self.target_url.is_none() {
+            return Err(eyre::eyre!("Target URL is required for retrieving"));
+        }
+
+        Ok(())
+    }
+
+    // Simple function to check if a path is allowed by robots.txt
+    fn is_allowed_by_robots_txt(robots_content: &str, user_agent: &str, path: &str) -> bool {
```
brandonskiser, line 151: We should probably have some tests to verify robots txt parsing is correct

crates/q_chat/src/tools/mod.rs
awescarsoto: +1. For ref, Claude Code's similar WebFetch requires permissions instead of being trusted by default: https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview#tools-available-to-claude

chaynabors:
We want web search, but this is very out of date and needs to be rethought when we get to that point.

PR #1280, opened 2025-04-18
feat(context): improve hooks execution UI

hayemaxi:
Update the UI of running hooks to align with the loading UI of MCP.

https://github.com/user-attachments/assets/f34700d7-0488-457d-b6ff-0aff5a58f693

### DIFF
--- a/crates/q_cli/src/cli/chat/hooks.rs
+++ b/crates/q_cli/src/cli/chat/hooks.rs
@@ -5,21 +5,32 @@ use std::time::{
     Instant,
 };
 
-use crossterm::style::Color;
+use crossterm::style::{
+    Color,
+    Stylize,
+};
 use crossterm::{
+    cursor,
     execute,
     queue,
     style,
+    terminal,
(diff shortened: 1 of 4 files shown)


crates/q_cli/src/cli/chat/hooks.rs
```diff
@@ -127,99 +138,148 @@ impl HookExecutor {
...
+                    },
+                }
             }
-        }
 
-        if !skip_output {
-            let _ = execute!(updates, style::Print("\n"),);
-        }
+            // Process results regardless of output enabled
+            total_duration += duration;
```
brandonskiser, line 237: These execute in parallel so this isn't correct, we should instead calculate total duration after we finish awaiting all the hook futures. Since we present this time to the user we should update this

brandonskiser (changes requested):

brandonskiser (approved):

brandonskiser (approved):

PR #1320, opened 2025-04-23
chore: write credentials to a file

bergjaak:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/crates/fig_auth/src/builder_id.rs
+++ b/crates/fig_auth/src/builder_id.rs
@@ -831,7 +831,7 @@ mod tests {
         let json: serde_json::Value = match serde_json::from_str(&content) {
             Ok(json) => json,
             Err(e) => {
-                println!(?e, "Failed to parse JSON");
+                println!("Failed to parse JSON: {:?}", e);
                 return false;
             },
         };
@@ -865,7 +865,6 @@ mod tests {
 
     #[test]
     fn test_write_credentials() {
-        // putting these in a single test so that there is no concurrency.
         test_write_credentials_builder_id();
         test_write_credentials_identity_center();
         test_refresh_writes_credentials();
--- a/crates/fig_auth/src/builder_id.rs
+++ b/crates/fig_auth/src/builder_id.rs
@@ -537,14 +537,68 @@ pub async fn poll_create_token(
     }
 }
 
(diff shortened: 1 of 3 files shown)


dingfeli:
Not entirely sure why the lock file was changed here since it does not seem like you had pulled in new dependencies. 
Our CI is run with --lock flag, which menas the lock file would need to be up to date in the commit, otherwise it would fail since the --lock flag prevents it from updating. 

When it becomes out of sync for whatever reason, checkout the version of the lock file from main with `git checkout main -- Cargo.lock` as a baseline and run one of `cargo check`, `cargo build` or any commit that would trigger a build to update the lock file. This way all of the existing versions are preserved and you won't run into compatibility warnings like you did there.

crates/fig_auth/src/builder_id.rs
```diff
@@ -537,14 +537,73 @@ pub async fn poll_create_token(
     }
 }
 
+/// Write credentials to ~/.aws/amazonq/creds.json
+pub fn write_credentials_to_file(token: &BuilderIdToken) -> Result<()> {
```
dingfeli, line 541: Since the callee of this function already has an async scope we should probably subscribe to the same IO behavior here and make this async here as well. Otherwise we would be blocking the executing thread for as long as the file write takes.

crates/fig_auth/src/builder_id.rs
```diff
@@ -537,14 +537,73 @@ pub async fn poll_create_token(
...
+        "region": token.region.clone(),
+        "start_url": token.start_url.clone(),
+        "token_type": match token.token_type() {
+            TokenType::BuilderId => "BuilderId",
+            TokenType::IamIdentityCenter => "IamIdentityCenter",
+        }
+    });
+
+    // Ensure ~/.aws/amazonq directory exists
+    let mut path = PathBuf::from(std::env::var("HOME").unwrap_or_else(|_| "~".to_string()));
```
dingfeli, line 559: I don't think tilde expansion works out of the box like this. 
See https://github.com/aws/amazon-q-developer-cli/blob/main/crates/fig_desktop_api/src/util.rs#L10-L17

dingfeli:
Not entirely sure if this is still relevant. Feel free to contact us if you think it is and we'll take another look at it.

PR #1324, opened 2025-04-23
fix: add check for tool use timeouts

brandonskiser:
*Description of changes:*
- Updating the tool use stream timeout to handle cases where we do receive a stop event, but still fail serialization after waiting for a long time

Screenshot 2025-04-23 at 2 57 18 PM

### DIFF
--- a/crates/q_cli/src/cli/chat/parser.rs
+++ b/crates/q_cli/src/cli/chat/parser.rs
@@ -197,11 +197,9 @@ impl ResponseParser {
                 // If we failed deserializing after waiting for a long time, then this is most
                 // likely bedrock responding with a stop event for some reason without actually
                 // including the tool contents. Essentially, the tool was too large.
-                //
-                // Setting the timeout at 90 secs since it usually happens over 2 mins, but
-                // providing some buffer.
+                // Timeouts have been seen as short as ~1 minute, so setting the time to 30.
                 let elapsed = start.elapsed();
-                if self.peek().await?.is_none() && elapsed > std::time::Duration::from_secs(90) {
+                if self.peek().await?.is_none() && elapsed > std::time::Duration::from_secs(30) {
                     error!(
                         "Received an unexpected end of stream after spending ~{}s receiving tool events",
                         elapsed.as_secs_f64()
--- a/crates/q_cli/src/cli/chat/parser.rs
+++ b/crates/q_cli/src/cli/chat/parser.rs
@@ -1,4 +1,7 @@
-use std::time::Instant;
+use std::time::{
+    Duration,
+    Instant,
+};
 
 use eyre::Result;
 use fig_api_client::clients::SendMessageOutput;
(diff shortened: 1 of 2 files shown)


crates/q_cli/src/cli/chat/parser.rs
```diff
@@ -180,29 +180,29 @@ impl ResponseParser {
...
-                // If the stream ended before we saw the final tool use event (and thus failed
-                // deserializing the tool use), this is most likely due to the backend dropping the
-                // connection. The tool was too large!
-                if self.peek().await?.is_none() && !stop_seen {
+                // If we failed deserializing after waiting for a long time, then this is most
+                // likely bedrock responding with a stop event for some reason without actually
+                // including the tool contents. Essentially, the tool was too large.
+                // Timeouts have been seen as short as ~1 minute, so setting the time to 30.
+                let elapsed = start.elapsed();
+                if self.peek().await?.is_none() && elapsed > std::time::Duration::from_secs(30) {
```
mschrage, line 202: Why 30 seconds rather than something longer?

crates/q_cli/src/cli/chat/parser.rs
```diff
@@ -180,29 +180,29 @@ impl ResponseParser {
...
                     error!(
                         "Received an unexpected end of stream after spending ~{}s receiving tool events",
-                        Instant::now().duration_since(start).as_secs_f64()
+                        elapsed.as_secs_f64()
```
mschrage, line 205: Can we present this to the user?

crates/q_cli/src/cli/chat/parser.rs
brandonskiser: We were discussing always just retrying the request, so I just set it as 30 seconds to be small enough to catch almost every case where this happens. First I had it at 90, then noticed it also happened at ~1 min, so going even lower in case there's instances where it's e.g. 40 seconds

crates/q_cli/src/cli/chat/parser.rs
```diff
@@ -211,18 +211,14 @@ impl ResponseParser {
                             [(
                                 "key".to_string(),
                                 serde_json::Value::String(
-                                    "fake tool use args - actual tool use was too large to include".to_string(),
+                                    "WARNING: the actual tool use arguments were too complicated to be generated"
+                                        .to_string(),
```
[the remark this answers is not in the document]
brandonskiser: This is sent to the model instead, the actual user-facing messaging is here: https://github.com/aws/amazon-q-developer-cli/blob/main/crates/q_cli/src/cli/chat/mod.rs#L2581

crates/q_cli/src/cli/chat/parser.rs
brandonskiser: Sure, we can add this time amount to the message presented to the user

aseemxs (approved):

PR #1333, opened 2025-04-23
feat: region expansion

chaynabors:


crates/fig_auth/src/profile.rs
```diff
@@ -0,0 +1,26 @@
...
+        }
+    }
+}
+
+pub fn set_profile(profile: Profile) -> Result<(), Error> {
+    Ok(fig_settings::state::set_value(
+        "api.codewhisperer.service",
+        serde_json::to_string(&profile)?,
+    )?)
+}
```
evanliu048, line 26: Quick question — I noticed that the `Profile` struct we store in `fig_settings::state` only contains the `arn` and `profile_name` fields, but in `load_codewhisperer()` we try to access `endpoint` and `region` via `.get("endpoint")` / `.get("region")`.
https://github.com/aws/amazon-q-developer-cli/pull/1333/files#diff-1a9b0949a6ab32e74e674be062a4151ae46aac201a65f37dd96a9e4b3f56ccdaR43-R47
I’m wondering how endpoint and region are expected to be retrieved here, given that we don’t explicitly store them in `fig_settings::state`.
(For context, in the IDE plugin, we derive the endpoint by extracting the region from the arn, since there’s a 1:1 mapping between region and endpoint.)

evanliu048: the mismatch is solved in endpoints.ts

crates/q_cli/src/cli/user.rs
```diff
@@ -161,6 +184,24 @@ impl RootUserSubcommand {
...
+                }
+
+                if let Ok(Some(token)) = fig_auth::builder_id_token().await {
+                    if matches!(token.token_type(), TokenType::BuilderId) {
+                        bail!("This command is only available for Pro users");
+                    }
+                }
+
+                select_profile_interactive(true).await?;
+
```
evanliu048, line 202: I changed the `whoami` variable in `select_profile_interactive` to false in the **profile subcommand** and set it to true in the **login** flow. The idea is that when users log in and only have one profile, we automatically select it and skip the profile selection step. However, when switching profiles, users will still be able to see and select from their available profiles.

crates/q_cli/src/cli/user.rs
```diff
@@ -316,3 +367,93 @@ async fn try_device_authorization(
     }
     Ok(())
 }
+
+async fn select_profile_interactive(whoami: bool) -> Result<()> {
+    let profiles = list_available_profiles().await;
+    if profiles.is_empty() {
+        info!("Available profiles was empty");
+        return Ok(());
+    }
```
evanliu048, line 376: Every IdC user should have at least one profile. Should we consider throwing an error message if the profile list is empty or if fetching the profiles fails?

brandonskiser (approved):

PR #1340, opened 2025-04-24
feat(tools): execute_bash uses PTY for shell integrations

hayemaxi:
- Create a PTY and spawn the command inside that
- Reads from user input for text and key events, allowing some use for interactive commands
- Uses current program env vars, and users custom shell integrations (dot files e.g. .zshrc)
- Works for both `execute_bash` tool and context hooks
- Moved PTY creation code from `figterm` to `fig_util` instead.

NOTES:
- terminal is set to dumb mode as explained in the code. we should find a way around this.
- we provde all command output which is expensive for some interactive commands that re-draw. Our next steps should involve writing more than a max output to a local file, then prompting Q to read from this file if it wants more context about what the tool did.
- Not tested on windows (testable?) or other shells besides zsh (yet)

*Issue #, if available:*

*Description of changes:*


crates/q_cli/src/cli/chat/tools/execute_bash.rs
```diff
@@ -93,19 +104,205 @@ impl ExecuteBash {
...
+
+        // Wait for the child process to exit
+        let exit_status = child.wait()?;
+
+        let stdout = stdout_lines.into_iter().collect::<String>();
+        Ok(CommandResult {
+            exit_status: exit_status.exit_code(),
+            stdout: format!(
+                "{}{}",
+                truncate_safe(&stdout, max_result_size),
```
hayemaxi, line 296: Does this need `to_str_lossy` ?

crates/fig_util/src/terminal.rs
```diff
@@ -793,6 +798,58 @@ impl IntelliJVariant {
...
+                if c.is_ascii_lowercase() {
+                    return vec![(c as u8) - b'a' + 1];
+                } else if c.is_ascii_uppercase() {
+                    return vec![(c as u8) - b'A' + 1];
+                }
+            }
+            // Regular character
+            c.to_string().into_bytes()
+        },
+        KeyCode::Enter => vec![b'\r'],
```
jsamuel1, line 836: This looks like a non-exhaustive list.   Is there an API can that do this for us?

crates/q_cli/src/cli/chat/tools/execute_bash.rs
```diff
@@ -93,19 +104,205 @@ impl ExecuteBash {
...
+
+        // Wait for the input handler to complete
+        let _ = input_handle.await?;
+
+        // Wait for the child process to exit
+        let exit_status = child.wait()?;
+
+        let stdout = stdout_lines.into_iter().collect::<String>();
+        Ok(CommandResult {
+            exit_status: exit_status.exit_code(),
```
jsamuel1, line 293: often the last few lines are important.

Consider:   
First N lines
a sample of "important lines"  -- . eg. with ERROR, WARNING, etc on them.
Last N lines

Which may then look like.

```
Started app blah
lots of startup stuff

[ skipped X rows of content ] 
WARN: blah blah is wrong
[ skipped X rows of content ]
application is now existing with 49 warnings
exited
```

jsamuel1: also note, using the [ ] type of tagged lines, per how Claude likes to spit them out itself.

crates/q_cli/src/cli/chat/tools/execute_bash.rs
```diff
@@ -93,19 +104,205 @@ impl ExecuteBash {
...
+    async fn _execute_pty(&self, max_result_size: usize, updates: bool, with_input: bool) -> Result<CommandResult> {
+        const LINE_COUNT: usize = 1024;
+
+        // Open a new pseudoterminal
+        let pty_pair = open_pty(&get_terminal_size()).map_err(|e| eyre!("Failed to start PTY: {}", e))?;
+
+        // Create a command builder for the shell command
+        let shell = Shell::current_shell().map_or("bash", |s| s.as_str());
+        let mut cmd_builder = CommandBuilder::new(shell);
+        cmd_builder.args(["-cli", &self.command]);
```
brandonskiser, line 164: what are these args? I believe `-i` is the only required option, we really just need an interactive shell.

This also seems to break with `fish` in my testing

crates/q_cli/src/cli/chat/tools/execute_bash.rs
```diff
@@ -93,19 +104,205 @@ impl ExecuteBash {
...
+                loop {
+                    tokio::select! {
+                        // Check if the process is done
+                        Some(_) = rx.recv() => break Ok(()),
+
+                        // Use a separate task to poll for events to avoid blocking
+                        // Note: this reads one character a time basically. Which is fine for
+                        // everything unless the user pastes a large amount of text.
+                        // Could use an upgrade to avoid this (maybe read from stdin and events at the same time)
+                        event = tokio::task::spawn_blocking(crossterm::event::read) => {
```
brandonskiser, line 240: Why do we use crossterm events instead of just reading straight from stdin? We're in raw mode, so reading from stdin and copying straight to the master fd should be sufficient right?

There's probably edge cases like with handling resizes but we could probably hold off on that for now

crates/q_cli/src/cli/chat/tools/execute_bash.rs
brandonskiser: Where? it's already a UTF-8 type here though?

PR #1349, opened 2025-04-24
fix: add flag for logging in with device flow on CLI, fix broken sigint handling during auth

brandonskiser:
*Description of changes:*
- Adding a flag to always use device code flow for `q login`. This is useful in cases where we think we can open a browser successfully, but are actually unable to handle redirects required for PKCE.
- Fixing sigint handling to actually respect ctrl+c


GoodluckH:
#1390

GoodluckH (approved):

PR #1363, opened 2025-04-25
fix: don't retry compaction if it fails

brandonskiser:
*Description of changes:*
- Don't retry compacting the history if it fails due to context window overflow. This should resolve a scenario where compaction fails and gets auto-retried, ultimately resulting in an infinite loop of failures.


dingfeli (approved):
LGTM but lock file probably needs to be regenerated.

PR #1368, opened 2025-04-26
feat: QCHAT_PROCESS_ID and pre-push hook.

jsamuel1:
Q Chat now sets an environment variable with its process id, which may be used by downstream processes (MCP Servers, Shell Commands, Git hooks), to specify behaviour based on being run inside Q Chat.

A sample git pre-push hook is included in docs/samples/ to show how it could work, and the husky hooks in the project have been updated to include the push prevention.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,5 +1,4 @@
-#!/bin/bash
-. "$(dirname "$0")/_/husky.sh"
+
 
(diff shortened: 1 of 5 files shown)


brandonskiser:
Hey! Sorry for the late response - can you rebase main and push again?

Issue #1303, opened 2025-04-21
[RFC] Real-time end-user communication


GoodluckH:
We need ways to communicate new releases, service updates, etc.

Our existing communication channel to external users is too weak to drive new release adoption.

PR #1403, opened 2025-04-28
feat: new update box + reusable box component

GoodluckH:
#1303 

https://github.com/user-attachments/assets/d7c0ccca-0993-41c2-8728-9848eb0d0612


GoodluckH:
Currently this is checking for update every time user launches Q. Can we make this async?

brandonskiser:
Can you rebase main first? I can see e.g. `dialoguer` being added back in the dependencies when it shouldn't, instead use `strip_ansi_escapes`

crates/q_chat/Cargo.toml
```diff
@@ -15,11 +15,13 @@ clap.workspace = true
 color-print.workspace = true
 convert_case.workspace = true
 crossterm.workspace = true
+dialoguer = { version = "0.11.0", features = ["fuzzy-select"] }
```
brandonskiser, line 18: Sorry can you rebase main first? I can see e.g. dialoguer being added back in the dependencies when it shouldn't, instead use strip_ansi_escapes

crates/q_chat/src/lib.rs
```diff
@@ -674,85 +674,35 @@ impl ChatContext {
...
-
-        // Bottom rounded corner line: ╰────────────╯
-        let bottom = format!("╰{}╯", "─".repeat(box_width - 2));
-        execute!(self.output, style::Print(format!("{}\n", bottom)))?;
-
         Ok(())
     }
 
     async fn try_chat(&mut self) -> Result<()> {
+        tokio::time::timeout(Duration::from_secs(2), self.check_for_updates()).await??;
```
brandonskiser, line 704: This shouldn't happen and block in the main entrypoint, instead can we spawn a task that checks for updates, sets a state value if it's found, and then we'll print the update message in the next go around? (also be sure to gate this only to interactive users)

We do something similar in `figterm`, see:
- Update key: https://github.com/aws/amazon-q-developer-cli/blob/main/crates/figterm/src/update.rs#L19
- Relevant code that spawns task and checks it: https://github.com/aws/amazon-q-developer-cli/blob/main/crates/figterm/src/update.rs#L58-L91

crates/q_chat/src/lib.rs
```diff
@@ -674,85 +674,35 @@ impl ChatContext {
...
-            let content = format!(
-                "│ {: <pad$}{}{: <rem$} │",
-                "",
-                line,
-                "",
-                pad = left_pad,
-                rem = box_width - 4 - left_pad - visible_line_len
-            );
-            execute!(self.output, style::Print(format!("{}\n", content)))?;
+        match fig_install::check_for_updates(false).await {
```
brandonskiser, line 688: Instead we can add `semver` as a dependency here to make this saving/parsing of the version easier, similar to `figterm`. It's a really small dep so no worries here

crates/q_chat/src/lib.rs
```diff
@@ -801,6 +757,31 @@ impl ChatContext {
             )
         )?;
         execute!(self.output, style::Print("\n"), style::SetForegroundColor(Color::Reset))?;
+
+        let new_version = self.state.get_string("chat.new_version_available")?;
+        if new_version.is_some() {
```
brandonskiser, line 762: We should check if the saved version is greater than the current version here - also don't fail on not getting the state value or anything

crates/q_chat/src/lib.rs
brandonskiser: Also, let's not return an error if this fails in any way. Instead just log an error message and move on

brandonskiser (changes requested):

Cargo.toml
```diff
@@ -19,7 +19,7 @@ authors = [
 edition = "2021"
 homepage = "https://aws.amazon.com/q/"
 publish = false
-version = "1.10.0"
+version = "1.11.0"
```
brandonskiser, line 22: Don't bump the version here

crates/q_cli/src/cli/update.rs
```diff
@@ -158,6 +162,7 @@ async fn try_linux_update() -> Result<ExitCode> {
                             ..
                         })) => {
                             spinner.stop_with_message("Update complete".into());
+                            fig_settings::state::remove_value("update.new-version-available")?;
```
brandonskiser, line 165: Can we not error out here if we fail removing this value? Just log an error message and otherwise return success.

Don't want to make the user think updating failed when in actuality it succeeded

crates/q_chat/Cargo.toml
```diff
@@ -15,11 +15,13 @@ clap.workspace = true
 color-print.workspace = true
 convert_case.workspace = true
 crossterm.workspace = true
+semver = { version = "1.0.26", features = ["serde"] }
```
brandonskiser, line 18: This should be a workspace dependency

brandonskiser (changes requested):

brandonskiser (approved):

PR #1417, opened 2025-04-29
fix: use_aws tool adds app name to the AWS CLI user agent string

clareliguori:
The AWS CLI allows you to add information to the user agent string by setting the AWS_EXECUTION_ENV env var. This string then shows up in CloudTrail logs to enable users to audit their Q CLI usage.

This PR adds a string like "AmazonQ-For-CLI Version/0.0.0" to the user agent of AWS CLI commands executed by the use_aws tool. This is the same string as used in [crates/fig_aws_common/src/user_agent_override_interceptor.rs](https://github.com/aws/amazon-q-developer-cli/blob/f0b5a0a3263c0f901d5c685f74117d2c84dff8f4/crates/fig_aws_common/src/user_agent_override_interceptor.rs) for requests to the Q Dev backend.

### DIFF
--- a/crates/q_chat/src/tools/use_aws.rs
+++ b/crates/q_chat/src/tools/use_aws.rs
@@ -26,6 +26,12 @@ use super::{
 
 const READONLY_OPS: [&str; 6] = ["get", "describe", "list", "ls", "search", "batch_get"];
 
+/// The environment variable name where we set additional metadata for the AWS CLI user agent.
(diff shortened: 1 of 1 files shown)


clareliguori:
Note: The test for the invoke method is marked as "not in ci", but does pass on my laptop.

brandonskiser (approved):

brandonskiser:
Thank you for the contribution!

PR #1442, opened 2025-04-30
fix: make default for fuzzy find ctrl + s

bergjaak:
*Issue #, if available:*

*Description of changes:* I have seen a good number of complaints that the default choice for opening fuzzy find menu is annoying, because emacs users like to use ctrl + k to kill their current line of text. So, we can do ctrl + s as the default to avoid annoying users. However, it is possible to change the key bind using q settings chat (e.g.,`q settings chat.skimCommandKey k` to make the skim menu opened by ctrl + k).

Also, remove /profile add because that's not a real command. Also add context hooks because those are commands.


bergjaak:
macos tests have been broken for awhile

brandonskiser (approved):

PR #1446, opened 2025-04-30
Add terminal notification tip to rotating tips

saisrinivas26:
*Issue #, if available:*

*Description of changes:*

Add a new tip to the rotating tips array that informs users about enabling terminal notifications with the settings command.

Screenshot 2025-05-01 at 11 40 00 AM


brandonskiser (approved):

brandonskiser:
I think I would prefer a message that more clearly says what "terminal notifications" even mean, like `Get notified when Q completes a response with command ...`
@chaynabors thoughts?

saisrinivas26:
> I think I would prefer a message that more clearly says what "terminal notifications" even mean, like `Get notified when Q completes a response with command ...`

I think makes sense if user is new and doesn't now what to make of word "terminal notifications"

Issue #1232, opened 2025-04-14
Support generating code from image


ranrmak:
CLI to support generating code from an input image

Issue #952, opened 2025-03-26
Enable image input support in Amazon Q CLI for enhanced visual development workflows


dpmramesh:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.3.2 (24D81)

### Expected behaviour

# Background
Amazon Q CLI currently lacks support for processing image inputs, limiting its ability to assist with visual development workflows.

# Problem Statement
Developers and designers cannot leverage Amazon Q CLI's capabilities for image-based tasks, such as:
- UI component generation from screenshots
- Design review assistance with diagrams
- Visual asset analysis and suggestions

# Proposed Solution
Add image input support to Amazon Q CLI to enable:
1. Direct image file processing
2. Screenshot analysis
3. Visual context understanding
4. UI component suggestions based on visual inputs

# Benefits
- Improved efficiency for UI/UX development workflows
- Enhanced collaboration between design and development teams
- Faster prototyping and iteration cycles
- Better support for visual documentation analysis

# Success Criteria
- Amazon Q CLI can accept and process image file inputs
- Support for common image formats (JPG, PNG, etc.)
- Ability to generate relevant code suggestions from visual inputs
- Integration with existing development workflows

# Implementation Considerations
- Image processing capabilities in the backend
- Security and privacy considerations for image data
- Performance impact on the CLI
- Integration with existing Amazon Q features

### Actual behaviour

Feature not supported

### Environment

```yaml
[q-details]
version = "1.7.2"
hash = "b6bf5f1675dd7395009e8bc65c1145820671f211"
date = "2025-03-21T01:36:16.66764Z (6d ago)"
variant = "full"

[system-info]
os = "macOS 15.3.2 (24D81)"
chip = "Apple M1 Pro"
total-cores = 10
memory = "32.00 GB"
```

mschrage:
The API currently supports this. We need to add give a mechanism for users to include images in requests (UX) and then encode it as part of the request.

This could be a `/image` command and then pass the path to a file? Open to ideas and discussion

sagarspatil:
Is anyone working on this?

r0b0ji:
Or maybe /context add_image /context add_video /context add_audio

jief123:
[...]

PR #1464, opened 2025-05-01
feat: image inputs

GoodluckH:
#952 
#1232 

https://github.com/user-attachments/assets/2bc4e850-8684-40e6-be2f-76e04fc1928f

### DIFF
--- a/crates/fig_api_client/src/model.rs
+++ b/crates/fig_api_client/src/model.rs
@@ -1,4 +1,7 @@
-use aws_smithy_types::Document;
+use aws_smithy_types::{
+    Blob,
+    Document,
+};
 use serde::{
     Deserialize,
     Serialize,
@@ -642,12 +645,14 @@ pub struct UserInputMessage {
     pub content: String,
     pub user_input_message_context: Option<UserInputMessageContext>,
     pub user_intent: Option<UserIntent>,
+    pub images: Option<Vec<ImageBlock>>,
 }
 
 impl From<UserInputMessage> for amzn_codewhisperer_streaming_client::types::UserInputMessage {
     fn from(value: UserInputMessage) -> Self {
         Self::builder()
             .content(value.content)
+            .set_images(value.images.map(|images| images.into_iter().map(Into::into).collect()))
             .set_user_input_message_context(value.user_input_message_context.map(Into::into))
             .set_user_intent(value.user_intent.map(Into::into))
             .origin(amzn_codewhisperer_streaming_client::types::Origin::Cli)
@@ -660,6 +665,7 @@ impl From<UserInputMessage> for amzn_qdeveloper_streaming_client::types::UserInp
     fn from(value: UserInputMessage) -> Self {
         Self::builder()
             .content(value.content)
+            .set_images(value.images.map(|images| images.into_iter().map(Into::into).collect()))
             .set_user_input_message_context(value.user_input_message_context.map(Into::into))
             .set_user_intent(value.user_intent.map(Into::into))
             .origin(amzn_qdeveloper_streaming_client::types::Origin::Cli)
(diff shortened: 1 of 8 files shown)


crates/fig_api_client/src/model.rs
```diff
@@ -701,6 +707,85 @@ impl From<UserInputMessageContext> for amzn_qdeveloper_streaming_client::types::
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ImageBlock {
+    pub format: ImageFormat,
+    pub source: ImageSource,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ImageFormat {
```
jsamuel1, line 717: Can we use Mime types instead?   There are hundreds of data formats that we'll want to support, and maintaining a list like this will not be very maintainable.

crates/fig_api_client/src/model.rs
```diff
@@ -701,6 +707,85 @@ impl From<UserInputMessageContext> for amzn_qdeveloper_streaming_client::types::
...
+    Gif,
+    Jpeg,
+    Png,
+    Webp,
+}
+
+impl From<ImageFormat> for amzn_codewhisperer_streaming_client::types::ImageFormat {
+    fn from(value: ImageFormat) -> Self {
+        match value {
```
jsamuel1, line 726: What about mov, mp4,  iphone hiac images, svg, ...    we need extensibility here.  Is there a library that can do this for us?

crates/q_chat/src/consts.rs
```diff
@@ -17,3 +17,8 @@ pub const MAX_USER_MESSAGE_SIZE: usize = 600_000;
 pub const CONTEXT_WINDOW_SIZE: usize = 200_000;
 
 pub const MAX_CHARS: usize = TokenCounter::token_to_chars(CONTEXT_WINDOW_SIZE); // Character-based warning threshold
+
+pub const MAX_NUMBER_OF_IMAGES_PER_REQUEST: usize = 10;
+
+/// In bytes - 10 MB
+pub const MAX_IMAGE_SIZE: usize = 10 * 1024 * 1024;
```
jsamuel1, line 24: 10MB wont handle many screen captures on high-res monitors.  Is this constraint related to some LLM constraint?  If so, can we define it based on what the backend can handle?

crates/q_chat/src/conversation_state.rs
```diff
@@ -8,6 +8,7 @@ use fig_api_client::model::{
     AssistantResponseMessage,
     ChatMessage,
     ConversationState as FigConversationState,
+    ImageBlock,
```
jsamuel1, line 11: Is this able to be an AttachmentBlock, rather than ImageBlock, so that other attachment types are a small change only?  (eg. movies, voice, PDF, word documents, ...)

crates/q_chat/src/lib.rs
```diff
@@ -1366,26 +1367,56 @@ impl ChatContext {
...
                 if self.interactive {
+                    if !&valid_images.is_empty() {
+                        execute!(
+                            self.output,
+                            style::SetForegroundColor(Color::Cyan),
+                            style::Print("\nImages detected:\n"),
+                            style::SetForegroundColor(Color::Reset)
+                        )?;
+                        for (_, metadata) in &valid_images {
+                            let image_size_str = if metadata.size > 1024 * 1024 {
```
jsamuel1, line 1384: Is there a library call for human readable sizes?  Or should we define a new format macro for this, for reuse?

crates/q_chat/src/util/images.rs
```diff
@@ -0,0 +1,288 @@
...
+    MAX_IMAGE_SIZE,
+    MAX_NUMBER_OF_IMAGES_PER_REQUEST,
+};
+use crate::shared_writer::SharedWriter;
+
+#[derive(Clone, Debug)]
+pub struct ImageMetadata {
+    pub filepath: String,
+    pub size: u64, // in bytes
+    pub filename: String,
```
jsamuel1, line 25: Should we have the image type (mime type) here too?

crates/q_chat/src/message.rs
```diff
@@ -35,6 +36,7 @@ pub struct UserMessage {
     pub additional_context: String,
     pub env_context: UserEnvContext,
     pub content: UserMessageContent,
+    pub images: Option<Vec<ImageBlock>>,
```
jsamuel1, line 39: If this was attachments (where Image is one type of attachment), this would be awesome for extensibility here -- PDF, Word documents, movies, whatever.   If the LLM doesn't support a type, it will happily tell us. :) (Might be a future pathway for backend routing of attachment processing model before passing to the coding/thinking model)

PR #1477, opened 2025-05-02
feat: Include /context hooks and conversation summary in `/context show --expand`

evanliu048:
`*Issue` #, if available:*
https://github.com/aws/amazon-q-developer-cli/issues/1422

*Description of changes:*
This PR enhances the `/context show` command by embedding associated hooks (both global and profile) under their respective sections and displaying a conversation summary for. Improves clarity and user experience.

https://github.com/user-attachments/assets/b117e24f-1d9f-4d9f-a4f1-a02bac2eb0a7


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2632,6 +2670,45 @@ impl ChatContext {
         })
     }
 
+    // Prints hook configuration grouped by trigger: conversation sesiion start or per user message
+    fn print_hook_section(output: &mut impl Write, hooks: &HashMap<String, Hook>, trigger: HookTrigger) -> Result<()> {
```
brandonskiser, line 2674: This should take a `&mut self` receiver instead, otherwise let's place it as a standalone function rather than an associated function

Also for doc comments, use `///` instead of `//`

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1727,6 +1774,39 @@ impl ChatContext {
...
+                                            style::Print("                       CONVERSATION SUMMARY"),
+                                            style::Print("\n"),
+                                            style::Print(&border),
+                                            style::SetAttribute(Attribute::Reset),
+                                            style::Print("\n\n"),
+                                            style::Print(&summary),
+                                            style::Print("\n\n")
+                                        )?;
+                                    } else {
+                                        self.compact_history(
```
brandonskiser, line 1799: Definitely don't do this - just check if there is a summary cached, and if so print it.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1727,6 +1774,39 @@ impl ChatContext {
                                     style::Print(format!("\nTotal: ~{} tokens\n\n", total_tokens)),
                                 )?;
 
+                                // Show last cached conversation summary if available, otherwise regenerate it
+                                if expand {
+                                    if let Some(summary) =
+                                        self.conversation_state.latest_summary().map(|s| s.to_owned())
```
brandonskiser, line 1780: This is inside the `else` block of `if global_context_files.is_empty() && profile_context_files.is_empty() {`, so it won't ever get executed if both global and profile context matches nothing.

Also shouldn't be necessary to `to_owned` here, we should be able to just print a reference

brandonskiser (changes requested):

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Thanks for the suggestion! 🙏
You're right — since the method doesn't use any fields from self, I've moved it out as a standalone function and updated the doc comment to use `///`.

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Makes total sense. Moved the logic out of the else block and ditched the unnecessary `.to_owned()`. Learned something new again — thanks for the sharp eyes!

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1781,6 +1828,28 @@ impl ChatContext {
 
                                 execute!(self.output, style::Print("\n"))?;
                             }
+
+                            // Show last cached conversation summary if available, otherwise regenerate it
```
brandonskiser, line 1832: nit - remove the "regenerate it" comment

brandonskiser (approved):

PR #1489, opened 2025-05-02
Support images with fs_read

GoodluckH:
#1232

### DIFF
--- a/crates/chat-cli/src/api_client/model.rs
+++ b/crates/chat-cli/src/api_client/model.rs
@@ -1,4 +1,7 @@
-use aws_smithy_types::Document;
+use aws_smithy_types::{
+    Blob,
+    Document,
+};
 use serde::{
     Deserialize,
     Serialize,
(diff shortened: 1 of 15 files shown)


crates/q_chat/src/tools/tool_index.json
```diff
@@ -21,22 +21,30 @@
...
           "enum": [
             "Line",
             "Directory",
-            "Search"
+            "Search",
+            "Image"
           ],
-          "description": "The mode to run in: `Line`, `Directory`, `Search`. `Line` and `Search` are only for text files, and `Directory` is only for directories."
+          "description": "The mode to run in: `Line`, `Directory`, `Search`. `Line` and `Search` are only for text files, and `Directory` is only for directories. `Image` is for image files, in this mode `image_paths` is required."
         },
```
mschrage, line 48: Do we need to have a separate `Image` mode? Could we just detect if a path is an image based on the file extension? (I worry this may result in tool validation errors if the model is confused about how to use this tool?

crates/q_chat/src/lib.rs
```diff
@@ -2807,7 +2836,11 @@ impl ChatContext {
                     }
                     tool_results.push(ToolUseResult {
                         tool_use_id: tool.id,
-                        content: vec![result.into()],
+                        content: if image_blocks.is_empty() {
+                            vec![result.into()]
```
brandonskiser, line 2840: Instead we can implement `impl From<InvokeOutput> for Vec<ToolUseResultBlock>` so we don't need this logic here.

crates/q_chat/src/message.rs
```diff
@@ -58,6 +60,18 @@ impl UserMessage {
...
+            images: None,
+            additional_context: String::new(),
+            env_context: UserEnvContext::generate_new(),
+            content: UserMessageContent::Prompt { prompt },
+        }
+    }
+
+    #[allow(dead_code)]
+    /// Use this when we support extracting images from the prompt.
+    pub fn new_prompt_with_images(prompt: String, images: Vec<ImageBlock>) -> Self {
```
brandonskiser, line 72: let's remove this

crates/q_chat/src/util/images.rs
```diff
@@ -0,0 +1,389 @@
...
+            style::Print(format!(
+                "\nThe following images are dropped due to exceeding size limit ({}MB):\n",
+                MAX_IMAGE_SIZE / (1024 * 1024)
+            )),
+            style::SetForegroundColor(Color::Reset)
+        )
+        .ok();
+        for (_, metadata) in &images_exceeding_size_limit {
+            let image_size_str = if metadata.size > 1024 * 1024 {
+                format!("{:.2} MB", metadata.size as f64 / (1024.0 * 1024.0))
```
brandonskiser, line 88: maybe a small function/test for this formatting? Just since it's hard to verify manually

brandonskiser (changes requested):

crates/q_chat/src/tools/tool_index.json
brandonskiser: I think we definitely should have a separate mode for reading images, the logic for handling file paths vs images is different and I'm guessing it's better to be explicit here

crates/chat-cli/src/api_client/model.rs
```diff
@@ -565,17 +568,113 @@ impl From<GitState> for amzn_qdeveloper_streaming_client::types::GitState {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ImageFormat {
+    Gif,
+    Jpeg,
+    Png,
+    Webp,
+}
+
+impl std::str::FromStr for ImageFormat {
+    type Err = ();
```
brandonskiser, line 586: Use an error message instead? so we know what exactly failed to parse

brandonskiser (approved):

Issue #1741, opened 2025-05-08
[UX] Display spinner immediate after tool completed


GoodluckH:
> Sluggishness. There is always a long delay (1-2 seconds) between "Completed in XYZ seconds" and the "Thinking" spinner. It makes Q feel sluggish. Can you reduce the perceived gap by eliminating whatever it's going in that time or at least print "thinking'" right away after "Completed"?

https://amzn-aws.slack.com/archives/C064EVBE0LR/p1746716811774539

PR #1747, opened 2025-05-08
fix: display loading UI after tool execution

GoodluckH:
#1741

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -2868,6 +2868,11 @@ impl ChatContext {
         } else {
             self.conversation_state.add_tool_results(tool_results);
         }
+        if self.interactive {
+            execute!(self.output, cursor::Hide)?;
+            execute!(self.output, style::Print("\n"), style::SetAttribute(Attribute::Reset))?;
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2888,6 +2893,19 @@ impl ChatContext {
...
 
+        if self.interactive && self.spinner.is_some() {
+            drop(self.spinner.take());
+            queue!(
+                self.output,
+                style::SetForegroundColor(Color::Reset),
+                terminal::Clear(terminal::ClearType::CurrentLine),
+                cursor::MoveToColumn(0),
+                cursor::Show,
+                cursor::MoveUp(1),
```
brandonskiser, line 2904: why is the moveup(1) and clear required here?

brandonskiser (approved):

crates/chat-cli/src/cli/chat/mod.rs
GoodluckH: to make sure the two "thinking" are on the same line. I think somewhere in this function we printed a newline as a hack for something.

chaynabors (approved):

PR #1757, opened 2025-05-09
refactor: telemetry, state, and settings handling

chaynabors:
1. Async Channel Architecture - Replaced global state and mutex locks with a dedicated worker thread and message passing for non-blocking telemetry operations

2. Dependency Injection - Eliminated global singletons in favor of explicit dependencies for better testability, lifetime, and lifecycle management

3. Error Handling Improvements - Significantly improved error types and reduced error type sizes dramatically. This reduces the size of Result returned from almost every function in the codebase.

4. Database type: Greatly simplify local DB interaction and improve consistency across the codebase.

---

I've tested the fuck out of this but it's not enough. Will bug bash with the team.


crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -262,10 +265,11 @@ impl Highlighter for ChatHelper {
 }
 
 pub fn rl(
+    database: &Database,
     sender: std::sync::mpsc::Sender<Option<String>>,
     receiver: std::sync::mpsc::Receiver<Vec<String>>,
 ) -> Result<Editor<ChatHelper, DefaultHistory>> {
-    let edit_mode = match crate::settings::settings::get_string_opt("chat.editMode").as_deref() {
+    let edit_mode = match database.settings.get_string(Setting::ChatEditMode).as_deref() {
```
mschrage, line 272: Will the break the existing settings.json configuration?

crates/chat-cli/src/auth/builder_id.rs
```diff
@@ -52,24 +51,26 @@ use tracing::{
...
     SecretStore,
 };
-use crate::aws_common::app_name;
-use crate::telemetry::send_refresh_credentials;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum OAuthFlow {
     DeviceCode,
+    // This must remain backwards compatible
+    #[serde(rename = "PKCE")]
```
brandonskiser, line 65: Is this what was causing the different auth state while trying to login across the `q` and `qchat` binaries?

crates/chat-cli/src/database/sqlite_migrations/006_make_state_blob.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE state RENAME TO state_old;
+CREATE TABLE state (
+    key TEXT PRIMARY KEY,
+    value BLOB
```
brandonskiser, line 4: Why are we making the state table store BLOB's? Is this really required?

some discussion on this - https://sqlite-users.sqlite.narkive.com/sBgPvXrZ/using-text-vs-blob

brandonskiser (approved):

dingfeli (approved):

Issue #153, opened 2024-11-28
Windows support


grant0417:

wilikeit:
我也希望添加 windows支持。

JoeShi:
Here is the workaround of using Amazon Q Developer CLI on Windows.

**Step 1: Install [WSL](https://learn.microsoft.com/en-us/windows/wsl/install)**

- Run CMD or PowerShell in **administrator** mode, execute `wsl --install`. After the installation is finished, you will need to restart your computer.
- Upon restarting, the installation of Ubuntu will begin automatically. You will be able to see Ubuntu and WSL in the Start menu.
- Launch Ubuntu. Upon first login, you will need to set up a username and password.


**Step 2: Install Q Developer CLI in WSL**

All the following command needs to be executed in WSL.

- Update the apt. `sudo apt update`
- Download the package. `curl --proto '=https' --tlsv1.2 -sSf https://desktop-release.q.us-east-1.amazonaws.com/latest/amazon-q.deb -o amazon-q.deb`
- Install the package. `sudo apt install -y ./amazon-q.deb`

**Step 3: Authentication with Q Developer CLI**

- Type `q login`.  You will be see an URL. 
- Open the URL, input your credentials, and click "Allow Access".
- You will see Logged in successfully.

**Step 4: Enter the agentic chat mode**

- Type `q chat`

Now you are in the agentic chat mode. Enjoy the agent.

wilikeit:
Think you.This is a good idea, but it seems that the auto-completion function cannot be used in WSL.

atxtechbro:
+1 to @wilikeit's comment about auto-completion not working in WSL.

I'm experiencing the exact same issue with Amazon Q in WSL2 (Ubuntu 24.04.1 LTS). The inline completions simply don't appear despite:
- `q inline status` showing that inline is enabled
- `q doctor` reporting that everything looks good
- All other Amazon Q features like `q chat` working perfectly

I've tried various troubleshooting steps including disabling/re-enabling inline, testing in a minimal bash environment, and restarting WSL, but the issue persists.

Despite this limitation, I've been having a wonderful experience using Amazon Q in WSL for everything else. It's been powerful enough that I've created a guide for my company to adopt it. The tool works great on native Linux installations, but there's a definitive difference in the auto-completion functionality when using it in Windows with WSL setups.

Would love to see this fixed as it would make an already excellent tool even better in the WSL environment!

grepit:
[...]

PR #1765, opened 2025-05-10
feat: support Windows processes in `mcp_client`

cwoolum:
# Cross-Platform Process Management for Windows Support

This PR adds cross-platform process management capabilities to support Windows in the Amazon Q Developer CLI. It addresses build issues that were occurring on Windows due to Unix-specific code.

### Process Information Module
- Added `from_raw` method to the `Pid` struct for consistent process ID handling across platforms
- Implemented platform-specific `terminate_process` functions in their respective OS modules:
  - Uses `kill` with `SIGTERM` on Unix platforms (Linux and macOS)
  - UsesWindows API to kill processes on Windows platforms
- Updated module exports to properly expose platform-specific implementations

### MCP Client
- Made `process_group` method conditional on Unix platforms
- Updated process termination to use the cross-platform `terminate_process` function
- Added fig_util as a dependency to mcp_client

## Testing
- Verified that the code builds successfully on Windows

Continues work on https://github.com/aws/amazon-q-developer-cli/issues/153

### DIFF
--- a/crates/chat-cli/src/mcp_client/client.rs
+++ b/crates/chat-cli/src/mcp_client/client.rs
@@ -541,6 +541,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    #[ignore]
     async fn test_client_stdio() {
         std::process::Command::new("cargo")
             .args(["build", "--bin", TEST_SERVER_NAME])
--- a/crates/fig_auth/src/builder_id.rs
+++ b/crates/fig_auth/src/builder_id.rs
@@ -138,7 +138,7 @@ impl DeviceRegistration {
     }
 
     /// Loads the OIDC registered client from the secret store, deleting it if it is expired.
-    async fn load_from_secret_store(secret_store: &SecretStore, region: &Region) -> Result<Option<Self>> {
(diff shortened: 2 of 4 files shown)


crates/fig_util/src/process_info/linux.rs
```diff
@@ -10,6 +12,11 @@ use super::{
     PidExt,
 };
 
+pub fn terminate_process(pid: Pid) -> Result<(), String> {
```
cwoolum, line 15: I duplicated this code for Unix and macOS since there are individual files for each OS. Not in love with it but it felt more cohesive than doing the OS specific logic in `mod.rs`.

crates/mcp_client/src/client.rs
```diff
@@ -170,12 +175,7 @@ impl Client<StdioTransport> {
             command.args(args).spawn()?
         };
         let server_process_id = child.id().ok_or(ClientError::MissingProcessId)?;
-        #[allow(clippy::map_err_ignore)]
-        let server_process_id = Pid::from_raw(
-            server_process_id
-                .try_into()
-                .map_err(|_| ClientError::MissingProcessId)?,
```
cwoolum, line 177: Accidentally dropped this. Need to fix

brandonskiser (changes requested):

Issue #1466, opened 2025-05-01
Background server loading


dingfeli:
## Target behavior
- Server loading should be optional for a given message. Users should be able to choose to not wait for servers to load before sending a message. 
- Messages sent at a given time will be contextualized with the available servers at that time. 
- Users should be able to check server statuses.

## Some background
Today, the MCP lifecycle in q cli is _tool centric_. This means that of all the features that are part of the MCP protocol (at the time of writing these are prompts, resources and tools), tools are seen as a must-have. If a server is to offer no tools, the cli will see that as an error and terminate the init process. This actually defies what is expected of a client by the protocol and as a result, prevents servers that offer no tools from being loaded (e.g. https://github.com/aws/amazon-q-developer-cli/issues/1465). 

What would be less surprising to server developers (and perhaps end users) is if we relegate tools and treat them the same as any other server features. 

## What does this have to do with background server loading
Because of the aforementioned special treatment for tools, `tools/list` calls is done as part of the server loading: 
https://github.com/aws/amazon-q-developer-cli/blob/df007271b6f77c1acd1a83b506a5a69a8a87f73c/crates/q_chat/src/tools/custom_tool.rs#L90-L115
If we were to relegate tools feature (i.e. downgrade it to the same level as prompts and resources), it would make sense to delegate `tools/list` to the MCP client (similar to where prompts are prefetched and cached, as opposed to a client consumer activity done on the cli), which would be done in a task that would already put this process in the background. 

Currently the init is done as such:

<img width="1010" alt="Image" src="https://github.com/user-attachments/assets/20a9ccd2-e72b-4847-b16c-1ae4723c606f" />

If we relegate tools, it should looking something like this:

<img width="1001" alt="Image" src="https://github.com/user-attachments/assets/5a67f734-5cc9-4fec-8c7f-581dc72ac680" />

PR #1775, opened 2025-05-12
Background server load

dingfeli:
*Issue #, if available:*
https://github.com/aws/amazon-q-developer-cli/issues/1466

*Description of changes:*
**Enables background server load**:
- As described in https://github.com/aws/amazon-q-developer-cli/issues/1466, relegates `tools/list` to the same level as prompts, which are also fetched in the background. 
- Enables users to set timeout for servers init, beyond which the stragglers continue to load in the background, enabling users to start typing. This is done via `q settings mcp.initTimeout`. I have opted to *not* have this in `mcp.json` because this might conflict with how IDE interprets the config.
- Enables users to interrupt server loading (even with timeout set) to immediately start conversation. 
- `/tools` now show servers still in loading.

**Removal of mcp_client crate**:
- No longer needed since the refactor

**Refinement of conversation invariant logic with regards to tool calls with incorrect names**:
There are three scenarios in which intervention would be needed:
1. The model had decided to call a tool with its partial name AND there is only one such tool, in
which case we would automatically resolve this tool call to its correct name. This will NOT
result in an error in its tool result. The intervention here is to substitute the partial name
with its full name.
2. The model had decided to call a tool with its partial name AND there are multiple tools it
could be referring to, in which case we WILL return an error in the tool result. The
intervention here is to substitute the ambiguous, partial name with a dummy.
3. The model had decided to call a tool that does not exist. The intervention here is to
substitute the non-existent tool name with a dummy.

**List changed support for prompts and tools**:
- List changed support for prompts and tools.
- This is added because it is enabled by background server load. 


crates/chat-cli/src/telemetry/mod.rs
```diff
@@ -159,7 +162,7 @@ impl TelemetryThread {
     pub async fn finish(self) -> Result<(), TelemetryError> {
         drop(self.tx);
         if let Some(handle) = self.handle {
-            handle.await?;
+            let _ = tokio::time::timeout(std::time::Duration::from_millis(100), handle).await;
```
chaynabors, line 165: can we change this to 1000ms and keep the error surfacing

chaynabors (changes requested):

crates/chat-cli/src/telemetry/mod.rs
dingfeli: sounds good

brandonskiser (dismissed):

brandonskiser (approved):

evanliu048 (approved):

PR #1860, opened 2025-05-16
* feat(chat): Add semantic search crate to Q CLI

kensave:
Add a new semantic_search_client crate that provides vector e…

…mbedding and

    semantic search capabilities for the Amazon Q CLI. This implementation:

    - Supports text embedding generation using Candle and ONNX runtimes and BM25
    - Provides hardware acceleration via Metal on macOS
    - Implements efficient vector indexing for semantic search
    - Includes file processing utilities for various file types
    - Supports persistent storage of semantic contexts
    - Includes comprehensive test coverage

    This crate will enable memory bank functionality for Amazon Q, allowing
    users to create, manage, and search through semantic memory contexts.

### DIFF
--- a/.lintstagedrc.mjs
+++ b/.lintstagedrc.mjs
@@ -8,5 +8,5 @@ export default {
   ],
   "*.py": ["ruff format --check", "ruff check"],
   "*.{ts,js,tsx,jsx,mjs}": "prettier --check",
-  "!(*test*)*": "typos",
+  "!(*test*)*": "typos --config .typos.toml",
 };
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ clap = { version = "4.5.32", features = [
     "unicode",
     "wrap_help",
 ] }
+chrono = { version = "0.4", features = ["serde"] }
 cocoa = "0.26.0"
 color-print = "0.3.5"
 convert_case = "0.8.0"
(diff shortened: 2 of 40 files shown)


Cargo.toml
```diff
@@ -102,12 +103,14 @@ objc2 = "0.5.2"
 objc2-app-kit = "0.2.2"
 objc2-foundation = "0.2.2"
 objc2-input-method-kit = "0.2.2"
+once_cell = "1.19.0"
```
brandonskiser, line 106: Why use once_cell? Is using the `std` types not enough?

brandonskiser (changes requested):

Cargo.toml
kensave: Actually this is a good one. Rust 1,7+ I could use it instead.

.lintstagedrc.mjs
```diff
@@ -8,5 +8,5 @@ export default {
   ],
   "*.py": ["ruff format --check", "ruff check"],
   "*.{ts,js,tsx,jsx,mjs}": "prettier --check",
-  "!(*test*)*": "typos",
+  "!(*test*)*": "typos --config .typos.toml",
```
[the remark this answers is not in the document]
kensave: Let me check. I think had many issues with this one.

brandonskiser (approved):

Issue #1212, opened 2025-04-13
Feature request: Support newlines with Shift+Return in chat


TrevorBurnham:
In chat, entering Shift+Return currently submits my prompt. I'd like to be able to enter newline characters with Shift+Return instead to facilitate multi-line prompts. This is a well-established convention in web interfaces (see e.g. Claude).

Of course, the new `/ed` feature to allow prompts to be entered in a proper editor (#1029) is much appreciated, but Shift+Return could save me from some context-switching.

carlesso:
The latest version (1.7.3) introduced multiline support:
```
Use Ctrl(^) + j to provide multi-line prompts.

> 
```

See #1044. Is that the behavior you would expect?

TrevorBurnham:
Nice! I didn't know about Ctrl+J; that's a great improvement. I see that Alt+Enter has the same effect. I'd still like to see Shift+Return do the same thing, though.

carlesso:
That make sense. It should be just a simple addition here: https://github.com/aws/amazon-q-developer-cli/blob/main/crates/q_cli/src/cli/chat/prompt.rs#L238-L248

mschrage:
Shift+Enter is not a keybinding that is supported in most terminals.

mdinescu:
Big +1 to support Shift + Enter.  Muscle memory is strong and I found myself repeatedly submitting incomplete prompts due to attempting to press Shit + Enter.

soleares:
Shift+Enter is used in every other AI editor I use including: Q Chat (JetBrains), Cline, ChatGPT. Q CLI is the only one I need to use a different key combination Alt+Enter for.

TrevorBurnham:
I was playing around with Claude Code and noticed that it offered this prompt when running it for the first time in VS Code:

```
╭──────────────────────────╮
│ ✻ Welcome to Claude Code │
╰──────────────────────────╯

 Use Claude Code's terminal setup?

 For the optimal coding experience, enable the recommended settings
 for your terminal: Shift+Enter for newlines

 ❯ 1. Yes, use recommended settings
   2. No, maybe later with /terminal-setup
```

The [Claude Code docs](https://docs.anthropic.com/en/docs/claude-code/terminal-config) elaborate that Shift+Enter can be enabled in VS Code and iTerm2.

It'd be nice for Q CLI to have feature parity with Claude Code here.

yxzwayne:
[...]

PR #1865, opened 2025-05-16
feat(chat): Add Shift+Enter for newlines (#1212)

s12v:
*Issue #:* https://github.com/aws/amazon-q-developer-cli/issues/1212

*Description of changes:*

Add Shift+Enter as an additional keyboard shortcut for inserting newlines in chat prompts, alongside the existing Ctrl+J and Alt+Enter options. This follows common conventions in web interfaces and improves user experience.

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -220,9 +220,9 @@ const ROTATING_TIPS: [&str; 12] = [
 
 const GREETING_BREAK_POINT: usize = 80;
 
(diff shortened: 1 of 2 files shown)


brandonskiser:
This doesn't seem to work for me in WezTerm or macOS Terminal. What terminals have you tested with?

As far as I can tell, shift+enter isn't a supported keybinding in most terminals, so I don't think this PR can really be merged in any reasonable capacity.

s12v:
@brandonskiser, you're absolutely right, thanks a lot for catching this!

Looks like there's no feasible way to get this done. Closing the PR.

PR #1875, opened 2025-05-17
feat(windows): implement multi platform process kill

cwoolum:
## Summary
This PR enhances the Amazon Q CLI with cross-platform process termination capabilities and optimizes Windows performance through stack size adjustments. These improvements ensure consistent behavior across operating systems while addressing Windows-specific memory constraints.

## Changes
• Added Windows-specific implementation for process termination
• Ensured consistent behavior between Windows, macOS, and Linux platforms
• Implemented proper error handling for process termination failures
• **Increased thread stack size allocation for Windows environments**
• Optimized memory management for Windows-specific constraints

### Process Termination
• Implemented platform-specific code paths for process termination
• Used Windows API for Windows and POSIX signals for Unix-based systems
• Added proper error handling and logging for failed termination attempts

### Stack Size Optimization
• Increased default thread stack size on Windows to prevent stack overflow errors
• Windows has different memory management characteristics compared to Unix-based systems, requiring larger stack allocations
• This change prevents crashes during deep recursion or when handling complex data structures
• The stack size increase is crucial for stable operation on Windows, particularly when processing large amounts of data or during intensive operations

### Default Stack Size Differences
• **Windows default stack size**: 1MB for user threads (significantly smaller than Linux)
• **Linux default stack size**: 8MB for user threads
• This 8x difference is critical when running memory-intensive operations
• Windows' smaller default stack size can lead to stack overflow errors in scenarios that work fine on Linux
• Our increase aligns the Windows stack size more closely with Linux defaults, ensuring consistent behavior across platforms
• This adjustment is particularly important for recursive operations and deep call stacks that are common in our CLI operations

## Testing
• Verified process termination works correctly on Windows systems
• Confirmed backward compatibility with existing Unix/Linux implementations

### DIFF
--- a/crates/chat-cli/src/util/process/windows.rs
+++ b/crates/chat-cli/src/util/process/windows.rs
@@ -1,0 +1,53 @@
+use std::ops::Deref;
+
+use sysinfo::Pid;
+use windows::Win32::Foundation::{
+    CloseHandle,
+    HANDLE,
+};
+use windows::Win32::System::Threading::{
+    OpenProcess,
+    PROCESS_TERMINATE,
+    TerminateProcess,
+};
+
+/// Terminate a process on Windows using the Windows API
+pub fn terminate_process(pid: Pid) -> Result<(), String> {
+    unsafe {
+        // Open the process with termination rights
+        let handle = OpenProcess(PROCESS_TERMINATE, false, pid.as_u32())
+            .map_err(|e| format!("Failed to open process: {}", e))?;
+
+        // Create a safe handle that will be closed automatically when dropped
+        let safe_handle = SafeHandle::new(handle).ok_or_else(|| "Invalid process handle".to_string())?;
+
+        // Terminate the process with exit code 1
+        TerminateProcess(*safe_handle, 1).map_err(|e| format!("Failed to terminate process: {}", e))?;
+
+        Ok(())
+    }
+}
+
+struct SafeHandle(HANDLE);
+
+impl SafeHandle {
+    fn new(handle: HANDLE) -> Option<Self> {
+        if !handle.is_invalid() { Some(Self(handle)) } else { None }
+    }
+}
+
+impl Drop for SafeHandle {
+    fn drop(&mut self) {
+        unsafe {
+            let _ = CloseHandle(self.0);
+        }
+    }
+}
+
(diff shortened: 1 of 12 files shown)


crates/chat-cli/src/util/process/windows.rs
```diff
@@ -0,0 +1,121 @@
...
+impl Deref for SafeHandle {
+    type Target = HANDLE;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[cfg(test)]
+#[cfg(windows)]
```
brandonskiser, line 56: nit - `#[cfg(windows)]` shouldn't be required here since this module is only included on windows, right?

crates/chat-cli/src/util/process/windows.rs
```diff
@@ -0,0 +1,121 @@
...
+mod tests {
+    use std::process::Command;
+    use std::time::Duration;
+
+    use super::*;
+
+    // Helper to create a long-running process for testing
+    fn spawn_test_process() -> std::process::Child {
+        let mut command = Command::new("cmd");
```
brandonskiser, line 65: What is this command doing? Would rather tests not do stuff like send network requests to localhost if we can rather do a `sleep 10000` or something otherwise

crates/chat-cli/src/util/system_info/mod.rs
```diff
@@ -10,6 +13,23 @@ use serde::{
 
 use crate::platform::Env;
 
+/// Fields for OS release information
```
brandonskiser, line 16: Can we include the reference to the Linux manpage still in this doc comment? https://www.man7.org/linux/man-pages/man5/os-release.5.html

brandonskiser (dismissed):

crates/chat-cli/src/util/process/windows.rs
cwoolum: Good call, no need to make network calls. Changed to be a timeout instead

crates/chat-cli/src/util/system_info/mod.rs
cwoolum: Done

brandonskiser (approved):

Issue #1918, opened 2025-05-20
inconsistency in /save and /load behavior with regard to file extensions


bcchipm:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.4.1 (24E263)

### Expected behaviour

/save does not require .json and auto appends it if absent
```
!> /save convo

✔ Exported conversation state to convo.json
```

/load requires the .json file extension
```
!> /load convo

Failed to import from convo: No such file or directory (os error 2)

!> /load convo.json

✔ Imported conversation state from convo.json
```

Proposed behavior change: automatically try `<fileinput>.json` if no file extension provided, e.g.:
* search for `<fileinput>` , if exists, try to load
if `<fileinput>` file does not exist
search for `<fileinput>.json` , if exists try to load

### Actual behaviour

See other section

### Environment

```yaml
<This will be visible to anyone. Do not include personal or sensitive information>

[q-details]
version = "1.10.1"
hash = "cb90e381cf8391beef4b966623d913e258cbb08d"
date = "2025-05-16T13:53:39.016927Z (4d ago)"
variant = "full"

[system-info]
os = "macOS 15.4.1 (24E263)"
chip = "Apple M1 Pro"
total-cores = 10
memory = "32.00 GB"

[environment]
cwd = "/Users/USER/Work/AmazonQ/src/AmazonBuilderGenAIPowerUsersQContext"
cli-path = "/Users/USER/Work/AmazonQ/src/AmazonBuilderGenAIPowerUsersQContext"
os = "Mac"
shell-path = "/Users/USER/.toolbox/tools/toolbox/1.1.1876.0/toolbox-exec"
shell-version = "5.9"
terminal = "iTerm 2"
install-method = "toolbox"

[env-vars]
[...]

PR #1939, opened 2025-05-22
fix: inconsistency in /save and /load behavior with regard to file ex…

Konippi:
*Issue #, if available:*
resolve: #1918

*Description of changes:*
This PR allows q chat to automatically grant .json and retry if the .json extension is not present when the load command is executed.


konippi:
move to https://github.com/aws/amazon-q-developer-cli-autocomplete/pull/289

PR #2019, opened 2025-06-08
chat: Add global context folder.

suminhong:
*Issue #, if available:*
hi. I was looking for a folder to set global rules, but realized that only project-specific settings were possible.
Of course, i can add a context via the `/context add --global` command, but I want all .md files under "~/.amazonq/rules/" to be read automatically.

*Description of changes:*
When load_global_config, add the path to read md files. : `~/.amazonq/rules/**/*.md`


dingfeli:
Thank you for your contribution.

We have since then employed a new way of configuring a session via agent. By default (i.e. the default agent assumed by q cli) does include this path and will read everything in its context. For that reason I am closing this PR. 

Please feel free to contact us if you think the current behavior is not quite what you are trying to achieve in this PR and we will be happy to take another look at it.

Issue #1804, opened 2025-05-13
[Feature] Shortcuts/aliases for commands


fsmiamoto:
Hi team,

I'm a newcomer to the Q CLI but already had the chance to get a lot of value out of it so thank you so much for the great work.

I like to use the /editor command a lot and was wondering if there could be some faster way of doing it. 

I could think of two ways to do it and wanted to get some feedback:
- Have command shortcuts e.g. `/e == /editor`
- Or have a Ctrl-E (or something else) mapped to /editor

I assume other users might have their own favorite commands so maybe we can think of a more scalable way to handle this.

In any case, I'd be happy to contribute with the implementation a potential solution.

PR #2041, opened 2025-06-11
Add keybinding for opening current prompt in $EDITOR

fsmiamoto:
### Why this change?

A lot of times I want to do a big edit or re-format my prompt and using the editor is a great option for that.

But if I already started typing the prompt them I have to clear everything and use the /editor command.

With this change, you can simply press `Ctrl + f` to open the current prompt as is in your $EDITOR and make your desired changes

Related to https://github.com/aws/amazon-q-developer-cli/issues/1804

### DIFF
--- a/crates/cli/src/cli/chat/editor_integration.rs
+++ b/crates/cli/src/cli/chat/editor_integration.rs
@@ -1,0 +1,366 @@
+use std::env;
+use std::fs;
+use std::process::Command;
+
+use eyre::{Result, eyre};
+use rustyline::{Cmd, ConditionalEventHandler, EventContext, Movement, RepeatCount};
+use uuid::Uuid;
+
+/// Handler for Ctrl+F keyboard shortcut that opens the current prompt content in an editor
+pub struct EditorLauncher;
+
+impl EditorLauncher {
+    pub fn new() -> Self {
+        Self
+    }
+
+    /// Create a command that replaces the entire line with new content
+    /// Uses a comprehensive approach to handle different scenarios
+    fn create_line_replacement_command(new_content: &str, current_text: &str, cursor_pos: usize) -> Option<Cmd> {
+        // Strategy: Use the best available approach based on rustyline capabilities
+
+        if new_content.is_empty() {
+            // User wants to clear the line
+            if current_text.is_empty() {
+                Some(Cmd::Noop) // Nothing to do
+            } else {
+                // Clear the entire line - move to beginning and kill to end
+                // This should clear all content on the current line
+                Some(Cmd::Kill(Movement::BeginningOfLine))
+            }
+        } else if current_text.is_empty() {
+            // Current line is empty, just insert new content
+            Some(Cmd::Insert(1, new_content.to_string()))
+        } else {
+            // Need to replace existing content with new content
+            //
+            // The most reliable approach: use Kill to clear from beginning to end,
+            // and then insert the new content
+            //
+            // But since we can only return one command, let's try a different approach:
+            // Use the cursor position to calculate how to best replace content
+
+            if cursor_pos == 0 {
+                // Cursor is at beginning - kill to end and insert
+                // But we can only do one command - let's try Replace with EndOfLine
+                Some(Cmd::Replace(Movement::EndOfLine, Some(new_content.to_string())))
+            } else if cursor_pos >= current_text.len() {
+                // Cursor is at end - move to beginning and replace all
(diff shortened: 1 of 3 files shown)


crates/cli/src/cli/chat/prompt.rs
```diff
@@ -300,6 +302,12 @@ pub fn rl(
         EventHandler::Simple(Cmd::Insert(1, "\n".to_string())),
     );
 
+    // Add custom keybinding for Ctrl+F to open current content in editor
+    rl.bind_sequence(
+        KeyEvent(KeyCode::Char('f'), Modifiers::CTRL),
```
l0ngvh, line 307: Some terminals by default bind CTRL+F to find, what about CTLR+E?

crates/cli/src/cli/chat/editor_integration.rs
```diff
@@ -0,0 +1,366 @@
...
+pub struct EditorLauncher;
+
+impl EditorLauncher {
+    pub fn new() -> Self {
+        Self
+    }
+
+    /// Create a command that replaces the entire line with new content
+    /// Uses a comprehensive approach to handle different scenarios
+    fn create_line_replacement_command(new_content: &str, current_text: &str, cursor_pos: usize) -> Option<Cmd> {
```
l0ngvh, line 19: It seems we don't need to have this returns an `Option`?

crates/cli/src/cli/chat/editor_integration.rs
```diff
@@ -0,0 +1,366 @@
...
+                } else {
+                    // Replace the entire line with edited content
+                    // Strategy: Move to beginning of line, kill everything to end, then insert new content
+                    Self::create_line_replacement_command(&edited_content, current_text, cursor_pos)
+                }
+            },
+            Ok(None) => {
+                // User cleared all content in editor - clear the entire line
+                Self::create_line_replacement_command("", current_text, cursor_pos)
+            },
```
l0ngvh, line 131: I think we can group these two cases? If trimmed `edited_content` is empty we can just write it instead of having `launch_system_editor` returns `None` and specifically handle it

crates/cli/src/cli/chat/editor_integration.rs
```diff
@@ -0,0 +1,366 @@
...
+
+        // Create a temporary directory for our test
+        let temp_dir = tempdir().expect("Failed to create temp dir");
+        let mock_editor_path = temp_dir.path().join("mock_editor.sh");
+
+        // Create a mock editor script that adds text to the file
+        let mock_editor_script = r#"#!/bin/bash
+echo "Edited: $(cat "$1")" > "$1"
+"#;
+        fs::write(&mock_editor_path, mock_editor_script).expect("Failed to write mock editor");
```
l0ngvh, line 165: It seems we haven't cleaned up this script?

crates/cli/src/cli/chat/editor_integration.rs
fsmiamoto: Good catch, let me address this

crates/cli/src/cli/chat/prompt.rs
fsmiamoto: Ctrl + E is also a default binding to go to the end of the line in most shells for that's why I went for Ctrl + F.

Ideally this is configurable though 🤔

crates/cli/src/cli/chat/editor_integration.rs
fsmiamoto: Hmm fair point, since it is created in a temp dir, I think it would be okay not to clean it up but let me add a clean up logic.

crates/cli/src/cli/chat/editor_integration.rs
fsmiamoto: Good point, let me review this.

fsmiamoto:
There are various scenarios where the current code breaks, mostly due to only being able to return one Cmd to rustyline.

So depending on the cursor position and the lengths of the current and new prompt, this does not have the expected behavior of completely replacing the content.

Need to consider other solutions for this.

brandonskiser:
Hey @fsmiamoto  apologies for such a late response here - since this is so out of date, would you still be interested in updating this PR or should we close this?

fsmiamoto:
Hi @brandonskiser, no worries. 
I'll close this one for now but I can work on another one if this feature makes sense in the mainline.

Issue #1490, opened 2025-05-02
Feature Request: Make MCPs and their tools configurable by profile


imgaray:
• **Expected Behavior**: Users should be able to configure MCP servers and their associated tools based on different profiles or contexts. This would allow for customized tool sets depending on the user's current work context or project.
• **Actual Behavior**: Currently, MCP servers and their tools appear to be globally configured without the ability to create different profiles with different tool configurations.
• **Steps to Reproduce**:
  1. Add MCP servers to Amazon Q CLI
  2. Attempt to create different profiles with different MCP tool configurations
  3. Notice that there's no built-in way to have different MCP tool configurations for different contexts

ctudor1:
I am also extremely interested in this feature, I think it would be a very powerful functionality for users, and there are a lot of potential use cases.

I've raised a PR #2117 for a potential implementation. Please feel free to give feedback on it. Thanks!

PR #2117, opened 2025-06-26
feat(cli): Support specifying specific mcp configurations per profile (#1490)

ctudor1:
*Issue #, if available:*
#1490 

*Description of changes:*

This PR adds the ability for users to specify which MCP servers they would like to use for any given profile. I believe this would be a very powerful feature for users, as it allows them to pick specific sets of tools for certain tasks, resulting in better context window utilisation and mitigating other issues caused by having too many mcp servers enabled.  

The way this is currently implemented in this PR is like this:
- If a user places a `mcp.json` file into the `/.aws/amazonq/profiles/<profilename>/mcp.json` (or creates one using the `q mcp add` command) , and that profile is active, the mcp servers defined here will be included along with the global and workspace based servers (by default)
- A new field within this `mcp.json` file may exist called `useProfileServersOnly`. When this field exists and is set to `true`, the mcp servers configured at the profile level are used exclusively, and global/workspace servers are not loaded.
- When profile based mcp servers are included alongside the global/workspace ones (default behaviour), the order of precedence goes Profile > Workspace > Global in the event of a conflict.
- When a user switches profile using the `/profile set <profilename>` command, all mcp servers are reloaded to take the profile level preferences into account.
- The q mcp commands allow for showing/manipulating profile level mcp servers, in addition to global and workspace as it is currently. e.g `q mcp list --profile <profilename>`
- There is a new `q mcp` command for setting the `useProfileServersOnly` field in the json. Example: `q mcp use-profile-servers-only --profile gateway`

Apologies for the size of this PR, I tried to keep it brief but it interacts with a fair amount of logic so some refactors were necessary.

Next Steps:
- All feedback would be very welcome on either the design (UX) or implementation for this PR. 
- Nitpicks on the code are also helpful, I'm not too familiar with Rust so if I've missed some best practices please point them out.
- I think unit testing and documentation for this change can be improved further, so I will do that as well as incorporate any feedback.

--- 

### DIFF
--- a/crates/cli/src/cli/chat/cli.rs
+++ b/crates/cli/src/cli/chat/cli.rs
@@ -46,10 +46,9 @@ pub enum Mcp {
     /// Import a server configuration from another file
     Import(McpImport),
     /// Get the status of a configured server
-    Status {
-        #[arg(long)]
-        name: String,
-    },
+    Status(McpStatus),
+    /// Set whether a profile should exclusively use its own MCP servers
+    UseProfileServersOnly(McpUseProfileServersOnly),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Args)]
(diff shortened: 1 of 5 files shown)


ctudor1:
Closing this pull request. From my understanding there is a significant change coming soon that will not likely to be compatible with this PR. It was also made against a previous version of the repo. I have ported it over [to the new one here](https://github.com/ctudor1/amazon-q-developer-cli-autocomplete/commit/35a8e72a764c44f9225b0187ce2ef24ba32489b3) for my own usage. Thanks all.

Issue #1970, opened 2025-05-27
Feature Request: Support MCP in progress messages


NathanRoseCE:
Q does not implement [the progress notification](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress). This ticket is to request the CLI impliment this feature. This will be useful for longer running processes and for providing user observability into what the Q-cli LLM sees and is responding to(which is presently not possible)

Please impliment this spec

Current Q output:
```
?
🛠️  Using tool: test_tool from mcp server nmrose_build_tools
 ⋮ 
 ● Running test_tool with the param:
 ⋮  {
 ⋮    "arguments": {},
 ⋮    "name": "test_tool"
 ⋮  }
Allow this action? Use 't' to trust (always allow) this tool for the session. [y/n/t]:

> y

 ⋮ 
 ● Completed in 8.4s
```

Same invocation but directly via stdio
```
{"jsonrpc": "2.0", "method": "tools/call", "params": {"name": "nmrose_build_tools___test_tool", "arguments": {}}, "id": 1}
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"tool not found"}}
{"jsonrpc": "2.0", "method": "tools/call", "params": {"name": "test_tool", "arguments": {}}, "id": 1}
{"jsonrpc": "2.0", "method": "notifications/progress", "params": {"progressToken": "abc123", "progress": 1, "message": "Starting test tool operation"}}
{"jsonrpc": "2.0", "method": "notifications/progress", "params": {"progressToken": "abc123", "progress": 2, "message": "🔍 Initializing components"}}
{"jsonrpc": "2.0", "method": "notifications/progress", "params": {"progressToken": "abc123", "progress": 3, "message": "⚠️ Some resources might be unavailable"}}
{"jsonrpc": "2.0", "method": "notifications/progress", "params": {"progressToken": "abc123", "progress": 4, "message": "Processing data"}}
{"jsonrpc": "2.0", "method": "notifications/progress", "params": {"progressToken": "abc123", "progress": 5, "message": "✅ Operation completed successfully"}}
{"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"Test tool executed successfully with progress notifications"}],"isError":false}}
```

### Solutioning

Feel free to disregard if you have deeper more ingrained thoughts on this could look something like this w/ no total field(like above)

```
?
🛠️  Using tool: test_tool from mcp server nmrose_build_tools
[...]

PR #2205, opened 2025-07-02
fix(logs): sanitize strings and print the log message sent by mcp server

dabasjayant:
*Issue #1970*

## Overview
Q CLI supports receiving logs from MCP servers, but the current implementation keeps double quotes (example: `\"info\"` from the json strings, so the level doesn't match either of `debug` | `info` | `warning` | `error`.

Therefore, the issue is that while the message is received, it's never logged.

## Key Changes
- Read '**level**' as `string` instead of `json.to_string()`.
- Read '**data**' as either `string` or `json` as both are supported by MCP SDK
- ~Print the log to stdout so any tool with long runtime can show progress or status to the user for a better user experience.~

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/crates/chat-cli/src/mcp_client/client.rs
+++ b/crates/chat-cli/src/mcp_client/client.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::io::stdout;
 use std::process::Stdio;
 use std::sync::atomic::{
     AtomicBool,
@@ -12,10 +11,6 @@ use std::sync::{
 };
 use std::time::Duration;
 
-use crossterm::{
-    queue,
-    style,
-};
 use serde::{
     Deserialize,
     Serialize,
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -379,12 +384,25 @@ where
...
                                             .as_ref()
                                             .and_then(|p| p.get("level"))
-                                            .and_then(|v| serde_json::to_string(v).ok());
-                                        let data = params
-                                            .as_ref()
-                                            .and_then(|p| p.get("data"))
-                                            .and_then(|v| serde_json::to_string(v).ok());
+                                            .and_then(|v| v.as_str())
+                                            .map(|s| s.to_string());
+                                        let data = params.as_ref().and_then(|p| p.get("data")).map(|v| match v {
```
dabasjayant, line 384: MCP SDK specifies that `data` is: "The data to be logged, such as a string message or an object. Any JSON serializable type is allowed here."

Thus, we expect it to be a `string` or `json` object

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -379,12 +384,25 @@ where
...
+                                            serde_json::Value::String(s) => s.clone(),
+                                            _ => serde_json::to_string_pretty(v).unwrap_or_default(),
+                                        });
                                         if let (Some(level), Some(data)) = (level, data) {
+                                            let _ = queue!(
```
dingfeli: This prints to the terminal. I am not sure if this is the intended behavior of the debug logging.
I am not 100% sure but I think the current logging set up already has a layer configured to pipe out to stdout for verbosity > 0 (Please verify it if you can).

dabasjayant: I've tried running with verbosity `-v` to `-vvvv` but couldn't see the log in stdout/stderr or written to a file.
After discussing with my teammates, we believe that being able to see the tool notifications should be an expected behavior to communicate status/progress to the client or user. However, if the dev team prefers to separate it from standard output, my suggestion would be to add a bool argument to enable/disable printing of the logs to enhance the control for user experience.

dabasjayant: Removed log printing to stdout for this fix

dingfeli (approved):

PR #2236, opened 2025-07-05
Implement disableMarkdownRendering setting to allow users to disable markdown …

alex-bellomo:
*Issue #, if available:* 
- None.

*Why make this change:*
- This was personally bothering me as a user - asking Q to generate markdown-formatted text (e.g. CR descriptions) would result in rendered markdown. I would have to re-format the output after pasting it into CRUX.

*Description of changes:*
- Created a new `chat` setting, `disableMarkdownRendering`, that users can use to disable markdown rendering in the responses. The setting defaults to `true`, unless specified `false` by the user.
- When the `disableMarkdownRendering` setting is set to `false`, skip the markdown-related parsers in `parse.rs`.
- Added unit tests to validate behavior.

EDIT: Updated name of setting from `enableMarkdown` to `disableMarkdownRendering`.

### DIFF
--- a/crates/chat-cli/src/database/settings.rs
+++ b/crates/chat-cli/src/database/settings.rs
@@ -33,6 +33,7 @@ pub enum Setting {
     McpNoInteractiveTimeout,
     McpLoadedBefore,
     ChatDefaultModel,
+    ChatDisableMarkdownRendering,
     ChatDefaultAgent,
     ChatDisableAutoCompaction,
     ChatEnableHistoryHints,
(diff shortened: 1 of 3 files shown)


crates/chat-cli/src/database/settings.rs
```diff
@@ -91,6 +93,7 @@ impl TryFrom<&str> for Setting {
             "mcp.noInteractiveTimeout" => Ok(Self::McpNoInteractiveTimeout),
             "mcp.loadedBefore" => Ok(Self::McpLoadedBefore),
             "chat.defaultModel" => Ok(Self::ChatDefaultModel),
+            "chat.enableMarkdown" => Ok(Self::ChatDisableMarkdownRendering),
```
brandonskiser: This wasn't updated to `chat.disableMarkdownRendering`

brandonskiser (changes requested):

crates/chat-cli/src/database/settings.rs
alex-bellomo: Whoops! Fixed below!

brandonskiser (approved):

PR #2239, opened 2025-07-07
feat(mcp): Add MCP sampling support

swapneils:
Implement comprehensive MCP sampling functionality following the MCP specification.

Tested against Anthropic's `everything` server

*Description of changes:*

This PR adds support for [sampling](https://modelcontextprotocol.io/docs/concepts/sampling), which allows MCP servers to make LLM calls using the Agent's own LLM backend.

This is currently supported by [very few clients](https://modelcontextprotocol.io/clients), but has been a feature of the MCP protocol since the initial 2024-11-05 version. 
It is also a _very_ useful feature for MCP _authors_, who often have a use for writing their own subagents, summarizing information without using up chat context, guiding prompts better than the LLM Agent might, prompt-engineering to better control an Agent's response to their output, etc. Sampling can significantly improve the quality of an MCP server, and the server can still fall back to non-sampling approaches if some other client doesn't support it.

UX: To enable sampling, you set `sampling: true` in the config for an MCP server, after which any sampling requests are accepted
- I considered making per-sampling-request validations, but that would require extensive rearchitecting to create a second UI stream, capable of interrupting the main stream during both tool-calls _and_ general execution (I personally have at least one use-case where sampling occurs outside tool calls). This is a significant rearchitecting.
- Also, the current approach is a 2-way door; we can add per-request sampling later for servers without `sampling: true`.

Tested against [Anthropic's `everything` server](https://github.com/modelcontextprotocol/servers/tree/main/src/everything). `sampleLLM` successfully generates responses and seems to also respect token limits.

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/custom_tool.rs
+++ b/crates/chat-cli/src/cli/chat/tools/custom_tool.rs
@@ -127,9 +127,6 @@ impl CustomToolClient {
             trust_sampling,
         } = config;
 
-        // Pass sampling_sender to all servers - sampling permission is now based on tool call state
-        let conditional_sampling_sender = sampling_sender;
-
         // Process environment variables if present
         let processed_env = env.map(|mut env_vars| {
             process_env_vars(&mut env_vars, &os.env);
@@ -146,7 +143,7 @@ impl CustomToolClient {
                "version": "1.0.0"
             }),
             env: processed_env,
-            sampling_sender: conditional_sampling_sender,
+            sampling_sender: sampling_sender,
         };
         let client = McpClient::<JsonRpcStdioTransport>::from_config(mcp_client_config)?;
         Ok(CustomToolClient::Stdio {


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -508,6 +514,7 @@ impl ChatSession {
         model_id: Option<String>,
         tool_config: HashMap<String, ToolSpec>,
         interactive: bool,
+        sampling_receiver: tokio::sync::mpsc::UnboundedReceiver<crate::mcp_client::sampling_ipc::PendingSamplingRequest>,
```
dingfeli: Feel free to just bring these types in with use statements.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -480,6 +484,8 @@ pub struct ChatSession {
     conversation: ConversationState,
     tool_uses: Vec<QueuedTool>,
     pending_tool_index: Option<usize>,
+    /// Channel receiver for incoming sampling requests from MCP servers
+    sampling_receiver: tokio::sync::mpsc::UnboundedReceiver<crate::mcp_client::sampling_ipc::PendingSamplingRequest>,
```
dingfeli: Same here https://github.com/aws/amazon-q-developer-cli/pull/2239/files#r2191074016

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -293,6 +293,9 @@ impl ChatArgs {
             .await?;
         let tool_config = tool_manager.load_tools(os, &mut stderr).await?;
 
+        // Set the ApiClient for MCP clients that have sampling enabled
+        tool_manager.set_streaming_client(std::sync::Arc::new(os.client.clone()));
```
dingfeli: Same here: https://github.com/aws/amazon-q-developer-cli/pull/2239/files#r2191074016

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2360,6 +2379,12 @@ mod tests {
         agents
     }
 
+    #[cfg(test)]
```
dingfeli: This entire module is already configured to only be compiled during test. No need to annotate again here.

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -19,11 +20,65 @@ use thiserror::Error;
...
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SamplingResponse {
+    pub role: String,
+    pub content: SamplingContent,
+    pub model: String,
+    pub stop_reason: String,
+}
+
```
dingfeli: These can go in `crates/chat-cli/src/mcp_client/facilitator_types.rs`

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -67,9 +122,14 @@ struct ClientCapabilities {
 
 impl From<ClientInfo> for ClientCapabilities {
     fn from(client_info: ClientInfo) -> Self {
+        let mut capabilities = HashMap::new();
+        // Declare sampling capability support
+        capabilities.insert("sampling".to_string(), serde_json::json!({}));
```
dingfeli, line 84: If this isn't conditional we can perhaps add this into `Capabilities` default impl. 
I can't quite recall if I ever made use of this struct during the initial exchange though. I'll double check.

crates/chat-cli/src/api_client/mod.rs
```diff
@@ -31,6 +31,7 @@ pub use endpoints::Endpoint;
 pub use error::ApiClientError;
```
dingfeli: We can probably just revert the changes in this file since it looks unintentional.

crates/chat-cli/src/cli/chat/mod.rs
dingfeli: Also looking at the definition of the client ins `Os`, we have the following: 
```rust
#[derive(Clone, Debug)]
pub struct ApiClient {
    client: CodewhispererClient,
    streaming_client: Option<CodewhispererStreamingClient>,
    sigv4_streaming_client: Option<QDeveloperStreamingClient>,
    mock_client: Option<Arc<Mutex<std::vec::IntoIter<Vec<ChatResponseStream>>>>>,
    profile: Option<AuthProfile>,
}
```

`CodewhispererClient` is defined as follows: 
```rust
/// Client for Amazon CodeWhisperer
///
/// Client for invoking operations on Amazon CodeWhisperer. Each operation on Amazon CodeWhisperer
/// is a method on this this struct. `.send()` MUST be invoked on the generated operations to
/// dispatch the request to the service.
#[derive(::std::clone::Clone, ::std::fmt::Debug)]
pub struct Client {
    handle: ::std::sync::Arc<Handle>,
}
```

`CodewhispererStreamingClient` is defined as follows:
```rust
/// Client for Amazon CodeWhisperer Streaming
///
/// Client for invoking operations on Amazon CodeWhisperer Streaming. Each operation on Amazon
/// CodeWhisperer Streaming is a method on this this struct. `.send()` MUST be invoked on the
/// generated operations to dispatch the request to the service.
#[derive(::std::clone::Clone, ::std::fmt::Debug)]
pub struct Client {
    handle: ::std::sync::Arc<Handle>,
}
```

`QDeveloperStreamingClient`:
```rust
/// Client for Amazon Q Developer Streaming
///
/// Client for invoking operations on Amazon Q Developer Streaming. Each operation on Amazon Q
/// Developer Streaming is a method on this this struct. `.send()` MUST be invoked on the generated
/// operations to dispatch the request to the service.
#[derive(::std::clone::Clone, ::std::fmt::Debug)]
pub struct Client {
    handle: ::std::sync::Arc<Handle>,
}
```

All of which are already wrapped in `Arc`s. Unless there is some other reasons that I am not seeing I am not entirely convinced that we would need to introduce another indirection here.

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -131,6 +193,8 @@ pub struct Client<T: Transport> {
     // TODO: move this to tool manager that way all the assets are treated equally
     pub prompt_gets: Arc<SyncRwLock<HashMap<String, PromptGet>>>,
     pub is_prompts_out_of_date: Arc<AtomicBool>,
+    sampling_sender: Option<tokio::sync::mpsc::UnboundedSender<crate::mcp_client::sampling_ipc::PendingSamplingRequest>>,
+    streaming_client: Arc<Mutex<Option<std::sync::Arc<crate::api_client::ApiClient>>>>,
```
dingfeli: This is a static shared reference of a lock of a nullable static shared reference of a client, whose inners are themselves static shared references.

Can you help me understand why we are putting the `Arc<ApiClient>` behind a mutex? It's guarding an `Arc<T>`, which in terms of mutability is just analogous to &T. That is already a shared reference. 

Also, if this is nullable as we might not actually receive a streaming in some cases OR there exists a period of time during the initialization of the application where we won't have enough information to to assign the `Client<T>` with one, `Option<Arc<T>>` makes more sense here (as opposed to `Arc<Option<T>>`), not that I think we actually need an `Arc` here.

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -371,7 +440,70 @@ where
...
+                                        let client_ref = client_ref.clone();
+                                        let transport_ref = transport_ref.clone();
+                                        let req_clone = req.clone();
+                                        
+                                        tokio::spawn(async move {
+                                            tracing::debug!(target: "mcp", "Processing sampling request in spawned task");
+                                            match client_ref.handle_sampling_request(req_clone.clone()).await {
+                                                Ok(response) => {
+                                                    let msg = JsonRpcMessage::Response(response);
+                                                    if let Err(e) = transport_ref.send(&msg).await {
```
dingfeli, line 420: Admittedly I haven't looked into the use case for sampling a whole lot. I think you mentioned that you are writing a server that will leverage this in some way. Can you tell me a bit about how it leverages what is being sent here? The disconnect I have here is that the content in this jrpc response are unstructured responses that takes whatever form the LLM produces.

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -587,6 +719,211 @@ where
...
+    pub async fn handle_sampling_request(&self, request: JsonRpcRequest) -> Result<JsonRpcResponse, ClientError> {
+        tracing::info!(target: "mcp", "Received sampling request from server: {}", self.server_name);
+
+        // Parse the sampling request
+        let sampling_request: SamplingRequest = serde_json::from_value(
+            request.params.unwrap_or_default()
+        ).map_err(|e| ClientError::NegotiationError(format!("Invalid sampling request: {}", e)))?;
+
+        // Check if sampling is enabled for this server
+        if self.sampling_sender.is_some() {
```
dingfeli: I am struggling to see how this is used beyond just being checked for its existence. Can you please help me understand if I had missed it anywhere.

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -587,6 +719,211 @@ where
...
+    /// Handle sampling/createMessage requests from MCP servers
+    pub async fn handle_sampling_request(&self, request: JsonRpcRequest) -> Result<JsonRpcResponse, ClientError> {
+        tracing::info!(target: "mcp", "Received sampling request from server: {}", self.server_name);
+
+        // Parse the sampling request
+        let sampling_request: SamplingRequest = serde_json::from_value(
+            request.params.unwrap_or_default()
+        ).map_err(|e| ClientError::NegotiationError(format!("Invalid sampling request: {}", e)))?;
+
+        // Check if sampling is enabled for this server
```
dingfeli: Should we perhaps perform all the sanity check at the call sites of `handle_sampling_request`? Not entirely sure what other logic you have in mind but if this is the only flow we could do ourselves a favor here and not have to go through the trouble of cloning and deserializing only to find out we don't have what it takes to sample in the first place.

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -587,6 +719,211 @@ where
...
+                        }
+                    }
+                }
+
+                let model_name = request.model_preferences
+                    .as_ref()
+                    .and_then(|prefs| prefs.hints.as_ref())
+                    .and_then(|hints| hints.first())
+                    .map(|hint| hint.name.clone())
+                    .unwrap_or_else(|| "default-model".to_string());
```
dingfeli: I think we have already obtained this info when we constructed user_message.

dingfeli (changes requested):

crates/chat-cli/src/api_client/mod.rs
swapneils: Yes, this is accidental from the rebase, will remove.

crates/chat-cli/src/mcp_client/client.rs
swapneils: I'm not too familiar with Rust's compiler restrictions (hence "Assisted by Q Developer" :) ), but the overall dataflow here is that `streaming_client` is initially empty, and then we update the Mutex to contain the global client from the general `chat_cli` environment.

As I understood, since the Mutex is being passed through different classes and called potentially-in-parallel (to get its contents or see that it's empty), to avoid race conditions it needs to be inside an `Arc` regardless of whether it's been populated with an atomic `api_client` object. Then the internal `Arc` is just to reflect the type of the `api_client` object passed into the Option.

Is this understanding incorrect?

crates/chat-cli/src/mcp_client/client.rs
swapneils: The purpose of sampling is to let the MCP server call an LLM, without needing to have its own internal LLM setup/auth/payments/etc for the user to manage. These can be used either to generate some portion of a final tool output sent to the Agent (basically dynamic summarization/prompt-engineering) or to chain them together in structured data-processing pipelines for more reliable behavior than the general user-side Agent could manage.

For instance, one of my use-cases is an MCP debugging tool for our team's service. There are cases where we _know_ data needs to be processed with an LLM, but it either needs to be prompted in some specific way (e.g. specific reasoning patterns for analyzing debug output) or just takes up a lot of context (e.g. anomaly detection in log/metric data).

For these cases, prompting the LLM Agent to do this itself is:
A) Fragile, and prone to hallucination, deviance, or negative impact on the user's prompt engineering (via strong optimization pressure from our own prompts)
B) Potentially expensive in token-count
C) Difficult to prompt-engineer the outputs for continued investigation, since heuristic summarization rarely generalizes and returning large raw datasets causes token inefficiency and hallucinations.

On the other hand, if we can query the LLM ourselves outside the LLM agent's context and history, we get significantly more control over how it's applied to analysis problems, and can use the LLM to summarize and prompt-engineer our output, providing more reliable results with better context window usage.

(The other use-case I mentioned isn't fleshed out enough to describe in detail, but from exploratory prototypes it definitely needs asynchronous sampling, again to control/parallelize LLM calls for analyzing/summarizing domain-specific non-LLM data better than the agent would)

crates/chat-cli/src/mcp_client/client.rs
swapneils: When initializing the MCP servers, we need to know whether we'll tell them that the client supports sampling (since I couldn't figure out a workable UX for approving individual sampling requests outside the MCP config); so we at least need sanity checks before and at MCP initialization.

Not sure what other sanity checks you're proposing to move into `handle_sampling_request`?

crates/chat-cli/src/mcp_client/client.rs
swapneils: Ah, good point, we can just get the name from there.

crates/chat-cli/src/mcp_client/client.rs
swapneils: This is a good point. I think we're just using `transport_ref` to send the data now, so probably this can be switched out with a boolean.

crates/chat-cli/src/mcp_client/client.rs
dingfeli: I don't think there is a race condition here. The `send_message` method takes a shared reference:
https://github.com/aws/amazon-q-developer-cli/blob/00286fbe0688fe19fef9a5149bb9f9172f96783c/crates/chat-cli/src/api_client/mod.rs#L251

There is no need to lock.

crates/chat-cli/src/cli/chat/tool_manager.rs
```diff
@@ -850,12 +863,24 @@ impl Clone for ToolManager {
             is_interactive: self.is_interactive,
             mcp_load_record: self.mcp_load_record.clone(),
             disabled_servers: self.disabled_servers.clone(),
+            sampling_request_sender: self.sampling_request_sender.clone(),
```
dingfeli: I don't see how this field is used. Why does `ToolManager` need to own an instance of this mpsc sender?

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1267,6 +1275,17 @@ impl ChatSession {
...
+        // Check for incoming sampling requests and automatically approve them
+        // Since servers now opt-in via configuration, any request that comes through should be approved
+        while let Ok(mut sampling_request) = self.sampling_receiver.try_recv() {
+            tracing::info!(target: "mcp", "Auto-approving sampling request from configured server: {}", sampling_request.server_name);
+
+            // Automatically approve the sampling request
+            sampling_request.send_approval_result(
+                crate::mcp_client::sampling_ipc::SamplingApprovalResult::approved()
+            );
+        }
```
dingfeli: Not entirely sure what this is doing. Some questions:
- This is sending back an approval every time? 
- What is receiving what `send_approval_result` is sending?
- This is being called in the main chat loop under `prompt_user`, which means this will only get run when user is to be prompted (i.e. after Q CLI has responded). Correct me if I am wrong here, but I think sampling would occur during a tool call (i.e. before the user is to have their turn again). To me this is not the right point in the logic flow to check and respond for a sampling request. Let me know if I had misunderstood anything.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -284,7 +284,7 @@ impl ChatArgs {
         info!(?conversation_id, "Generated new conversation id");
         let (prompt_request_sender, prompt_request_receiver) = std::sync::mpsc::channel::<Option<String>>();
         let (prompt_response_sender, prompt_response_receiver) = std::sync::mpsc::channel::<Vec<String>>();
-        let mut tool_manager = ToolManagerBuilder::default()
+        let (mut tool_manager, sampling_receiver) = ToolManagerBuilder::default()
```
dingfeli: I had left my question below on how `sampling_receiver` is used. That question aside, I think `sampling_receiver` can just be owned by `ToolManager`.

crates/chat-cli/src/cli/chat/mod.rs
swapneils: > I think sampling would occur during a tool call (i.e. before the user is to have their turn again).

The docs don't mention anything either way (though the dataflow diagram doesn't mention an ongoing tool-call when the server initiates sampling), but to my understanding this is incorrect. 

MCP servers support long-running operations (see for instance [FastMCP's doc on progress reporting, which is meant to support such operations](https://gofastmcp.com/servers/progress)); it seems strange to support sampling for tool-calls but not such long-running operations.

There is also [at least one real customer use-case](https://github.com/orgs/modelcontextprotocol/discussions/368#discussion-8340777) on the MCP discussions where it seems the server proactively starts sampling requests even though the client isn't explicitly expecting them.

> What is receiving what send_approval_result is sending?

This was missed in a refactoring, [as mentioned in this other thread](https://github.com/aws/amazon-q-developer-cli/pull/2239#discussion_r2191215897) transport_ref is handling sending data to the server now so we can probably get rid of the sampling_result call. Will test that this works when refactoring.

> This is sending back an approval every time?

We're filtering whether a server has access to sampling by whether we provide a sampling callback at all, since right now it's a binary trust/don't-trust situation. If a server is able to make a sampling request and reach this section, it should be allowed.

Supporting manual approve/deny is difficult from a CLI interface given we only have one input-stream (and we can't rely on things like popups working without breaking customer orchestrations, either). 

Adding approvals _only_ during tool calls and then rejecting untrusted sampling requests outside that context is maybe doable, but I don't think it should block an initial implementation of the feature.

EDIT: The code-as-written of [the python SDK](https://github.com/modelcontextprotocol/python-sdk) also seems to support sampling at arbitrary times with a similar callback-based approach as the one we're using here (assuming I'm reading it correctly).

crates/chat-cli/src/mcp_client/client.rs
swapneils: This should be conditional, I'm hiding sampling capability from servers which aren't trusted for it.

Maybe in the future we could change that by adding a validation UI, but my initial attempts at that struggled to get a good UX, so it needs more thought and perhaps an RFC for both Sampling and Elicitation UX.

crates/chat-cli/src/cli/chat/mod.rs
swapneils: Discussed offline, while the spec is unclear here limiting to tool calls has the benefit that we can manually approve/deny requests with the existing UI, so we'll go with that approach for the current implementation.

swapneils: > What is receiving what send_approval_result is sending?

Restructured the code again while refactoring, and got the time to dig deeper into its dataflows w/ Q. 

In the current version of the code, the `mcp_client` receives the result of `send_approval_result`. Line 760 in mcp_client/client.rs sets up the sender/receiver channels, the former of which is added in the PendingSamplingRequest so we can tell the mcp_client whether the request was approved or denied.

crates/chat-cli/src/cli/chat/tool_manager.rs
swapneils: `sampling_request_sender` was a refactoring mistake, it has been removed.

crates/chat-cli/src/mcp_client/client.rs
swapneils: Refactored to a degree, but refactoring fully into a single instance makes the "default-model" value cause issues when saved into the `user_message`.

crates/chat-cli/src/mcp_client/client.rs
swapneils: Apologies, did not fully understand the code flow last time due to lack of review bandwidth; `sampling_receiver` and `sampling_sender` are used to ferry sampling requests from the mcp_client to the `invoke` method so that they can be approved/denied.

Currently this will always be present, since we are exposing the sampling functionality to all servers, but in theory we may want something like the initial revision where some servers aren't allowed to see the sampling feature.

crates/chat-cli/src/cli/chat/mod.rs
swapneils: Moved into ToolManager

crates/chat-cli/src/mcp_client/client.rs
swapneils: In Rev2 sampling is exposed to all servers, but I can see reasons to make it so in the future (i.e. if servers have different behaviors when sampling is/isn't present, users may want to intentionally not expose the capability). Inclined to still have this processed at creation time rather than in the `default`.

crates/chat-cli/src/mcp_client/client.rs
swapneils: Looked into this further, we write to this mutex via `set_streaming_client`, so it's not just in the client. I don't think it's possible to remove all locking from this variable, though please correct me if I'm misunderstanding how locks/assignment work in Rust.

swapneils:
Rebased again. Local unit tests are still succeeding (`cargo test`). Local manual tests of the new functionality also succeed:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 28.54s
     Running `target/debug/chat_cli chat --agent trusted-everything`
⠹ 0 of 1 mcp servers initialized. ctrl-c to start chatting now

    ⢠⣶⣶⣦⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣤⣶⣿⣿⣿⣶⣦⡀⠀
 ⠀⠀⠀⣾⡿⢻⣿⡆⠀⠀⠀⢀⣄⡄⢀⣠⣤⣤⡀⢀⣠⣤⣤⡀⠀⠀⢀⣠⣤⣤⣤⣄⠀⠀⢀⣤⣤⣤⣤⣤⣤⡀⠀⠀⣀⣤⣤⣤⣀⠀⠀⠀⢠⣤⡀⣀⣤⣤⣄⡀⠀⠀⠀⠀⠀⠀⢠⣿⣿⠋⠀⠀⠀⠙⣿⣿⡆
 ⠀⠀⣼⣿⠇⠀⣿⣿⡄⠀⠀⢸⣿⣿⠛⠉⠻⣿⣿⠛⠉⠛⣿⣿⠀⠀⠘⠛⠉⠉⠻⣿⣧⠀⠈⠛⠛⠛⣻⣿⡿⠀⢀⣾⣿⠛⠉⠻⣿⣷⡀⠀⢸⣿⡟⠛⠉⢻⣿⣷⠀⠀⠀⠀⠀⠀⣼⣿⡏⠀⠀⠀⠀⠀⢸⣿⣿
 ⠀⢰⣿⣿⣤⣤⣼⣿⣷⠀⠀⢸⣿⣿⠀⠀⠀⣿⣿⠀⠀⠀⣿⣿⠀⠀⢀⣴⣶⣶⣶⣿⣿⠀⠀⠀⣠⣾⡿⠋⠀⠀⢸⣿⣿⠀⠀⠀⣿⣿⡇⠀⢸⣿⡇⠀⠀⢸⣿⣿⠀⠀⠀⠀⠀⠀⢹⣿⣇⠀⠀⠀⠀⠀⢸⣿⡿
 ⢀⣿⣿⠋⠉⠉⠉⢻⣿⣇⠀⢸⣿⣿⠀⠀⠀⣿⣿⠀⠀⠀⣿⣿⠀⠀⣿⣿⡀⠀⣠⣿⣿⠀⢀⣴⣿⣋⣀⣀⣀⡀⠘⣿⣿⣄⣀⣠⣿⣿⠃⠀⢸⣿⡇⠀⠀⢸⣿⣿⠀⠀⠀⠀⠀⠀⠈⢿⣿⣦⣀⣀⣀⣴⣿⡿⠃
 ⠚⠛⠋⠀⠀⠀⠀⠘⠛⠛⠀⠘⠛⠛⠀⠀⠀⠛⠛⠀⠀⠀⠛⠛⠀⠀⠙⠻⠿⠟⠋⠛⠛⠀⠘⠛⠛⠛⠛⠛⠛⠃⠀⠈⠛⠿⠿⠿⠛⠁⠀⠀⠘⠛⠃⠀⠀⠘⠛⠛⠀⠀⠀⠀⠀⠀⠀⠀⠙⠛⠿⢿⣿⣿⣋⠀⠀
 ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠛⠿⢿⡧

╭─────────────────────────────── Did you know? ────────────────────────────────╮
│                                                                              │
│      You can use /editor to edit your prompt with a vim-like experience      │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯

/help all commands  •  ctrl + j new lines  •  ctrl + s fuzzy search
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
🤖 You are chatting with claude-sonnet-4

✓ everything loaded in 1.06 s

> I'll use the sampleLLM tool to ask about the weather.


🛠️  Using tool: sampleLLM (trusted) from mcp server everything
 ⋮ 
 ● Running sampleLLM with the param:
 ⋮  {
 ⋮    "name": "sampleLLM",
 ⋮    "arguments": {
 ⋮      "prompt": "What's the weather like today?"
 ⋮    }
 ⋮  }

 ⋮ 
 ● Completed in 2.353s


> The LLM responded that it doesn't have access to current weather information and suggested several 
alternatives for checking the weather, including weather websites, mobile apps, voice assistants, and weatherAPIs for application integration.


(To exit the CLI, press Ctrl+C or Ctrl+D again or type /quit)

   Compiling chat_cli v1.14.1 (path)
warning: struct `NullMessenger` is never constructed
  --> crates/chat-cli/src/mcp_client/messenger.rs:55:12
   |
55 | pub struct NullMessenger;
   |            ^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: struct `UnknownDesktopErrContext` is never constructed
  --> crates/chat-cli/src/util/mod.rs:43:12
   |
43 | pub struct UnknownDesktopErrContext {
   |            ^^^^^^^^^^^^^^^^^^^^^^^^

warning: `chat_cli` (bin "chat_cli") generated 2 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.29s
     Running `target/debug/chat_cli chat --agent everything 'Please run sampleLLM to ask about the weather'`
✓ everything loaded in 1.06 s


    ⢠⣶⣶⣦⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣤⣶⣿⣿⣿⣶⣦⡀⠀
 ⠀⠀⠀⣾⡿⢻⣿⡆⠀⠀⠀⢀⣄⡄⢀⣠⣤⣤⡀⢀⣠⣤⣤⡀⠀⠀⢀⣠⣤⣤⣤⣄⠀⠀⢀⣤⣤⣤⣤⣤⣤⡀⠀⠀⣀⣤⣤⣤⣀⠀⠀⠀⢠⣤⡀⣀⣤⣤⣄⡀⠀⠀⠀⠀⠀⠀⢠⣿⣿⠋⠀⠀⠀⠙⣿⣿⡆
 ⠀⠀⣼⣿⠇⠀⣿⣿⡄⠀⠀⢸⣿⣿⠛⠉⠻⣿⣿⠛⠉⠛⣿⣿⠀⠀⠘⠛⠉⠉⠻⣿⣧⠀⠈⠛⠛⠛⣻⣿⡿⠀⢀⣾⣿⠛⠉⠻⣿⣷⡀⠀⢸⣿⡟⠛⠉⢻⣿⣷⠀⠀⠀⠀⠀⠀⣼⣿⡏⠀⠀⠀⠀⠀⢸⣿⣿
 ⠀⢰⣿⣿⣤⣤⣼⣿⣷⠀⠀⢸⣿⣿⠀⠀⠀⣿⣿⠀⠀⠀⣿⣿⠀⠀⢀⣴⣶⣶⣶⣿⣿⠀⠀⠀⣠⣾⡿⠋⠀⠀⢸⣿⣿⠀⠀⠀⣿⣿⡇⠀⢸⣿⡇⠀⠀⢸⣿⣿⠀⠀⠀⠀⠀⠀⢹⣿⣇⠀⠀⠀⠀⠀⢸⣿⡿
 ⢀⣿⣿⠋⠉⠉⠉⢻⣿⣇⠀⢸⣿⣿⠀⠀⠀⣿⣿⠀⠀⠀⣿⣿⠀⠀⣿⣿⡀⠀⣠⣿⣿⠀⢀⣴⣿⣋⣀⣀⣀⡀⠘⣿⣿⣄⣀⣠⣿⣿⠃⠀⢸⣿⡇⠀⠀⢸⣿⣿⠀⠀⠀⠀⠀⠀⠈⢿⣿⣦⣀⣀⣀⣴⣿⡿⠃
 ⠚⠛⠋⠀⠀⠀⠀⠘⠛⠛⠀⠘⠛⠛⠀⠀⠀⠛⠛⠀⠀⠀⠛⠛⠀⠀⠙⠻⠿⠟⠋⠛⠛⠀⠘⠛⠛⠛⠛⠛⠛⠃⠀⠈⠛⠿⠿⠿⠛⠁⠀⠀⠘⠛⠃⠀⠀⠘⠛⠛⠀⠀⠀⠀⠀⠀⠀⠀⠙⠛⠿⢿⣿⣿⣋⠀⠀
 ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠛⠿⢿⡧

╭─────────────────────────────── Did you know? ────────────────────────────────╮
│                                                                              │
│   Q can use tools without asking for confirmation every time. Give /tools    │
│                                 trust a try                                  │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯

/help all commands  •  ctrl + j new lines  •  ctrl + s fuzzy search
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
🤖 You are chatting with claude-sonnet-4


> I'll use the sampleLLM tool to ask about the weather.


🛠️  Using tool: sampleLLM (trusted) from mcp server everything
 ⋮ 
 ● Running sampleLLM with the param:
 ⋮  {
 ⋮    "name": "sampleLLM",
 ⋮    "arguments": {
 ⋮      "prompt": "What's the weather like today?"
 ⋮    }
 ⋮  }

🤖 MCP Sampling Request from 'everything'
📝 Prompt: Resource sampleLLM context: What's the weather like today?
🎯 System: You are a helpful test server.
📊 Max tokens: 100

Allow this AI sampling request? [y/n]: 
n

 ⋮ 
 ● Execution failed after 35.136s:
{"code":-32601,"message":"MCP error -32601: User denied the sampling request"}


> The sampling request was denied by the user. The sampleLLM tool requires user permission to make LLM sampling
calls, and it appears that permission was not granted for this request.

To use the sampleLLM tool, you would need to approve the sampling request when prompted by your MCP client.

[everything] > Please try again, I'll approve this time.




🛠️  Using tool: sampleLLM (trusted) from mcp server everything
 ⋮ 
 ● Running sampleLLM with the param:
 ⋮  {
 ⋮    "name": "sampleLLM",
 ⋮    "arguments": {
 ⋮      "prompt": "What's the weather like today?"
 ⋮    }
 ⋮  }

🤖 MCP Sampling Request from 'everything'
📝 Prompt: Resource sampleLLM context: What's the weather like today?
🎯 System: You are a helpful test server.
📊 Max tokens: 100

Allow this AI sampling request? [y/n]: 
y

 ⋮ 
 ● Completed in 4.715s


> The LLM responded that it doesn't have access to current weather information. It suggested checking weather 
websites, apps, or using weather APIs for real-time weather data.

If you need actual weather information, I could help you make an API call to a weather service if you have 
access to one, or we could look into command-line weather tools available on your system.

[everything] > /quit
```

dingfeli:
Thank you for your effort in this. We have since moved away from the custom mcp client implementation and adopted the use of anthropic's official rust crate, which unfortunately renders this PR no longer applicable. 

We are also in the process of rewriting our core event loop. I don't quite have a precise ETA but it should be sometime before the end of the year. Once that is done, we are ready to revisit this topic again. 

Thanks again.

Issue #2101, opened 2025-06-23
Q CLI Profile-Specific MCP Server Not Loading Environment Variables


kagarmoe:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue

### Operating system

macOS 15.5.0 (24F74)

### Expected behaviour

I don't need all of the servers running all of the time. I would like to use profile-specific servers, particularly for the figma server. The figma server works correctly, which indicates that this is likely a q cli issue.

When using `q chat --profile figma`, the MCP server defined in `~/.aws/amazonq/profiles/figma/mcp.json` should start successfully with the environment variables specified in the configuration file.

### Actual behaviour

<This will be visible to anyone. Do not include personal or sensitive information>

The Figma MCP server fails to start because it's not receiving the `FIGMA_API_KEY` environment variable from the profile-specific mcp.json configuration. The server exits with error: "[ERROR] Figma API key is required (set environment variable FIGMA_API_KEY)" even though the environment variable is properly defined in the profile's mcp.json file.

```
[chat-transcript]
I found the issue! The Figma

> y

The timeout indicates the server started successfully (it would have exited immediately with an error if the API key was still missing). Let me also check if there are any Q CLI logs that might give us more insight:
[Tool uses: execute_bash]

> t


[Tool uses: fs_read]

The logs don't show specific MCP server startup failures. Let me try running Q chat with the figma profile to see what happens:
[Tool uses: execute_bash]

Let me try without the debug flag:
[Tool uses: execute_bash]

Let me try a different approach to test the MCP server loading:
[Tool uses: execute_bash]

Based on my investigation, here's what I found and the likely issues:

## Root Cause Analysis

The main issue is that **the Figma MCP server is failing to start** because it's not receiving the `FIGMA_API_KEY` environment variable properly. Here's what's happening:

1. **Environment Variable Issue**: When I tested the server manually without the environment variable, it failed with: `[ERROR] Figma API key is required (set environment variable FIGMA_API_KEY)`

2. **MCP Configuration**: Your `mcp.json` file looks correct and includes the environment variable:
   \```json
   "env": {
     "FIGMA_API_KEY": "TOKEN"
   }
   \```

## Potential Solutions

### 1. **Verify API Key Validity**
The API key might be expired or invalid. Test it manually:

\```bash
[...]

PR #2241, opened 2025-07-07
feat: support env vars in the mcp config file

Konippi:
*Issue #, if available:*
#2101 

*Description of changes:*
With this PR, when `${env:VAR_NAME}` is set in `mcp.json`, it can be replaced with an environment variable.

```json
{
    "mcpServers": {
        "tavily-mcp": {
            "command": "npx",
            "args": ["-y", "tavily-mcp@0.2.4"],
            "env": {
                "TAVILY_API_KEY": "${env:TAVILY_API_KEY}"
            },
            "disabled": false,
            "autoApprove": []
        }
    }
}
```

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/custom_tool.rs
+++ b/crates/chat-cli/src/cli/chat/tools/custom_tool.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::env;
 use std::io::Write;
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
@@ -8,6 +9,7 @@ use crossterm::{
     style,
 };
 use eyre::Result;
+use regex::Regex;
 use serde::{
     Deserialize,
     Serialize,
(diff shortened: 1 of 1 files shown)


konippi:
@brandonskiser 
Please review when you have a moment.

liy1lr:
we need this!

crates/chat-cli/src/cli/chat/tools/custom_tool.rs
```diff
@@ -279,3 +308,53 @@ impl CustomTool {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_substitute_env_vars() {
+        // Set a test environment variable
+        unsafe { env::set_var("TEST_VAR", "test_value") };
```
brandonskiser: This looks like it can cause flaky tests - env vars are global to the process, and tests are ran in multiple threads.

Can we instead use the env vars as set by the `Os` type?

konippi: thanks! I'll fix it right away.

brandonskiser:
Thank you for the contribution!

PR #2244, opened 2025-07-07
move commit from q-developer-cli-autocomplete to q-developer-cli + ad…

syaamkhan:
Continuation of PR: https://github.com/aws/amazon-q-developer-cli-autocomplete/pull/332

Description of changes:
Added support for retrieving current user prompts through addition of a USER_PROMPT environment variable. The variable is invoked and created whenever a prompt is received and cleared upon subsequent invocations. In order to consider the case for long prompts, the env variable is limited to the first 4096 characters within a user's prompt. The prompt is also cleared upon the closing of a Q CLI session.

### DIFF
--- a/crates/chat-cli/src/cli/chat/conversation.rs
+++ b/crates/chat-cli/src/cli/chat/conversation.rs
@@ -111,6 +111,17 @@ pub struct ConversationState {
 }
 
 impl ConversationState {
+    fn sanitize_env_value(input: &str) -> String {
+        // Limit the size of input to first 4096 characters
+        let truncated = if input.len() > 4096 {
+            &input[0..4096]
+        } else {
+            input
+        };
+        
+        // Remove any potentially problematic characters
+        return truncated.replace(|c: char| c.is_control() && c != '\n' && c != '\r' && c != '\t' && c != '$', "")
+    }
     pub async fn new(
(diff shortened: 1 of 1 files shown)


evanliu048:
Hey @syaamkhan, thanks for moving your changes to this repo! I pulled your PR and tried to test this functionality locally. Could you please guide me on how to verify that the `USER_PROMPT` environment variable is being set after running `cargo run chat`? I tried running `env | sort` but didn't find the variable. Any guidance on testing this feature would be greatly appreciated!

syaamkhan:
I verified that it is being set by creating a test hook using this: "/hooks add --trigger per_prompt --command "echo $USER_PROMPT" test" within Q. I also added a print statement to print out the output after the hook invocation. This returned something in the form of 'test': 'This is my prompt'.

brandonskiser:
Setting environment variables in a multi-threaded program isn't safe - instead can we just launch the hook commands with the USER_PROMPT env var instead?

brandonskiser (changes requested):
Setting environment variables in a multi-threaded program isn't safe - instead can we just launch the hook commands with the USER_PROMPT env var instead?

brandonskiser (approved):

PR #2262, opened 2025-07-09
fix: revise model unavailable msg in both interactive and non-interactive mode

evanliu048:
*Issue #, if available:*

*Description of changes:*
Updated the error message for `ModelOverloadedError` to provide appropriate guidance based on whether the CLI is running in interactive or non-interactive mode and not show the backtrace information to users. 
### Before:
Screenshot 2025-07-09 at 10 28 11 AM

### After:
Interactive mode:
Screenshot 2025-07-09 at 10 25 20 AM

Non-interactive mode:
Screenshot 2025-07-09 at 10 25 44 AM

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -739,15 +739,31 @@ impl ChatSession {
                 },
                 ApiClientError::QuotaBreach { message, .. } => (message, Report::from(err), true),
                 ApiClientError::ModelOverloadedError { request_id, .. } => {
+                    let model_instruction = if self.interactive {
+                        "Please use '/model' to select a different model and try again."
+                    } else {
+                        "Please relaunch with '--model <model_id>' to use a different model."
+                    };
+
                     let err = format!(
-                        "The model you've selected is temporarily unavailable. Please use '/model' to select a different model and try again.{}\n\n",
+                        "The model you've selected is temporarily unavailable. {}{}\n\n",
+                        model_instruction,
                         match request_id {
                             Some(id) => format!("\n    Request ID: {}", id),
                             None => "".to_owned(),
                         }
                     );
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -739,15 +739,31 @@ impl ChatSession {
...
-                    ("Amazon Q is having trouble responding right now", eyre!(err), true)
+                    execute!(
+                        self.stderr,
+                        style::SetAttribute(Attribute::Bold),
+                        style::SetForegroundColor(Color::Red),
+                        style::Print("Amazon Q is having trouble responding right now:\n"),
+                        style::Print(format!("    {}\n", err.clone())),
+                        style::SetAttribute(Attribute::Reset),
+                        style::SetForegroundColor(Color::Reset),
+                    )?;
```
dingfeli, line 808: I think I might have been too hasty on stamping https://github.com/aws/amazon-q-developer-cli/pull/2263.

I see that we return what we gather from this match arm to a formatted variable: 
```rust 
let (context, report, display_err_message) = match err {
    ...
};
```

And what we print to the output prior to the actual err is roughly the same (i.e. "Amazon Q is having trouble responding right now:\n"). Do we just want to do that outside of the match?

evanliu048: Good point! For the new `QuotaBreach` and `ModelOverloadedError` arms I deliberately print the
err msg inside the match and set `display_err_message = false`, so the outer
block won’t print it a second time – that’s how I avoid duplicating the message
and keep the backtrace out of the user-facing output.
https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/mod.rs#L863-L883

dingfeli (approved):

PR #2281, opened 2025-07-10
fix(agent): agent fallback behavior

dingfeli:
*Issue #, if available:*

*Description of changes:*
Fixes scenario where invalid agents are chosen, fallback does not correctly show active agent. 
Agents are assumed in the following order: 
        1. The agent name specified by the start command via --agent (this is the agent_name that's
           passed in)
        2. If the above is missing or invalid, assume one that is specified by chat.defaultAgent
        3. If the above is missing or invalid, assume the in-memory default

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -223,7 +223,10 @@ impl ChatArgs {
             let mut agents = Agents::load(os, self.agent.as_deref(), skip_migration, &mut stderr).await;
             agents.trust_all_tools = self.trust_all_tools;
 
-            if agents.get_active().is_none_or(|a| a.mcp_servers.mcp_servers.is_empty()) {
+            if agents
+                .get_active()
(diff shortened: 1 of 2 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -219,39 +219,20 @@ impl ChatArgs {
...
-                                )
-                            )?;
-                        }
-                        os.database.settings.set(Setting::McpLoadedBefore, true).await?;
-                    },
-                    Err(e) => {
-                        let _ = execute!(stderr, style::Print(format!("Error switching profile: {}", e)));
-                    },
-                    _ => {},
+            if agents.get_active().is_none_or(|a| a.mcp_servers.mcp_servers.is_empty()) {
```
brandonskiser: We only want to enter this branch if there is an active agent with non-empty mcp servers, right? This looks like it's true only if there's either no agent or there's no mcp servers

dingfeli: Oh you're right. Good catch.

PR #2282, opened 2025-07-10
feat: add agent documentation

chaynabors:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,19 @@
-#### Amazon Q CLI brings IDE-style autocomplete and agentic capabilities to your terminal.
+# Amazon Q CLI
 
-[![Rust Test](https://github.com/aws/amazon-q-developer-cli/actions/workflows/rust.yml/badge.svg)](https://github.com/aws/amazon-q-developer-cli/actions/workflows/rust.yml)
-[![Typos Test](https://github.com/aws/amazon-q-developer-cli/actions/workflows/typos.yml/badge.svg)](https://github.com/aws/amazon-q-developer-cli/actions/workflows/typos.yml)
-[![Typescript Test](https://github.com/aws/amazon-q-developer-cli/actions/workflows/typescript.yml/badge.svg)](https://github.com/aws/amazon-q-developer-cli/actions/workflows/typescript.yml)
-
-## ⚡️ Installation
+## Installation
 
 - **macOS**:
   - **DMG**: [Download now](https://desktop-release.q.us-east-1.amazonaws.com/latest/Amazon%20Q.dmg)
-  - **Homebrew**: `brew install amazon-q`
 - **Linux**:
   - [Ubuntu/Debian](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html#command-line-installing-ubuntu)
   - [AppImage](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html#command-line-installing-appimage)
   - [Alternative Linux builds](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html#command-line-installing-alternative-linux)
-- **Windows**:
-  - Follow the discussions for
-    [Windows](https://github.com/aws/q-command-line-discussions/discussions/15)
-  - Or [use it on Windows with WSL](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html#command-line-installing-windows)
-- **Remote machines**
-  - [Autocomplete in SSH](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-autocomplete-ssh.html)
 
-## 🚀 Start Contributing
+## Contributing
+
+Thank you so much for considering to contribute to Amazon Q.
+
+Before getting started, see our [contributing docs](CONTRIBUTING.md#security-issue-notifications).
 
 ### Prerequisites
 
@@ -28,15 +21,13 @@
   - Xcode 13 or later
   - Brew
 
-### 1. Clone repo
+#### 1. Clone repo
 
 ```shell
 git clone https://github.com/aws/amazon-q-developer-cli.git
 ```
 
-### 2. Setup
-
-#### 2. Install Rust toolchain using [Rustup](https://rustup.rs):
+#### 2. Install the Rust toolchain using [Rustup](https://rustup.rs):
 
 ```shell
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -70,180 +31,55 @@
...
 ## 🏗️ Project Layout
 
-Several projects live here:
-
-- [`autocomplete`](packages/autocomplete/) - The autocomplete react app
-- [`dashboard`](packages/dashboard-app/) - The dashboard react app
-- [`figterm`](crates/figterm/) - figterm, our headless terminal/pseudoterminal that
-  intercepts the user’s terminal edit buffer.
-- [`q_cli`](crates/q_cli/) - the `q` CLI, allows users to interface with Amazon Q Developer from
+- [`chat_cli`](crates/chat_cli/) - the `q` CLI, allows users to interface with Amazon Q Developer from
```
brandonskiser, line 50: How do we want to frame calling binaries across the autocomplete `q` and this chat binary? Should we just call this `qchat` instead?

README.md
```diff
@@ -70,180 +31,55 @@
...
-- [`tests/`](tests/) - Contain integration tests for the projects
-
-Below is a high level architecture of how the different components of the app and
-their IPC:
-
-![architecture](docs/assets/architecture.svg)
-
-
-
+- [`scripts/`](scripts/) - Contains ops related scripts
```
brandonskiser: ops and build related scripts

docs/agent.md
```diff
@@ -0,0 +1,285 @@
+# The Agent Format
```
brandonskiser: This doc should probably include a definition for `agent` - start with an overview/summary of what an agent is, then follow with "an agent is defined by a configuration file called the agent manifest" or something

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+- [`inputSchema`](#the-input-schema-field) --- The input schema of the agent.
+- [`mcpServers`](#the-mcp-servers-field) --- The MCP servers the agent has access to.
+- [`tools`](#the-tools-field) --- The tools available to the agent.
+- [`allowedTools`](#the-allowed-tools-field) --- Tools that can be used without prompting.
+- [`toolsSettings`](#the-tools-settings-field) --- Configuration for specific tools.
+- [`resources`](#the-resources-field) --- Declarative resources and context.
+
+### The `name` field
+
+The `name` field specifies the name of the agent. This is used for identification and display purposes.
```
brandonskiser: The actual implementation should probably use newtypes that enforce the schema we outline for these fields (if it isn't already), and reference this doc so we don't forget about updating if anything changes

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+
+A pre-release part can be added after a dash such as `1.0.0-alpha`.
+The pre-release part may be separated with periods to distinguish separate
+components. Numeric components will use numeric comparison while
+everything else will be compared lexicographically.
+For example, `1.0.0-alpha.11` is higher than `1.0.0-alpha.4`.
+
+### The `description` field
+
+The `description` field provides a description of what the agent does to be read by both humans and machines. It's important that descriptions succinctly define an agent behavior, as these descriptions may take up LLM context in certain workflows.
```
brandonskiser: The description will *always* take up LLM context when used as a tool, right? Should probably call that out

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+Model identifiers can be found on the [AWS Bedrock documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html).
+
+As of now, the two supported models are
+
+1. `anthropic.claude-3-7-sonnet-20250219-v1:0`
+2. `anthropic.claude-sonnet-4-20250514-v1:0`
+
+### The `inputSchema` field
+
+The `inputSchema` field defines the expected input format for the agent. Because agents can be executed as MCP tools, this aligns directly with the input schema of MCP tools themselves. Input parameters can be templated in the agent using the following syntax:
```
brandonskiser: "input format for the agent" is confusing to me personally, in actuality `inputSchema` defines the arguments for the agent when used as a tool.

Also can't assume the reader knows that this is essentially just json schema, we should call that out separate from any notion of MCP tools.

Also I have no idea how to use input parameters as explained here, probably include an example or add more information

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+
+```
+${my-input-parameter}
+```
+
+Note that this is the same syntax used for environment variables, in this way, environment variables act as user-defined inputs while input parameters service as machine inputs.
+
+```json
```
brandonskiser: what is this? just an example? If we're saying task is required for all agents, then I would want to disagree

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+      "task": {
+        "type": "string",
+        "description": "The task to perform"
+      }
+    },
+    "required": ["task"]
+  }
+}
+```
+
```
mschrage: I don't think this is currently supported?

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+    "@my-enterprise-mcp.my-tool": {
+      "some-configuration-value": true
+    }
+  }
+}
+```
+
+### The `resources` field
+
+The `resources` field defines declarative resources that provide context to the agent. At this time, these resources are handled uniquely by individual client implementations.
```
brandonskiser: > these resources are handled uniquely by individual client implementations

what does this mean? how is templating expected to be used here?

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+    "file://my-excellent-prompt.md",
+    "file://${workspace}",
+    "file://my-mcp-resource.json@builder-mcp"
+  ]
+}
+```
+
+## Complete Example
+
+Here's a complete example of an agent manifest:
```
brandonskiser: Do we want to include a hooks section?

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+
+```json
+{
+  "name": "rust-developer-agent",
+  "version": "1.2.0",
+  "description": "A specialized agent for Rust development tasks",
+  "model": "anthropic.claude-sonnet-4-20250514-v1:0",
```
mschrage: this isn't the format models are defined in the Q CLI - also I don't think we currently support a `model` field? (We should!)

docs/agent.md
```diff
@@ -0,0 +1,285 @@
...
+    "fs-write": {
+      "allowedPaths": ["src/**", "tests/**", "Cargo.toml"]
+    }
+  },
+  "resources": [
+    "You are a principal Rust engineer who writes safe, efficient backend code.",
+    "file://rust-style-guide.md",
+    "file://${workspace}/README.md",
+    "file://project-context.md@workspace-mcp"
+  ],
```
mschrage: I don't think we support this?

docs/agent.md
chaynabors: yes, will do later. Incremental progress is good here

docs/agent.md
chaynabors: maybe, doesn't have to for now. Conforming to the spec isn't a hard requirement for us, but lets us do it later. Customers can't complain if they don't follow the spec

docs/agent.md
chaynabors: I think so, but it isn't in the spec I wrote for some reason. Seems like I'll need to talk to some folks.

docs/agent.md
chaynabors: spoke to felix already, will sync with you

docs/agent.md
chaynabors: we don't yet

docs/agent.md
chaynabors: its not

PR #2285, opened 2025-07-10
chore: Add telemetry event for slash command execution

evanliu048:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/hooks.rs
+++ b/crates/chat-cli/src/cli/chat/cli/hooks.rs
@@ -448,6 +448,10 @@ impl HooksArgs {
             skip_printing_tools: true,
         })
     }
+
+    pub fn subcommand_name(&self) -> Option<&'static str> {
+        self.subcommand.as_ref().map(|s| s.name())
+    }
 }
 
 #[deny(missing_docs)]
@@ -632,6 +636,18 @@ impl HooksSubcommand {
         Ok(ChatState::PromptUser {
             skip_printing_tools: true,
         })
+    }
+
+    pub fn name(&self) -> &'static str {
+        match self {
+            HooksSubcommand::Add { .. } => "add",
+            HooksSubcommand::Remove { .. } => "remove",
+            HooksSubcommand::Enable { .. } => "enable",
+            HooksSubcommand::Disable { .. } => "disable",
+            HooksSubcommand::EnableAll => "enableall",
+            HooksSubcommand::DisableAll => "disableall",
+            HooksSubcommand::Show => "show",
+        }
     }
 }
 
--- a/crates/chat-cli/src/cli/chat/cli/hooks.rs
+++ b/crates/chat-cli/src/cli/chat/cli/hooks.rs
@@ -644,8 +644,8 @@ impl HooksSubcommand {
             HooksSubcommand::Remove { .. } => "remove",
             HooksSubcommand::Enable { .. } => "enable",
             HooksSubcommand::Disable { .. } => "disable",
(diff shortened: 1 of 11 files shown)


crates/chat-cli/telemetry_definitions.json
```diff
@@ -35,6 +35,26 @@
...
+      "type": "string",
+      "description": "The slash command that was executed (e.g., 'context', 'compact', 'model')"
+    },
+    {
+      "name": "codewhispererterminal_chatSlashSubcommand",
+      "type": "string",
+      "description": "The subcommand of the slash command, if applicable (e.g., 'list' for '/prompts list')"
+    },
+    {
+      "name": "codewhispererterminal_chatSlashCommandResult",
```
brandonskiser: If we're using `TelemetryResult` then we should probably just stick with the `"reason"` field instead

(description is off - says 'Success' but we'd actually be sending 'Succeeded' I think, judging from `TelemetryResult::Succeeded`)

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1386,16 +1386,42 @@ impl ChatSession {
 
             match SlashCommand::try_parse_from(args) {
                 Ok(command) => {
+                    let command_name = match orig_args[0].as_str() {
```
brandonskiser: not a fan of duplicating this naming unless we really have to - can we instead define `SlashCommand::command_name` and `SlashCommand::subcommand_name` instead? This way whenever we add more slash commands, we'll have to match on the enum. It's more verbose but imo much more maintainable

crates/chat-cli/telemetry_definitions.json
evanliu048: Thanks! I’ve updated the metric to use the existing `result` and `reason` fields for consistency with other events, and removed `codewhispererterminal_chatSlashCommandResult` and `codewhispererterminal_chatSlashCommandFailureReason`.

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Thanks for the suggestion! Agreed that using the enum is more maintainable in the long run. Refactored this part!

crates/chat-cli/src/cli/chat/cli/hooks.rs
```diff
@@ -633,6 +637,18 @@ impl HooksSubcommand {
...
         })
     }
+
+    pub fn name(&self) -> &'static str {
+        match self {
+            HooksSubcommand::Add { .. } => "add",
+            HooksSubcommand::Remove { .. } => "remove",
+            HooksSubcommand::Enable { .. } => "enable",
+            HooksSubcommand::Disable { .. } => "disable",
+            HooksSubcommand::EnableAll => "enableall",
```
brandonskiser: `enable-all` and `disable-all` are the actual names

crates/chat-cli/src/cli/chat/cli/tools.rs
```diff
@@ -383,4 +387,14 @@ impl ToolsSubcommand {
             skip_printing_tools: true,
         })
     }
+
+    pub fn name(&self) -> &'static str {
+        match self {
+            ToolsSubcommand::Schema => "schema",
+            ToolsSubcommand::Trust { .. } => "trust",
+            ToolsSubcommand::Untrust { .. } => "untrust",
+            ToolsSubcommand::TrustAll => "trustall",
```
brandonskiser: `trust-all` is the actual name

brandonskiser (approved):

PR #2292, opened 2025-07-12
feat(agent): adds mcp json backwards compatibility

dingfeli:
*Issue #, if available:*

*Description of changes:*
- Agent config now supports ["*"] in its tools and servers field
- Migration now fills the aforementioned field in `tools` and `mcp_servers`.
- Adds support for "@native" to specify all native tools in `tools`.

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -81,7 +81,7 @@ mod wrapper_types;
 /// Another example is the mcp config. To support backwards compatibility of users existing global
 /// mcp.json, we allow users to supply a list of mcp server names they want the agent to
 /// instantiate. But in order for this config to actually be useful to the CLI, it needs to contain
-/// actual information about the comamnd, not just the list of names. Thus the "warm" state in this
+/// actual information about the command, not just the list of names. Thus the "warm" state in this
 /// field would be a filtered version of [McpServerConfig], while the "cold" state could be either.
 ///
 /// Where agents are instantiated from their config, we would need to convert them from "cold" to
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -140,7 +140,7 @@ impl Default for Agent {
             description: Some("Default agent".to_string()),
             prompt: Default::default(),
             mcp_servers: Default::default(),
-            tools: NATIVE_TOOLS.iter().copied().map(str::to_string).collect::<Vec<_>>(),
+            tools: vec!["*".to_string()],
             alias: Default::default(),
             allowed_tools: {
                 let mut set = HashSet::<String>::new();
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -38,6 +38,7 @@ use tracing::{
 };
 pub use wrapper_types::{
     CreateHooks,
+    McpServerConfigWrapper,
     OriginalToolName,
     PromptHooks,
     ToolSettingTarget,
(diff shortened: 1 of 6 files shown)


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -60,16 +66,38 @@ use crate::util::{
...
+/// who would need to write these config), the agent config has two states of existence: "cold" and
+/// "warm".
+///
+/// A "cold" state describes the config as it is written. And a "warm" state is an alternate form
+/// of the same config, modified for the convenience of the business logic that relies on it in the
+/// application.
+///
+/// For example, the "cold" state does not require the field of "path" to be populated. This is
+/// because it would be redundant and tedious for user to have to write the path of the file they
+/// had created in said file. This field is thus populated during its parsing.
```
chaynabors, line 80: Please use serde decorations.
https://serde.rs/attr-skip-serializing.html

Manually deserialize into "warm" and cold rather than adding generics to the type.

Serde has all of this functionality built in.

chaynabors:
PhantomData has a very specific use case:
https://doc.rust-lang.org/nomicon/phantom-data.html

crates/chat-cli/src/cli/agent/mod.rs
dingfeli: It's more than just skipping though. 
It's essentially a transformation of certain fields based on a source of truth independent of what is in the config. 

There is an alternate approach which is to deserialize from seed (where the seed would store these aforementioned independent config). 
I haven't looked into it but I would imagine the same is true for serializing (that is also to be implemented manually). 

I think both approaches would take similar amount of work and are similar (though I do think manually implementing ser/de introduces a bit more complexity).

chaynabors: The agent owner will never need or want to stomach this complexity. If you've given an agent to something then that is the source of truth. You're implementing your own polymorphism when you don't even need it.

I am aware of what fields are being transformed, I read the code. The point is that it's the same type regardless so adding this generic is a no-op, and you're leaking complexity into the rest of the codebase with it.

chaynabors (dismissed):
Please fix use of generics.

crates/chat-cli/src/cli/agent/legacy/mod.rs
```diff
@@ -21,6 +21,11 @@ use crate::os::Os;
 use crate::util::directories;
 
 pub async fn migrate(os: &mut Os) -> eyre::Result<Vec<Agent>> {
+    let has_migrated = os.database.get_has_migrated()?;
```
brandonskiser, line 24: Curious - where is this function called? It feels like this should be idempotent instead, so don't fail if we've already migrated

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -130,20 +157,83 @@ impl Default for Agent {
...
+            .retain(|_name, config| !config.is_from_legacy_mcp_json);
+
+        Ok(())
+    }
+
+    /// This function mutates the agent to a state that is usable for runtime.
+    /// Practically this means to convert some of the fields value to their usable counterpart.
+    /// For example, we populate the agent with its file name, convert the mcp array to actual
+    /// mcp config and populate the agent file path.
+    pub fn thaw(&mut self, path: &Path, global_mcp_config: Option<&McpServerConfig>) -> eyre::Result<()> {
```
brandonskiser: Why is `feeze` and `thaw` required to be separate from just having e.g. `save_agent` and `load_agent` methods? Asking since this API is a bit unintuitive imo, like `path` could be anything? Even referencing another file path entirely?

It would be nice to instead have all of this translation logic wrapped inside a single source, `save` and `load`

We could just set file path as part of the agent config with `#[serde(skip)]` or something maybe?

crates/chat-cli/src/cli/mcp.rs
```diff
@@ -89,7 +89,8 @@ pub struct AddArgs {
     /// Arguments to pass to the command
     #[arg(long, action = ArgAction::Append, allow_hyphen_values = true, value_delimiter = ',')]
     pub args: Vec<String>,
-    /// Where to add the server to.
+    /// Where to add the server to. If an agent name is not supplied, the changes shall be made to
+    /// the global mcp.json
```
brandonskiser, line 93: Default shouldn't be to update the global mcp config imo since we want to deprecate it. Instead, fallback to the default agent?

Have a separate `--global` optional flag instead which can be more easily marked as deprecated

dingfeli: But what happens if users don't have a global set? Especially now that we don't have a step to prompt user to set as the last step of the migration anymore.

crates/chat-cli/src/cli/agent/mod.rs
dingfeli: It actually is only used in methods inside of Agent. Outside of this module these methods are not used directly. They are called indirectly via `to_str_pretty` and `load`. 

I'll make these methods private to avoid confusion.

crates/chat-cli/src/cli/agent/legacy/mod.rs
dingfeli: If it fails we actually just get an empty array. It's just easier to reason about this way.

PR #2312, opened 2025-07-15
Terminal bench manual trigger github action

arjun37602:
*Description of changes:* Adds a new yaml, bash script, and python file that allow terminal-bench to run on a GitHub runner through a manual trigger as a github action. These runs use arjbal AWS credentials for sigv4 + s3 bucket access.

### DIFF
--- a/.github/workflows/terminal-bench.yaml
+++ b/.github/workflows/terminal-bench.yaml
@@ -43,7 +43,7 @@ jobs:
     - name: Configure AWS credentials
       uses: aws-actions/configure-aws-credentials@v4
       with:
-        role-to-assume: arn:aws:iam::${{ secrets.TB_ACCOUNTID }}:role/ArjunPersonal
+        role-to-assume: arn:aws:iam::${{ secrets.AWS_ROLE }}:role/ArjunPersonal
         aws-region: us-east-1
        
 
--- a/.github/workflows/terminal-bench.yaml
+++ b/.github/workflows/terminal-bench.yaml
@@ -35,7 +35,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.13'
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/terminal-bench.yaml
+++ b/.github/workflows/terminal-bench.yaml
@@ -1,0 +1,59 @@
+# This is a terminal-bench workflow that is manually triggered
+# Template taken from https://github.com/actions/starter-workflows/blob/main/automation/manual.yml for reference 
+
+name: TB workflow
+
+# Controls when the action will run. Workflow runs when manually triggered using the UI
+# or API.
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      name:
+        # Friendly description to be shown in the UI instead of 'name'
+        description: 'Run terminal bench'
+        # Default value if no value is explicitly provided
+        default: 'all'
+        # Input has to be provided for the workflow to run
+        required: true
+        # The data type of the input
+        type: string
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  run-benchmark:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up Python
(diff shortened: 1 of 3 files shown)


.github/workflows/terminal-bench.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+# Controls when the action will run. Workflow runs when manually triggered using the UI
+# or API.
+on:
+  workflow_dispatch:
+    # Inputs the workflow accepts.
+    inputs:
+      name:
+        # Friendly description to be shown in the UI instead of 'name'
+        description: 'Run terminal bench'
```
brandonskiser: What should `name` be set to? The description doesn't seem to make this clear

.github/workflows/terminal-bench.yaml
```diff
@@ -0,0 +1,84 @@
...
+        required: true
+        # The data type of the input
+        type: string
+        
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  run-benchmark:
+    runs-on: ubuntu-latest-8-cores
+    env:
+      FIGCHAT_GAMMA_ID: ${{ secrets.FIGCHAT_GAMMA_ID }}
```
brandonskiser: maybe `FIGCHAT_GAMMA_ACCOUNT_ID` instead, so it's clearer what id is being referred to

Also - is this account id needed outside of just for setting the IAM role and bucket name? If it's not, then can we instead pass `CHAT_DOWNLOAD_ROLE_ARN` and `CHAT_BUILD_BUCKET_NAME`? This matches the build script used in autocomplete, and lets us more easily update to Prod infra if we want to

.github/workflows/terminal-bench.yaml
```diff
@@ -0,0 +1,84 @@
...
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+    - name: Cleanup and free disk space
+      run: |
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf /opt/ghc
+        sudo rm -rf "/usr/local/share/boost"
+        sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```
brandonskiser, line 35: where is `AGENT_TOOLSDIRECTORY` set?

.github/workflows/terminal-bench.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install terminal-bench
+    
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: arn:aws:iam::${{ secrets.AWS_TB_ROLE }}:role/ArjunPersonal
```
brandonskiser: `secrets.AWS_TB_ROLE` is actually an account id, right? Probably instead replace this whole line with `secrets.AWS_TB_ROLE`, so we can set arbitrary roles and not just yours

terminal-bench-test/main.py
```diff
@@ -0,0 +1,50 @@
...
+class AmazonQCLIAgent(AbstractInstalledAgent):
+
+    @staticmethod
+    def name() -> str:
+        return "Amazon Q CLI"
+
+    def __init__(self, model_name: str | None = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._model_name = model_name
+        self._start_url = 'https://amzn.awsapps.com/start'
```
brandonskiser: what are these member variables used for? doesn't seem to be referenced anywhere

.github/workflows/terminal-bench.yaml
arjun37602: Inspired from https://stackoverflow.com/questions/75536771/github-runner-out-of-disk-space-after-building-docker-image. TLDR: I believe it removes unnecessary tools to create more disk storage and set auto

.github/workflows/terminal-bench.yaml
arjun37602: Only the fields nested inside name are meant to be filled: https://github.com/actions/starter-workflows/blob/main/automation/manual.yml

.github/workflows/terminal-bench.yaml
brandonskiser: maybe include a link to that post so we don't lose the context here, since this seems like a very strange implementation detail of github actions specifically

brandonskiser (approved):

PR #2330, opened 2025-07-17
Add retry mechanism for ModelOverloadedError

mkreder:
*Issue #, if available:* 2315

*Description of changes:*

This PR adds a retry mechanism with exponential backoff for the "Amazon Q is having trouble responding right now" error (ModelOverloadedError). When this error occurs, the CLI will automatically retry up to 10 times with increasing delays between attempts, rather than immediately showing an error message to the user.

Changes

- Modified the `send_message` method to include retry logic with exponential backoff
- Set a maximum of 10 retry attempts starting at 500ms
- Added random jitter to prevent thundering herd problems
- Only retry for `ModelOverloadedError` which corresponds to "Amazon Q is having trouble responding right now"
- Added debug logs to show retry attempts and progress

Testing

- All existing tests pass
- Manually tested under ModelOverloadedError conditions

### DIFF
--- a/crates/chat-cli/src/api_client/mod.rs
+++ b/crates/chat-cli/src/api_client/mod.rs
@@ -31,6 +31,7 @@ pub use endpoints::Endpoint;
 pub use error::ApiClientError;
 use parking_lot::Mutex;
 pub use profile::list_available_profiles;
+use rand;
 use serde_json::Map;
 use tracing::{
     debug,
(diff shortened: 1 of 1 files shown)


mkreder:
One user in the AWS Community Builder Slack was asking if we should add this exception to the retry mechanism too:

```
Amazon Q is having trouble responding right now: 
   0: unhandled error (InternalServerException)
   1: service error
   2: unhandled error (InternalServerException)
   3: Error { code: "InternalServerException", message: "Encountered an unexpected error when processing the request, please try again.", aws_request_id: "af70fbd9-01c1-48d6-85b7-bffe57a9c704" }

Location:
   crates/chat-cli/src/cli/chat/mod.rs:862

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

I don't think this needs to be retried, because 500 errors could be caused because of so many other things, but just wanted to bring it up.

mkreder:
looks like this has been implemented in the latest version

PR #2344, opened 2025-07-18
fix: prompt user to model selection when throttled and retry their prompt

evanliu048:
*Issue #, if available:*

*Description of changes:*


evanliu048:
implement in https://github.com/aws/amazon-q-developer-cli-autocomplete/pull/451

PR #2419, opened 2025-07-29
chore: integrate backend API and remove hardcoded model options

evanliu048:
*Issue #, if available:*

*Description of changes:*
This PR removes the hardcoded `MODEL_OPTIONS` and replaces it with a dynamic fetch via the `listAvailableModels` API. The result is cached for the duration of the CLI session to avoid redundant network calls and improve responsiveness. No need to map the model ID to a user-friendly name on the client side, as the backend will handle the mapping.

image


crates/chat-cli/src/api_client/mod.rs
```diff
@@ -66,13 +68,17 @@ pub const X_AMZN_CODEWHISPERER_OPT_OUT_HEADER: &str = "x-amzn-codewhisperer-opto
 // TODO(bskiser): confirm timeout is updated to an appropriate value?
 const DEFAULT_TIMEOUT_DURATION: Duration = Duration::from_secs(60 * 5);
 
+type ModelListResult = (Vec<Model>, Model);
```
brandonskiser: nit - this would probably be better as a struct with `{ models: Vec<Model>, default: Model }` instead. Also I see this type duplicated below but not used with the alias `ModelListResult`

crates/chat-cli/src/auth/builder_id.rs
```diff
@@ -477,12 +477,6 @@ impl BuilderIdToken {
             Some(_) => TokenType::IamIdentityCenter,
         }
     }
-
-    /// Check if the token is for the internal amzn start URL (`https://amzn.awsapps.com/start`),
-    /// this implies the user will use midway for private specs
-    pub fn is_amzn_user(&self) -> bool {
```
brandonskiser, line 483: guessing this is removed because it's unused - might be helpful to just mark this as `#[allow(unused_variables)]` since I wouldn't be surprised if we end up needing it again in the future

crates/chat-cli/src/cli/chat/cli/model.rs
```diff
@@ -8,46 +9,123 @@ use crossterm::{
...
+            model_name: None,
+            context_window_tokens: 200_000,
+        }
+    }
+
+    pub fn display_name(&self) -> &str {
+        self.model_name.as_deref().unwrap_or(&self.model_id)
+    }
+}
+impl<'de> Deserialize<'de> for ModelInfo {
```
brandonskiser: Instead of implementing `Deserialize` manually, can we just change the name of the `model` field inside `ConversationState`? So instead we'd have:
```
    /// Legacy, unused
    /// Model explicitly selected by the user in this conversation state via `/model`.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub model: Option<String>,

    /// Model explicitly selected by the user in this conversation state via `/model`.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub model_info: Option<ModelInfo>,
```

this way, we won't have to manually implemented Deserialize which is a bit strange

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -314,22 +312,33 @@ impl ChatArgs {
...
+            .database
+            .settings
+            .get_string(Setting::ChatDefaultModel)
+            .and_then(|model_id| {
+                models
+                    .iter()
+                    .find(|m| m.model_id == model_id)
+                    .map(|m| m.model_id.clone())
+            })
+        {
```
brandonskiser: Can we print an error message to warn the user if their default model from the settings is missing? Since we're changing model names with this release, lots of people might have an incorrect name set up going forward

evanliu048: Thanks for checking! 
We didn’t change the model name in this release — the API still returns both the name and ID as `claude-4-sonnet`. In the past, if a user set an arbitrary value in the settings, it simply wouldn’t take effect and no warning was shown. Do we want to start notifying users about this now?

evanliu048: But thanks for checking since I need to compare model name first!

crates/chat-cli/src/auth/builder_id.rs
evanliu048: Sure! Added back. Thanks for sharing this tip!

crates/chat-cli/src/api_client/mod.rs
evanliu048: Yes changed this to be a struct. I might just forgot I defined above ....

crates/chat-cli/src/cli/chat/cli/model.rs
evanliu048: Yes, I thought about doing it that way too! I just didn’t want to keep an extra pub model: Option<String> field around since it’s effectively unused now and just there for backward compatibility. I’m fine to add it if we think keeping it for clarity and avoiding the manual Deserialize is worth the extra field. Let me know your thoughts!

brandonskiser: I think we can just remove the `pub` on it and keep it as an unused field, with a comment explaining why it's unused (for backwards compatibility) and to never add it for anything else going forward.

Think this is far better than a manual `Deserialize` impl - if the user loads a previous conversation, we just reset the active model to whatever the default is

crates/chat-cli/src/cli/chat/mod.rs
brandonskiser: Oh in that case I suppose we don't need to add any notification, although it would still be a nice to have imo!

evanliu048: Totally agree! Right now, if the user runs q chat --model <xx> with an invalid value, they do get a notification. It’s only when the value comes from the settings that there’s no notification — it just silently doesn’t take effect

crates/chat-cli/src/cli/chat/cli/model.rs
evanliu048: Makes sense — I dropped the manual Deserialize and keep a non-pub legacy model: `Option<String> `on `ConversationState` for backward compatibility. We’ll ignore it at runtime and fall back to the default model if no `model_info` is present

brandonskiser (approved):

PR #2426, opened 2025-07-30
fix(agent): directory change

dingfeli:
*Issue #, if available:*

*Description of changes:*
Following a discussion with flare team, we have decided disambiguate the config directories of the two products. 
Hence forth, the cli shall source the following for its agent config:
- `~/.aws/amazonq/cli-agents` for global agents
- `.amazonq/cli-agents` for local agents

Also changed / fixed is the hiding of default agent name in the prompt when it is selected (this was the behavior previously, but it was broken when we switched from "default" to "q_cli_default" because the filtering done previously was with a hard coded value". 

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -71,7 +71,7 @@ use crate::util::{
     directories,
 };
 
-const DEFAULT_AGENT_NAME: &str = "q_cli_default";
+pub const DEFAULT_AGENT_NAME: &str = "q_cli_default";
 
 #[derive(Debug, Error)]
 pub enum AgentConfigError {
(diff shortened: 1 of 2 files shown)


brandonskiser (approved):

abest0:
This change is small, however it does introduce a discontinuity with the documentation--a documentation update would be greatly appreciated.

Most documentation states that local workspace directory is `.amazonq/` and the expectation is that a local `mcp.json` in `<workspace_root>/.amazonq/mcp.json` should load, however this is not the case. The [chat_legacy_workspace_mcp_config](https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/util/directories.rs#L162) does not seem to be matching.

~~With version `1.41.1`, after checking the codebase, a quick change (moving an mcp config to `<workspace_root>/.amazonq/cli-agents/mcp.json`) loaded the a test MCP server as expected.~~

Okay, more digging and I did locate the [legacy-profile-to-agent-migration](https://github.com/aws/amazon-q-developer-cli/blob/5bf5afa782bad244245e0f4a29da84e3e85063d0/docs/legacy-profile-to-agent-migration.md).

PR #2476, opened 2025-08-02
fix(agent): tool settings

dingfeli:
*Issue #, if available:*

*Description of changes:*
As titled. 

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
+++ b/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
@@ -83,6 +83,11 @@ impl ExecuteCommand {
 
         // Check if each command in the pipe chain starts with a safe command
         for cmd_args in all_commands {
+            let cmd_as_str = cmd_args.join(" ");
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -83,6 +83,11 @@ impl ExecuteCommand {
 
         // Check if each command in the pipe chain starts with a safe command
         for cmd_args in all_commands {
+            let cmd_as_str = cmd_args.join(" ");
+            if allowed_commands.contains(&cmd_as_str) {
```
alex-bellomo, line 87: Just a heads up, I noticed that the `shlex split` command will sometimes add or remove quotes. This could cause some commands to miss matching here (try something like `aws accountid=0123456789`.

I have another PR open for using regex for matching, would love to get your 2c on it if you get a moment :) https://github.com/aws/amazon-q-developer-cli/pull/2483

dingfeli:
Closing this in favor of https://github.com/aws/amazon-q-developer-cli/pull/2483

Issue #2478, opened 2025-08-02
bug: Workspace specific mcp config (.amazonq/mcp.json) is not loaded


everett1992:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

Ubuntu

### Expected behaviour

https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-mcp-understanding-config.html

Documentation says that q chat will load mcp tools defined in `.amazonq/mcp.json` - Specific to the current workspace, however that file is never loaded.

### Actual behaviour

q chat never loads .amazonq/mcp.json

### Steps to reproduce

```
mkdir q-chat-workspace
cd $_
mkdir .amazonq
echo > .amazonq/mcp.json '{ 
  "mcpServers": {
    "false": {
      "command": "false",
      "timeout": 6000
    }
  }
}' 
q chat
```

q chat does not load the mcp server, there's no error or printed. If you write the same file to the global config (~/.aws/amazonq/mcp.json) you will get an error, so there's nothing wrong with the file format.
```
⚠ 0 of 1 mcp servers initialized. Servers still loading:
 - false
```

Running strace to see what files are opened I see that it never references the documented path.
However, it does read some files from .amazonq.
```
$ strace -f q chat |& rg '.amazonq'
[pid 1097374] open("~/q-chat-workspace/.amazonq/cli-agents", O_RDONLY|O_LARGEFILE|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
[pid 1097374] open("~/.aws/amazonq/cli-agents", O_RDONLY|O_LARGEFILE|O_CLOEXEC|O_DIRECTORY <unfinished ...>
[pid 1097354] stat("~/.aws/amazonq/cli-agents/agent_config.json.example", {st_mode=S_IFREG|0644, st_size=658, ...}) = 0
[pid 1097374] open("~/.aws/amazonq/mcp.json", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 1097354] stat("~/q-chat-workspace/.amazonq", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
[...]

PR #2479, opened 2025-08-03
fix: load workspace MCP config in default agent (#2478)

StoneDot:
The default agent was not loading workspace-specific MCP configuration (.amazonq/mcp.json), only the global config (~/.aws/amazonq/mcp.json). This fix ensures workspace MCP configs are loaded and take precedence over global configs when using the default agent.

*Issue #, if available:*
#2478

*Description of changes:*
- Modified `Agents::load()` to read workspace MCP configuration (`.amazonq/mcp.json`) for the default agent
- Workspace configs now override global configs as expected

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -63,6 +63,7 @@ use crate::cli::agent::hook::{
     Hook,
     HookTrigger,
 };
+use crate::cli::chat::tool_manager::workspace_mcp_config_path;
 use crate::database::settings::Setting;
 use crate::os::Os;
 use crate::util::{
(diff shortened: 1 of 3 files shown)


brandonskiser:
In my opinion this MCP loading paradigm is confusing enough, and adding this extra case where a local `mcp.json` is loaded only for the default agent doesn't make sense.

Instead, `useLegacyMcpJson` should probably load *both* the global `mcp.json` and workspace `mcp.json`. Thoughts @mschrage  @dingfeli ?

brandonskiser:
@StoneDot From an offline discussion - if you would be interested in updating this behavior to instead also load workspace MCP configuration for any agent with `"useLegacyMcpJson": true` then we would approve!

Essentially `"useLegacyMcpJson": true` would result in the exact same MCP loading behavior as with profiles.

StoneDot:
Thanks for the feedback! I've implemented the suggested approach where `useLegacyMcpJson` now consistently controls both global and workspace MCP config loading for all agent types.

The implementation became more extensive than initially planned, but it eliminates the confusing special case for the default agent. Now all agents with `useLegacyMcpJson: true` load both global and workspace configs (with workspace taking precedence), providing the consistent behavior you requested.

Could you please review the updated implementation? @brandonskiser

dingfeli:
Apologies. I was working on this issue and had pushed a fix for this https://github.com/aws/amazon-q-developer-cli/pull/2516. I should have looked around in the PR queue prior to working on this. 

Terribly sorry. Very much appreciate the effort!

PR #2483, opened 2025-08-04
Regex-formatted allowed commands for execute_bash tool

alex-bellomo:
*Issue #, if available:* None.

*Description of changes:*
- This PR adds support for regex-formatted allowed commands for the execute_bash tool. This will give developers greater flexibility to allowlist commands. For example, you could allowlist `aws authenticate --read-only --account-id [0-9]{10} (the main use case I had in mind when implementing this PR).
- Today, the allowed commands feature of the execute_bash tool supports only the *first* command in the chain. For example, allowlisting `git` would allow `git`, and all `git` subcommands. Allowing `git status` would *not* allow `git status`, since the current logic only examines the first argument in the command (`git status` != `git`). Adding regex support will add more flexibility to the allowed commands setting, and also allow users to allow commands that are longer than one argument in length (e.g. `git status`).

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
+++ b/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
@@ -6,6 +6,7 @@ use crossterm::style::{
     Color,
 };
 use eyre::Result;
+use regex::Regex;
 use serde::Deserialize;
 use tracing::error;
 
@@ -47,6 +48,17 @@ impl ExecuteCommand {
     pub fn requires_acceptance(&self, allowed_commands: Option<&Vec<String>>, allow_read_only: bool) -> bool {
         let default_arr = vec![];
         let allowed_commands = allowed_commands.unwrap_or(&default_arr);
+
+        let has_regex_match = allowed_commands
+            .iter()
+            .map(|cmd| Regex::new(cmd))
+            .filter(Result::is_ok)
+            .flatten()
+            .any(|regex| regex.is_match(&self.command));
+        if has_regex_match {
+            return false;
+        }
+
         let Some(args) = shlex::split(&self.command) else {
             return true;
         };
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -47,6 +48,17 @@ impl ExecuteCommand {
...
         let default_arr = vec![];
         let allowed_commands = allowed_commands.unwrap_or(&default_arr);
+
+        let has_regex_match = allowed_commands
+            .iter()
+            .map(|cmd| Regex::new(cmd))
+            .filter(Result::is_ok)
+            .flatten()
+            .any(|regex| regex.is_match(&self.command));
+        if has_regex_match {
```
alex-bellomo, line 58: Checking this here was a deliberate choice - IMO we should allow commands with `&&` if there is a direct match. This would allow the agent to run more complex commands if needed.

One other thing to consider here would be forcing the use of regex anchors to match the entire string. As-is, a partial match would succeed here. Figured we could leave that up to the users, but up to discuss that behavior more.

alex-bellomo: Thought about this a bit more and figured adding `\A` and `\Z` anchors to every regex pattern would be a safe way to enforce a full match. Even if the users adds them, a duplicate anchor doesn't do anything. Can raise a revision if that seems like it'd be appropriate 👍

dingfeli:
Thank you so much. This is definitely more well thought out than https://github.com/aws/amazon-q-developer-cli/pull/2476. 

Going to test this out some more before stamping this.

crates/chat-cli/src/cli/chat/tools/execute/mod.rs
brandonskiser: I really like the anchor approach!

So in this case, if I allow `git status` we'd have: `\Agit status\Z`, so if I want to explicitly enable dangerous things like `&&` and `|`, I would have to explicitly add it to the regex `\Agit status [&|]*` or something?

I would be very cautious about foregoing all of the dangerous checks just because I want to let `git status` run without requiring approval (I don't want `git status && rm -rf /` to be automatically approved)

PR #2484, opened 2025-08-04
Implementation of knowledge base directory setting

alex-bellomo:
*Issue #, if available:* None.

*Description of changes:*
- This change adds support for setting a knowledge base `base_dir` in the tool settings. This would allow different agents to access different knowledge bases. This would keep the knowledge bases for a particular agent clean - e.g. an oncall-helper agent would not recieve documentation about the team charter, or other unnecessary information.
- This is a rough draft implementation. I am not sure about the pattern of passing the session or agent around, and would like feedback on the approach. I've left comments below in/around sections that I'm particularly unsure of.

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/knowledge.rs
+++ b/crates/chat-cli/src/cli/chat/cli/knowledge.rs
@@ -1,4 +1,5 @@
 use std::io::Write;
+use std::path::Path;
 
 use clap::Subcommand;
 use crossterm::queue;
@@ -12,6 +13,8 @@ use semantic_search_client::{
     OperationStatus,
     SystemStatus,
 };
+use serde::Deserialize;
+use tracing::error;
 
 use crate::cli::chat::tools::sanitize_path_tool_arg;
 use crate::cli::chat::{
@@ -54,6 +57,13 @@ enum OperationResult {
     Error(String),
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Settings {
+    #[serde(default)]
+    pub(crate) base_dir: String,
+}
+
 impl KnowledgeSubcommand {
     pub async fn execute(self, os: &Os, session: &mut ChatSession) -> Result<ChatState, ChatError> {
         if !Self::is_feature_enabled(os) {
@@ -90,7 +100,27 @@ impl KnowledgeSubcommand {
         }
     }
 
+    fn get_knowledge_base_dir(session: &mut ChatSession) -> Option<impl AsRef<Path> + use<>> {
+        let agent = session.conversation.agents.get_active();
+        if agent.is_none() {
+            return None;
+        }
+
+        agent.unwrap().tools_settings.get("knowledge")?.get("base_dir");
+        match agent.unwrap().tools_settings.get("knowledge") {
+            Some(settings) => match serde_json::from_value::<Settings>(settings.clone()) {
+                Ok(settings) => Some(settings.base_dir),
+                Err(e) => {
+                    error!("Failed to deserialize tool settings for execute_bash: {:?}", e);
+                    None
+                },
+            },
+            None => None,
+        }
+    }
+
     async fn execute_operation(&self, os: &Os, session: &mut ChatSession) -> OperationResult {
+        let knowledge_base_dir = Self::get_knowledge_base_dir(session);
         match self {
             KnowledgeSubcommand::Show => {
                 match Self::handle_show(session).await {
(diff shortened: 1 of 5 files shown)


crates/chat-cli/src/cli/chat/cli/knowledge.rs
```diff
@@ -90,25 +100,48 @@ impl KnowledgeSubcommand {
...
+            KnowledgeSubcommand::Add { path } => Self::handle_add(os, path, knowledge_base_dir).await,
+            KnowledgeSubcommand::Remove { path } => Self::handle_remove(os, path, knowledge_base_dir).await,
+            KnowledgeSubcommand::Update { path } => Self::handle_update(os, path, knowledge_base_dir).await,
             KnowledgeSubcommand::Clear => Self::handle_clear(session).await,
-            KnowledgeSubcommand::Status => Self::handle_status().await,
-            KnowledgeSubcommand::Cancel { operation_id } => Self::handle_cancel(operation_id.as_deref()).await,
+            KnowledgeSubcommand::Status => Self::handle_status(knowledge_base_dir).await,
+            KnowledgeSubcommand::Cancel { operation_id } => {
+                Self::handle_cancel(operation_id.as_deref(), knowledge_base_dir).await
+            },
```
alex-bellomo, line 138: Would it be better to pass the session in here?

crates/chat-cli/src/cli/chat/tools/knowledge.rs
```diff
@@ -542,4 +548,22 @@ impl Knowledge {
...
+            Some(settings) => match serde_json::from_value::<Settings>(settings.clone()) {
+                Ok(settings) => Some(settings.base_dir),
+                Err(e) => {
+                    error!("Failed to deserialize tool settings for execute_bash: {:?}", e);
+                    None
+                },
+            },
+            None => None,
+        }
+    }
```
alex-bellomo, line 568: Is there a better way to share this function between the knowledge base tool and setting?

crates/chat-cli/src/cli/chat/tools/mod.rs
```diff
@@ -107,15 +107,15 @@ impl Tool {
...
+    pub async fn invoke(&self, os: &Os, stdout: &mut impl Write, agent: Option<&Agent>) -> Result<InvokeOutput> {
         match self {
             Tool::FsRead(fs_read) => fs_read.invoke(os, stdout).await,
             Tool::FsWrite(fs_write) => fs_write.invoke(os, stdout).await,
             Tool::ExecuteCommand(execute_command) => execute_command.invoke(stdout).await,
             Tool::UseAws(use_aws) => use_aws.invoke(os, stdout).await,
             Tool::Custom(custom_tool) => custom_tool.invoke(os, stdout).await,
             Tool::GhIssue(gh_issue) => gh_issue.invoke(os, stdout).await,
-            Tool::Knowledge(knowledge) => knowledge.invoke(os, stdout).await,
+            Tool::Knowledge(knowledge) => knowledge.invoke(os, stdout, agent).await,
```
alex-bellomo, line 118: Very unsure of this - is there another way for us to pass tool settings? Looks like we've started to pass the agent for requires_acceptance, so figured this would be an acceptable pattern. Again, open to feedback though.

crates/chat-cli/src/util/knowledge_store.rs
```diff
@@ -32,21 +33,21 @@ pub struct KnowledgeStore {
 
 impl KnowledgeStore {
     /// Get singleton instance
-    pub async fn get_async_instance() -> Arc<Mutex<Self>> {
+    pub async fn get_async_instance(path: Option<impl AsRef<Path>>) -> Arc<Mutex<Self>> {
```
alex-bellomo, line 36: Because any invocation of this could technically create the instance, we need to *always* pass the path if possible.

crates/chat-cli/src/cli/chat/cli/knowledge.rs
kensave: Agreed, the less parameters the better specially if more than 1 parameter has same type.

crates/chat-cli/src/cli/chat/tools/knowledge.rs
kensave: IMO, it's fine if the tool has the responsibility to know where this needs to be extracted, careful with usage of unwrap here. So we can gracefully handle failure scenarios.

crates/chat-cli/src/util/knowledge_store.rs
kensave: Yeah, that's the downside of a Lazy singleton

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1881,7 +1881,10 @@ impl ChatSession {
                 ev.is_accepted = true;
             });
 
-            let invoke_result = tool.tool.invoke(os, &mut self.stdout).await;
+            let invoke_result = tool
+                .tool
```
kensave, line 1885: Nice

kensave:
Now agents have knowledge specific directories.

PR #2506, opened 2025-08-06
feat: Moves to public CDN.

kensave:
* Points to Hosted CDN published models
* Adds new client hosted_model_client to download artifacts from CDN.
* Updates Candle to remove references to HF client. and fix issues if any with hooks

*Issue #, if available:*
Screenshot 2025-08-06 at 8 18 21 PM

*Description of changes:*

### DIFF
--- a/crates/semantic-search-client/Cargo.toml
+++ b/crates/semantic-search-client/Cargo.toml
@@ -33,8 +33,9 @@ bm25 = { version = "2.2.1", features = ["language_detection"] }
 
 # Common dependencies for all platforms
 anyhow = "1.0"
-ureq = "2.12"
-zip = "0.6"
+reqwest = { workspace = true }
+zip = { version = "4.3.0", default-features = false, features = ["deflate", "time"] }
+tokio-stream = "0.1.17"
 
 # Candle dependencies - not used on Linux ARM
 [target.'cfg(not(all(target_os = "linux", target_arch = "aarch64")))'.dependencies]
--- a/crates/semantic-search-client/src/client/hosted_model_client.rs
+++ b/crates/semantic-search-client/src/client/hosted_model_client.rs
@@ -4,19 +4,33 @@
 //! instead of directly from Hugging Face. Models are distributed as zip files
 //! containing model.safetensors and tokenizer.json files.
 
-use std::path::Path;
+#[cfg(test)]
 use std::fs;
-use anyhow::{Result as AnyhowResult, Context};
-use tracing::{info, debug, error};
-use indicatif::{ProgressBar, ProgressStyle};
+use std::path::Path;
+
+use anyhow::{
+    Context,
+    Result as AnyhowResult,
+};
+use indicatif::{
+    ProgressBar,
+    ProgressStyle,
+};
+use reqwest;
+use tracing::{
+    debug,
+    error,
+};
 
 /// Progress callback type for download operations
 pub type ProgressCallback = Box<dyn Fn(u64, u64) + Send + Sync>;
 
-/// Hosted model client for downloading models from CDN (synchronous like original HF)
+/// Hosted model client for downloading models from CDN (synchronous)
 pub struct HostedModelClient {
-    /// Base URL for the CDN (e.g., "https://desktop.gamma-us-east-1.codewhisperer.ai.aws.dev/models")
+    /// Base URL for the CDN
     base_url: String,
+    /// HTTP client
+    client: reqwest::Client,
 }
 
 impl HostedModelClient {
@@ -33,10 +47,13 @@ impl HostedModelClient {
     /// let client = HostedModelClient::new("http://example.com/models".to_string());
     /// ```
     pub fn new(base_url: String) -> Self {
-        Self { base_url }
+        Self {
+            base_url,
+            client: reqwest::Client::new(),
+        }
     }
 
-    /// Download a model if it doesn't exist locally (synchronous like original HF api.repo().get())
+    /// Download a model if it doesn't exist locally (asynchronous)
     ///
     /// # Arguments
     ///
@@ -46,8 +63,8 @@ impl HostedModelClient {
     /// # Returns
     ///
     /// Result indicating success or failure
-    pub fn ensure_model(&self, model_name: &str, target_dir: &Path) -> AnyhowResult<()> {
-        self.ensure_model_with_progress(model_name, target_dir, None)
+    pub async fn ensure_model(&self, model_name: &str, target_dir: &Path) -> AnyhowResult<()> {
+        self.ensure_model_with_progress(model_name, target_dir, None).await
     }
 
     /// Download a model if it doesn't exist locally with optional progress callback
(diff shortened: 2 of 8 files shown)


crates/semantic-search-client/Cargo.toml
```diff
@@ -33,14 +33,15 @@ bm25 = { version = "2.2.1", features = ["language_detection"] }
 
 # Common dependencies for all platforms
 anyhow = "1.0"
+ureq = "2.12"
```
brandonskiser: Why use `ureq` instead of `reqwest`? Which is already a workspace dependency?

All of our code is already pretty heavily based on tokio and async so it should just be an async API right?

crates/semantic-search-client/Cargo.toml
```diff
@@ -33,14 +33,15 @@ bm25 = { version = "2.2.1", features = ["language_detection"] }
...
+zip = "0.6"
```
brandonskiser: Any reason why to use v0.6 instead of latest? https://lib.rs/crates/zip/versions

crates/semantic-search-client/src/config.rs
```diff
@@ -83,6 +86,8 @@ impl Default for SemanticSearchConfig {
             timeout: 30000, // 30 seconds
             base_dir: get_default_base_dir(),
             max_files: 10000, // Default limit of 10000 files
+            //Gamma endpoint: "https://desktop.gamma-us-east-1.codewhisperer.ai.aws.dev/models"
+            hosted_models_base_url: "https://desktop-release.q.us-east-1.amazonaws.com/models".to_string(),
```
brandonskiser, line 89: Can this be kept in a consts file instead?

brandonskiser (approved):

crates/semantic-search-client/src/config.rs
```diff
@@ -83,6 +86,8 @@ impl Default for SemanticSearchConfig {
             timeout: 30000, // 30 seconds
             base_dir: get_default_base_dir(),
             max_files: 10000, // Default limit of 10000 files
+            //Gamma endpoint: "https://desktop.gamma-us-east-1.codewhisperer.ai.aws.dev/models"
```
kkashilk: Let's remove references to Gamma endpoint

crates/semantic-search-client/Cargo.toml
kkashilk: +1, let's use the latest version to avoid any potential CVEs that maybe present in older versions

crates/semantic-search-client/src/client/hosted_model_client.rs
```diff
@@ -0,0 +1,303 @@
...
+        target_dir: &Path,
+        progress_callback: Option<ProgressCallback>
+    ) -> AnyhowResult<()> {
+        // Check if model already exists and is valid
+        if self.is_model_valid(target_dir)? {
+            info!("Model '{}' already exists and is valid", model_name);
+            return Ok(());
+        }
+
+        info!("Downloading hosted model: {}", model_name);
```
kkashilk: change `info!` -> `debug!` wherever possible

crates/semantic-search-client/src/client/hosted_model_client.rs
```diff
@@ -0,0 +1,303 @@
...
+        let zip_filename = format!("{}.zip", model_name);
+        let zip_url = format!("{}/{}", self.base_url, zip_filename);
+        let zip_path = target_dir.join(&zip_filename);
+
+        info!("Constructing download URL:");
+        info!("  Base URL: {}", self.base_url);
+        info!("  Model name: {}", model_name);
+        info!("  Zip filename: {}", zip_filename);
+        info!("  Final URL: {}", zip_url);
+        info!("  Target path: {:?}", zip_path);
```
kkashilk: nit: `debug!`

crates/semantic-search-client/src/client/hosted_model_client.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        // Create target directory if it doesn't exist
+        if let Some(parent) = target_dir.parent() {
+            fs::create_dir_all(parent)
+                .context("Failed to create parent directories")?;
+        }
+        fs::create_dir_all(target_dir)
```
kkashilk: `create_dir_all` will create the parent directory if it doesn't exist - https://doc.rust-lang.org/beta/std/fs/fn.create_dir_all.html

kkashilk (changes requested):

crates/semantic-search-client/src/config.rs
kensave: Yes

crates/semantic-search-client/src/client/hosted_model_client.rs
kensave: Will do.

crates/semantic-search-client/src/client/hosted_model_client.rs
kensave: Will update

crates/semantic-search-client/Cargo.toml
kensave: Moved reqwest and async.

crates/semantic-search-client/Cargo.toml
kensave: Updating.

crates/semantic-search-client/src/config.rs
kensave: This is already the constants files for the crate.

crates/semantic-search-client/src/client/hosted_model_client.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// use semantic_search_client::client::HostedModelClient;
+    /// let client = HostedModelClient::new("http://example.com/models".to_string());
+    /// ```
+    pub fn new(base_url: String) -> Self {
+        Self {
+            base_url,
+            client: reqwest::Client::new(),
```
brandonskiser, line 52: We should instead use the reqwest client initialization we have here, which configures the user agent (+ other TLS related things that I'm not 100% sure of honestly) - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/request.rs#L31

Since we prob don't want to mess with the dependencies between these crates, this could instead be parameterized on the reqwest client instead?

brandonskiser (approved):

crates/semantic-search-client/src/client/hosted_model_client.rs
```diff
@@ -0,0 +1,354 @@
...
+        );
+
+        Ok(valid)
+    }
+
+    /// Check if model files exist and are valid
+    async fn is_model_valid(&self, target_dir: &Path) -> AnyhowResult<bool> {
+        let model_path = target_dir.join("model.safetensors");
+        let tokenizer_path = target_dir.join("tokenizer.json");
+
```
kkashilk, line 286: should this check be done based on the model that we're using?
E.g - 
```
all-MiniLM-L6-v2 -> needs these files to be present
another-model -> might need different set of files
default_case -> unsupported-model, consider updating q-cli
```

separately, if hosted_model_client is mainly responsible for downloading/extracting models in general, 
then this check could be moved out to where we call to download the model.

kkashilk: useful to consider this if/when we let users configure/choose different embedding models

kkashilk (approved):

PR #2512, opened 2025-08-06
feat(agent): adds globbing support for tool settings deny

dingfeli:
*Issue #, if available:*

*Description of changes:*
As titled.
image


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -317,7 +317,7 @@ impl Agent {
 pub enum PermissionEvalResult {
     Allow,
     Ask,
-    Deny,
+    Deny(Vec<String>),
```
brandonskiser, line 318: Can we add a doc comment on this type? Can't really tell what this `Vec<String>` is supposed to represent

From reading the rest of the code, it seems like Deny should be:
```rust
/// Tool is forbidden from being ran
Deny {
  /// List of configured rules that caused the tool use to be denied
  matches: Vec<String>
}
```

brandonskiser (approved):

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1821,22 +1821,40 @@ impl ChatSession {
...
                     })
                     || self.conversation.agents.trust_all_tools;
 
-            if denied {
+            if let Some(match_set) = denied_match_set {
+                let formatted_set = match_set.into_iter().fold(String::new(), |mut acc, rule| {
+                    acc.push_str(&format!("\n  - {rule}"));
+                    acc
+                });
+
```
evanliu048, line 1845: Nit: Consider hiding \A and \z in denied rule printout to improve readability

evanliu048 (approved):

PR #2513, opened 2025-08-06
feat: Add conditional compilation for Knowledge command using feature…

kensave:
… flag

- Add 'knowledge' feature flag to chat-cli Cargo.toml
- Make Knowledge command conditional with #[cfg(feature = "knowledge")]
- Restore original database setting logic when feature is enabled
- Hide Knowledge command from help when feature is disabled
- Fix unused import/variable warnings with conditional compilation
- Knowledge is disabled by default, enabled with --features knowledge

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/crates/chat-cli/Cargo.toml
+++ b/crates/chat-cli/Cargo.toml
@@ -14,6 +14,7 @@ workspace = true
 [features]
 default = []
 wayland = ["arboard/wayland-data-control"]
+knowledge = []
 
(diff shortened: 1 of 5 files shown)


kkashilk (approved):
This works, but you could potentially exclude compiling the feature entirely - https://doc.rust-lang.org/cargo/reference/features.html

PR #2520, opened 2025-08-07
Revert "feat: Add conditional compilation for Knowledge command using…

kensave:
… feature flag (#2513)"

This reverts commit 0e034b4dd28c379d2e87a2317d2ae233a0c9ce02.
The PR does a two main things:

* Enables back the knowledge tool.
* Adds SHA validation for any server downloaded model.  Finger print(sha256) exact match.

### DIFF
--- a/crates/semantic-search-client/src/model_validator.rs
+++ b/crates/semantic-search-client/src/model_validator.rs
@@ -1,0 +1,87 @@
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::Read;
+use std::path::Path;
+
+use sha2::{
+    Digest,
+    Sha256,
+};
+
+/// Validator for model files using allowlisted SHA256 hashes
+pub struct ModelValidator {
+    allowlisted_shas: HashMap<&'static str, Vec<&'static str>>,
+}
+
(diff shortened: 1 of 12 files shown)


crates/semantic-search-client/src/model_validator.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+impl ModelValidator {
+    /// Create a new model validator with official allowlisted SHAs
+    pub fn new() -> Self {
+        let mut allowlisted_shas = HashMap::new();
+
+        // Each file can have multiple valid SHAs (for different versions)
+        allowlisted_shas.insert("model.safetensors", vec![
```
dingfeli, line 28: Is the expectation going forward to keep adding to this list as we change / add more?

dingfeli (approved):

crates/semantic-search-client/src/model_validator.rs
kensave: This is an allow-List. We have a single model so we can add more as we ship more models. This can evolve to deny-list if allow-List seems problematic.

PR #2529, opened 2025-08-08
Quiet mode. A --quiet option to get only the AI output in a non-interactive way

omansour:
--quiet / -q option to be like a --non-interactive option with only the AI output

⚠️ diff are long, but i cant figure another way to achieve this feature

SCR-20250808-nyva

SCR-20250808-nztn


calebphess:
+1 for this. I want `q -q`

PR #2532, opened 2025-08-08
feat: add timestamp for new user prompts

brandonskiser:
*Description of changes:*
- Previously the backend API would add a timestamp as part of the system prompt. Now, it is being removed in favor of letting the client implement this behavior.
- Only adding the timestamp for new user prompts, since tool use results don't have any content to be sent with the prompt.
- Reasoning behind having a timestamp: this enables temporal questions like "what did we do in the last hour" to be answered.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,6 +1224,7 @@ dependencies = [
  "bytes",
  "camino",
  "cfg-if",
+ "chrono",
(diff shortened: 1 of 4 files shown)


dingfeli (approved):
LGTM

kensave (approved):

PR #2533, opened 2025-08-08
Add to-do list functionality to QCLI

kiran-garre:
Adds a to-do list tool (called todo_list) with several commands that allow Q to create a to-do list and update it as it completes tasks, along with a slash command (/todos) that allows users to view and manage their in-progress to-do lists.


build-config/buildspec-linux.yml
```diff
@@ -28,10 +28,10 @@ phases:
       - mise install
       - eval "$(mise activate bash --shims)"
       # Install python deps
-      - pip3 install -r scripts/requirements.txt
+      - pip3 install -r build-scripts/requirements.txt
```
dingfeli: I think there are some errors here with how the merge conflict is handled. 
This is the current state of the file: https://github.com/aws/amazon-q-developer-cli/blob/main/build-config/buildspec-linux.yml#L31

build-config/buildspec-macos.yml
```diff
@@ -25,11 +25,11 @@ phases:
       - eval "$(mise activate zsh --shims)"
```
dingfeli: Same here.

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,350 @@
...
+    Serialize,
+};
+
+use super::InvokeOutput;
+use crate::os::Os;
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(tag = "command")]
+pub enum TodoInput {
+    #[serde(rename = "create")]
```
dingfeli: I think you can just use `#[serde(rename_all = "camelCase")]`.

crates/chat-cli/src/database/mod.rs
```diff
@@ -371,6 +378,32 @@ impl Database {
...
+    }
+
+    pub fn get_current_todo_id(&self) -> Result<Option<String>, DatabaseError> {
+        self.get_entry(Table::State, "current_todo_id")
+    }
+
+    pub fn set_current_todo_id(&self, id: &str) -> Result<(), DatabaseError> {
+        self.set_entry(Table::State, "current_todo_id", id)?;
+        Ok(())
+    }
```
dingfeli: I think there is an edge case here. 

What happens if there are multiple instances of q chat running at the same time. Is the value going to be clobbered / overridden? 

If it is, there are a couple of paths forward: 
- We make the look up key dynamic. We can derive it from conversation id + some sort of suffix. The downside here is that we might need some sort of logic to clean up to rid of ones that are stale. (Maybe introduce a TTL or something)
- Make store this id somewhere else on the conversation state, serializable and deserializable.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2777,6 +2777,54 @@ impl ChatSession {
...
+    pub async fn resume_todo(&mut self, os: &mut Os, id: &str) -> Result<()> {
+        let request_state = self.conversation.create_todo_request(os, id).await;
+        match request_state {
+            Ok(state) => {
+                self.conversation.reset_next_user_message();
+                self.conversation
+                    .set_next_user_message(state.user_input_message.content)
+                    .await;
+            },
+            Err(_) => bail!("Todo could not be resumed"),
```
dingfeli: Might want to propagate the error upwards here.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2777,6 +2777,54 @@ impl ChatSession {
...
+        };
+
+        let conv_state = self
+            .conversation
+            .as_sendable_conversation_state(os, &mut self.stderr, false)
+            .await?;
+
+        let request_metadata: Arc<Mutex<Option<RequestMetadata>>> = Arc::new(Mutex::new(None));
+        let request_metadata_clone = Arc::clone(&request_metadata);
```
dingfeli: I think we are supposed to send telemetry data after `send_message` call? Otherwise there is no need to clone this.

crates/chat-cli/src/cli/chat/cli/todos.rs
```diff
@@ -0,0 +1,254 @@
...
+                            let spinner = if session.interactive {
+                                Some(Spinner::new(
+                                    Spinners::Dots,
+                                    format!("{} {}", "Resuming:".magenta(), entries[index].description.clone()),
+                                ))
+                            } else {
+                                None
+                            };
+
+                            let todo_result = session.resume_todo(os, &entries[index].id).await;
```
dingfeli: Do we want to just make `resume_todo` return a `Result<ChatState, ChatError>` and have the spinner logic within `resume_todo`.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2777,6 +2777,54 @@ impl ChatSession {
...
+                parser::ResponseEvent::ToolUse(e) => tool_uses.push(e),
+                parser::ResponseEvent::ToolUseStart { .. } => (),
+            }
+        }
+        if !tool_uses.is_empty() {
+            self.validate_tools(os, tool_uses).await?;
+        }
+
+        // FIX: hacky? not really sure how this works honestly LOL
+        self.conversation.reset_next_user_message();
```
dingfeli: This seems like a deviation of the established control flow. Is there a reason why we can't just utilize the existing event loop for this?

crates/chat-cli/src/cli/chat/cli/todos.rs
```diff
@@ -0,0 +1,254 @@
...
+                    if entries.is_empty() {
+                        execute!(session.stderr, style::Print("No to-do lists to view!\n"))?;
+                    } else if let Some(index) = fuzzy_select_todos(&entries, "Select a to-do list to view:") {
+                        if index < entries.len() {
+                            let list = match TodoState::load(os, &entries[index].id) {
+                                Ok(list) => list,
+                                Err(_) => {
+                                    return Err(ChatError::Custom("Could not load the selected to-do list".into()));
+                                },
+                            };
```
dingfeli:
```rust
let list = TodoState::load(os, &entries[index].id).ok_or(ChatError::Custom("Could not load the selected to-do list".into())?;
```

Though looking at `fuzzy_select_todos` it is not clear to me whether or not this should be an error.

crates/chat-cli/src/cli/chat/cli/todos.rs
```diff
@@ -0,0 +1,254 @@
...
+                        Err(_) => {
+                            return Err(ChatError::Custom("Could not load current to-do list".into()));
+                        },
+                    };
+                    match state.display_list(&mut session.stderr) {
+                        Ok(_) => execute!(session.stderr, style::Print("\n"))?,
+                        Err(_) => {
+                            return Err(ChatError::Custom("Could not display current to-do list".into()));
+                        },
+                    };
```
dingfeli: Do we want to propagate the error here as well. 

Also a less verbose way of writing this would be: 
```rust 
let state = TodoSate::load(os, &id).map_err(|_| ChatError::Custom("Could not load current to-do list".into())?;
```

crates/chat-cli/src/cli/chat/cli/todos.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub async fn execute(self, os: &mut Os, session: &mut ChatSession) -> Result<ChatState, ChatError> {
+        match self {
+            Self::ClearFinished => {
+                let entries = match os.database.get_all_todos() {
+                    Ok(e) => e,
+                    Err(_) => return Err(ChatError::Custom("Could not get all to-do lists".into())),
+                };
+                let mut cleared_one = false;
+                for (id, value) in entries.iter() {
+                    let temp_struct = match value.as_str() {
```
dingfeli: Perhaps a better name here would be `item_status` or `todo_status` or something more descriptive. 

Also, instead of turning to value to str and turning the str to `TodoState`, maybe we can just use `serde_json::from_value`.

crates/chat-cli/src/cli/chat/tools/tool_index.json
```diff
@@ -281,5 +281,88 @@
...
+          "description": "The command to run. Allowed options are `add`, `complete`."
+        },
+        "tasks": {
+          "description": "Required parameter of `create` command containing the list of DISTINCT tasks to be added to the TODO list.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "task_description": {
```
brandonskiser: Is this the description for the todo list as a whole? It seems that a todo list is some number of tasks (so the create command requires a `tasks` array), but `task_description` is only a single parameter?

crates/chat-cli/src/cli/chat/tools/tool_index.json
```diff
@@ -281,5 +281,88 @@
...
+        },
+        "insert_indices": {
+          "description": "Required parameter of `add` command containing a list of 0-INDEXED positions to insert the new tasks. There MUST be an index for every new task being added.",
+          "type": "array",
+          "items": {
+            "type": "integer"
+          }
+        },
+        "new_description": {
+          "description": "Optional parameter of `add` and `remove` containing a new task description. Use this when the updated set of tasks significantly change the goal or overall procedure of the task.",
```
brandonskiser: What is the definition of a `task`? It seems that `tasks` and `task` are being used interchangeably when the singular `task` is actually referring to an instance of a todo list that contains tasks.

crates/chat-cli/src/database/mod.rs
```diff
@@ -371,6 +378,23 @@ impl Database {
...
+    pub fn get_todo(&self, id: &str) -> Result<Option<TodoState>, DatabaseError> {
+        self.get_json_entry(Table::Todos, id)
+    }
+
+    pub fn set_todo(&self, id: &str, state: &TodoState) -> Result<(), DatabaseError> {
+        self.set_json_entry(Table::Todos, id, state)?;
+        Ok(())
+    }
+
+    pub fn get_all_todos(&self) -> Result<Map<String, Value>, DatabaseError> {
```
brandonskiser: Could this map to a `TodoState` instead of `Value`?

crates/chat-cli/src/database/mod.rs
brandonskiser, line 1: I wonder how much sense it makes to have todo lists stored in state (accessible globally) versus in a current working directory (e.g. inside `.amazonq/todo_lists/`).

What use cases do we have where it makes sense that todo lists should be stored in state regardless of cwd? I could imagine things like MCP flows that never use the filesystem and instead just call API's like updating docs, etc. but then that brings into question how to handle todo lists that are working on feature changes, and then you try to resume a todo list in a completely different directory.

crates/chat-cli/src/cli/chat/cli/todos.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+        Ok(ChatState::PromptUser {
+            skip_printing_tools: true,
+        })
+    }
+
+    /// Convert all to-do list state entries to displayable entries
+    fn get_descriptions_and_statuses(os: &Os) -> Result<Vec<TodoDisplayEntry>> {
+        let mut out = Vec::new();
+        let entries = os.database.get_all_todos()?;
```
brandonskiser: This `database.get_all_todos` api could instead be `(Vec<TodoState>, Vec<DatabaseError>)` so we just get a list of valid todos, and todos that failed to load or deserialize for whatever reason, so we don't have to duplicate this deser logic multiple times maybe

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2788,6 +2790,59 @@ impl ChatSession {
...
+            TODO LIST CONTENTS: {}\n
+            ID: {}\n",
+            contents,
+            id
+        );
+
+        let summary_message = UserMessage::new_prompt(summary_content.clone(), None);
+
+        // Only send the todo_list tool
+        let mut tools = self.conversation.tools.clone();
```
brandonskiser: So we essentially handle this template as if the user typed it themself - therefore, we're retaining the entire conversation history as part of this request - will this cause validation errors if we remove tools?

brandonskiser: doesn't seem to from testing

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,383 @@
...
+
+use super::InvokeOutput;
+use crate::os::Os;
+
+// Local directory to store todo lists
+const TODO_LIST_DIR: &str = ".amazonq/cli-todo-lists/";
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(tag = "command", rename_all = "camelCase")]
+pub enum TodoInput {
```
brandonskiser: nit - call this just `TodoList` to be aligned with the actual tool name in the tool index. Also add some small doc comments to these `pub` types.

also put the impl right below the enum definition

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,383 @@
...
+            if index < self.tasks.len() - 1 {
+                queue!(output, style::Print("\n"))?;
+            }
+        }
+        Ok(())
+    }
+
+    /// Displays a single empty or marked off to-do list task depending on
+    /// the completion status
+    fn queue_next_without_newline(output: &mut impl Write, task: String, completed: bool) -> Result<()> {
```
brandonskiser: In general, if functions don't require `self` then they should probably just be free functions, e.g. `delete_todo` parameterized on an `id` doesn't make sense to be associated with a single in-memory todo

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,383 @@
...
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_millis();
+
+        format!("{timestamp}")
+    }
+
+    pub fn id_to_path(os: &Os, id: &str) -> Result<PathBuf> {
+        Ok(os.env.current_dir()?.join(TODO_LIST_DIR).join(id))
```
brandonskiser: Wonder if we should also store these with a `json` extension? Just a personal preference of mine (for getting easier IDE support if for whatever reason we need to check these files directly)

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,383 @@
...
+    // Removes tasks from the current todo list
+    Remove {
+        remove_indices: Vec<usize>,
+        new_description: Option<String>,
+        current_id: String,
+    },
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct TodoState {
```
brandonskiser: `TodoListState` ?

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,383 @@
...
+            TodoInput::Remove {
+                remove_indices,
+                new_description,
+                current_id: id,
+            } => {
+                let state = TodoState::load(os, id).await?;
+                if has_duplicates(remove_indices) {
+                    bail!("Removal indices must be unique")
+                } else if remove_indices.iter().any(|i| *i > state.tasks.len()) {
+                    bail!("Index is out of bounds");
```
brandonskiser: We could prob make the error messaging here better for the model in case it hallucinates, e.g. tell it which index is out of bounds or give the length of tasks

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,383 @@
...
+
+impl TodoState {
+    pub async fn init_dir(os: &Os) -> Result<()> {
+        os.fs.create_dir_all(os.env.current_dir()?.join(TODO_LIST_DIR)).await?;
+        Ok(())
+    }
+
+    /// Loads a TodoState with the given id
+    pub async fn load(os: &Os, id: &str) -> Result<Self> {
+        let Ok(state_str) = os.fs.read_to_string(Self::id_to_path(os, id)?).await else {
```
brandonskiser: Can we print the actual error here in the error message? This could fail for a reason other than the to-do list not existing

brandonskiser (commented):
Appears to be some issue with the trust behavior with `/tools` - should be trusted
Screenshot 2025-08-13 at 11 27 01 AM

Also can you add tab completion? Add it to the array here - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/prompt.rs#L43

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -649,6 +652,11 @@ impl ChatSession {
             }
         });
 
+        // Create for cleaner error handling for todo lists
+        // This is more of a convenience thing but is not required, so the Result
+        // is ignored
+        let _ = TodoListState::init_dir(os).await;
```
brandonskiser: Any todo list directory creation should really be happening with the todo list code that depends on it, having it here is error prone should this ever be moved in the future.

brandonskiser (approved):

dingfeli (approved):

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,403 @@
...
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+use super::InvokeOutput;
+use crate::os::Os;
+
+// Local directory to store todo lists
+const TODO_LIST_DIR: &str = ".amazonq/cli-todo-lists/";
```
brandonskiser: Can this be a function parameterized on `Os` instead?
Example here - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/util/directories.rs#L162

crates/chat-cli/src/cli/chat/tools/todo.rs
```diff
@@ -0,0 +1,403 @@
...
+
+/// Contains all state to be serialized and deserialized into a todo list
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct TodoListState {
+    pub tasks: Vec<String>,
```
brandonskiser: Why are tasks and completed states stored in separate `Vec`? It's more intuitive to instead have it be something like `Vec<Task>` where the completion state is stored with the task

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2859,6 +2862,53 @@ impl ChatSession {
...
+            TODO LIST CONTENTS: {}\n
+            ID: {}\n",
+            contents,
+            id
+        );
+
+        let summary_message = UserMessage::new_prompt(summary_content.clone(), None);
+
+        // Only send the todo_list tool
+        let mut tools = self.conversation.tools.clone();
```
brandonskiser: This still sends the entire tool spec, you'd have to construct a special request for this

Here is a reference for how the compaction request is created - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/conversation.rs#L528

I think you can then switch to `ChatState::HandleResponseStream` with this conversation state.

Also be sure to call `ChatSession::reset_user_turn` for this as well

brandonskiser (approved):

PR #2539, opened 2025-08-10
feat(chat): /logdump cmd

### LINKED ISSUE #2462: Feature Request: make q chat logs dumpable from CLI
At the moment the existence of the qlogs folder is not well documented, this should probably be addressed as well, but better than that would be some sort of /logdump command that pulls all the logs from where ever they are stored into a zip file in the current directory, as well as any other logs the Service Team would need for investigation.  At the moment the way that I am finding the logs is,

For Mac this is the find call `find /System/Volumes/Data/private/var/folders -name "qchat.log"`

 

For WSL this is the find call `find /mnt/wslg/run/ -name "qchat.log"`

Konippi:
*Issue #, if available:*

resolve #2462

*Description of changes:*
 
 This PR makes q chat logs dumpable by using `/logdump` command that pulls all the logs from where ever they are stored into a zip file in the current directory, as well as any other logs the Service Team would need for investigation.

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/logdump.rs
+++ b/crates/chat-cli/src/cli/chat/cli/logdump.rs
@@ -1,0 +1,224 @@
+use std::io::Write;
+use std::path::{
+    Path,
+    PathBuf,
+};
+
+use clap::Args;
+use crossterm::execute;
+use crossterm::style::{
+    self,
+    Color,
+};
+use time::OffsetDateTime;
+use zip::ZipWriter;
+use zip::write::SimpleFileOptions;
+
+use crate::cli::chat::{
+    ChatError,
+    ChatSession,
+    ChatState,
+};
+use crate::util::directories::logs_dir;
+
+/// Arguments for the logdump command that collects logs for support investigation
+#[derive(Debug, PartialEq, Args)]
+pub struct LogdumpArgs;
+
+impl LogdumpArgs {
+    pub async fn execute(self, session: &mut ChatSession) -> Result<ChatState, ChatError> {
+        execute!(
+            session.stderr,
+            style::SetForegroundColor(Color::Cyan),
+            style::Print("Collecting logs...\n"),
+            style::ResetColor,
+        )?;
+
+        let timestamp = OffsetDateTime::now_local()
+            .unwrap_or_else(|_| OffsetDateTime::now_utc())
+            .format(&time::format_description::well_known::Iso8601::DEFAULT)
+            .unwrap_or_else(|_| "unknown".to_string())
+            .replace(':', "-"); // Replace colons for Windows compatibility
+
+        let zip_filename = format!("q-logs-{}.zip", timestamp);
+        let zip_path = PathBuf::from(&zip_filename);
+
+        match self.create_log_dump(&zip_path).await {
+            Ok(log_count) => {
+                execute!(
+                    session.stderr,
+                    style::SetForegroundColor(Color::Green),
+                    style::Print(format!(
+                        "✓ Successfully created {} with {} log files\n",
+                        zip_filename, log_count
+                    )),
+                    style::ResetColor,
+                )?;
+            },
+            Err(e) => {
+                execute!(
+                    session.stderr,
+                    style::SetForegroundColor(Color::Red),
+                    style::Print(format!("✗ Failed to create log dump: {}\n\n", e)),
+                    style::ResetColor,
+                )?;
+                return Err(ChatError::Custom(format!("Log dump failed: {}", e).into()));
+            },
+        }
+
+        Ok(ChatState::PromptUser {
+            skip_printing_tools: true,
+        })
+    }
+
+    async fn create_log_dump(&self, zip_path: &Path) -> Result<usize, Box<dyn std::error::Error>> {
+        let file = std::fs::File::create(zip_path)?;
+        let mut zip = ZipWriter::new(file);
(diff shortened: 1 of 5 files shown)


brandonskiser:
@Konippi Thank you for the contribution! Our logging story definitely needs more user visibility, and having it under a `/logdump` command makes sense to me - we could add extra help comments about how to e.g. set the logging env var even (doesn't need to happen in this PR, just get the functionality working).

I would prefer it changed though to instead just copy `qchat.log` and `mcp.log`.

My reasoning:
Logs under `qlog` includes everything unrelated to the actual `qchat` binary as well, e.g. the desktop app and `qterm`. Any slash command within chat should probably instead restrict itself to `qchat.log` and (maybe with a `--mcp` argument) `mcp.log`.

If a user is running `/logdump` within `q chat`, it's just a bit strange to couple it with everything else in that directory unrelated to `q chat`. I would expect something like `q logdump` from the top level to have this type of behavior.

We've never had any reason to request logs for the others. Plus side, this also simplifies the initial implementation a bit

konippi:
Thanks for the feedback! You're absolutely right about the scope issue. 

I'll update this PR to focus on `qchat.log` only to keep it simple and get the core functionality working.

For `mcp.log` inclusion with the `--mcp` flag, I'll handle that in a separate PR once this base functionality is merged.

Konippi:
I misunderstood. I will change it to /logdump.

konippi:
I misunderstood. I will change it to /logdump.

kensave:
Please add to the description of this PR how it was tested, and/or pictured which validate the functionality

konippi:
@kensave 
1. `/logdump --help`

```bash
> /logdump --help

Create a zip file with logs for support investigation

Usage: /logdump

Options:
  -h, --help  Print help
```

2. `/logdump`

```bash
> /logdump

Collecting logs...
✓ Successfully created q-logs-2025-09-11T22-51-29.510164000+09-00.zip with 1 log files

# A zip file will be created in current directory.
```

crates/chat-cli/src/cli/chat/cli/logdump.rs
```diff
@@ -0,0 +1,168 @@
...
+impl LogdumpArgs {
+    pub async fn execute(self, session: &mut ChatSession) -> Result<ChatState, ChatError> {
+        execute!(
+            session.stderr,
+            style::SetForegroundColor(Color::Cyan),
+            style::Print("Collecting logs...\n"),
+            style::ResetColor,
+        )?;
+
+        let timestamp = OffsetDateTime::now_local()
```
brandonskiser: Could we just use `Utc::now()` instead? No reason to have local time zone information I would think

crates/chat-cli/src/cli/chat/cli/logdump.rs
```diff
@@ -0,0 +1,168 @@
...
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    use tempfile::TempDir;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_logdump_creates_zip_file() {
```
brandonskiser: Can the logic for collecting the qchat log be restructured here to take a logs directory as a parameter, so that tests aren't collecting from the actual logs directory? Basically just have `Self::create_log_dump(&self, logs_dir: PathBuf)` then move all of the implementation inside there

crates/chat-cli/src/cli/logdump.rs
```diff
@@ -0,0 +1,183 @@
...
+use zip::ZipWriter;
+use zip::write::SimpleFileOptions;
+
+use crate::util::directories::logs_dir;
+
+/// Arguments for the logdump command that collects logs for support investigation
+#[derive(Debug, PartialEq, Args)]
+pub struct LogdumpArgs;
+
+impl LogdumpArgs {
```
brandonskiser: This code is all being copy pasted - can we avoid adding any top level command for this at the moment? Since this would require top level changes to the `q` binary in https://github.com/aws/amazon-q-developer-cli-autocomplete, and `q logdump` just dumping the qchat log also wouldn't make much sense

konippi: Sorry. I forgot to delete this.

konippi:
@brandonskiser 
Thank you for your review.
I've fixed the sections with comments, so please confirm them.

crates/chat-cli/src/cli/chat/cli/logdump.rs
```diff
@@ -0,0 +1,167 @@
...
+        assert!(zip_path.exists());
+
+        // Verify it's a valid zip file by trying to read it
+        let file = fs::File::open(&zip_path).unwrap();
+        let archive = zip::ZipArchive::new(file);
+        assert!(archive.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_add_log_file_to_zip() {
```
brandonskiser: We should just have two tests here, one for an empty zip and one for a logs directory that has the file. No need for a separate test just for `add_log_file_to_zip`

brandonskiser (approved):

crates/chat-cli/src/cli/chat/cli/logdump.rs
konippi: completed!

konippi:
Since I need to continuously resolve conflicts, please merge if there are no issues.

PR #2560, opened 2025-08-11
fix: prompt caching with timestamp included

brandonskiser:
*Description of changes:*
- Currently prompt caching is broken because a new timestamp is being added as part of the context files for every request. This PR fixes this by requiring the timestamp to be passed as an optional parameter when creating new user messages.

### DIFF
--- a/crates/chat-cli/src/cli/chat/message.rs
+++ b/crates/chat-cli/src/cli/chat/message.rs
@@ -286,7 +286,8 @@ impl UserMessage {
         }
     }
 
-    /// Returns a formatted [String] containing [Self::additional_context] and [Self::prompt].
+    /// Returns a formatted [String] containing [Self::additional_context], [Self::timestamp], and
+    /// [Self::prompt].
     fn content_with_context(&self) -> String {
         let mut content = String::new();
 


chaynabors (approved):

dingfeli (approved):

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -320,14 +321,19 @@ impl ConversationState {
 
     pub fn add_tool_results_with_images(&mut self, tool_results: Vec<ToolUseResult>, images: Vec<ImageBlock>) {
         debug_assert!(self.next_message.is_none());
-        self.next_message = Some(UserMessage::new_tool_use_results_with_images(tool_results, images));
+        self.next_message = Some(UserMessage::new_tool_use_results_with_images(
+            tool_results,
+            images,
+            Some(Utc::now()),
```
jguoamz, line 327: I think we don't need timestamp here?

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -320,14 +321,19 @@ impl ConversationState {
...
+        ));
     }
 
     /// Sets the next user message with "cancelled" tool results.
     pub fn abandon_tool_use(&mut self, tools_to_be_abandoned: &[QueuedTool], deny_input: String) {
         self.next_message = Some(UserMessage::new_cancelled_tool_uses(
             Some(deny_input),
             tools_to_be_abandoned.iter().map(|t| t.id.as_str()),
+            Some(Utc::now()),
```
jguoamz, line 336: I think we don't need timestamp here

PR #2563, opened 2025-08-11
Fix and Simplify retry-classifier

kkashilk:
### Problem
Updating retry-classifier after running some local tests:

1. ThrottlingException with MONTHLY_REQUEST_COUNT - Monthly limit errors were not being classified correctly (service returns a 400, instead of a 429 status code in this case)
2. InternalServerException and ServiceUnavailableException - The service returns a 500 status code for both
3. ThrottlingException with INSUFFICIENT_MODEL_CAPACITY - Already handled and classified correctly by aws-sdk retry-rules, so no additional handling required

### DIFF
--- a/crates/chat-cli/src/api_client/retry_classifier.rs
+++ b/crates/chat-cli/src/api_client/retry_classifier.rs
@@ -8,22 +8,10 @@ use aws_smithy_runtime_api::client::retries::classifiers::{
 };
 use tracing::debug;
 
-/// Error marker for monthly limit exceeded errors
 const MONTHLY_LIMIT_ERROR_MARKER: &str = "MONTHLY_REQUEST_COUNT";
-
-/// Error message for high load conditions that should be retried
 const HIGH_LOAD_ERROR_MESSAGE: &str =
     "Encountered unexpectedly high load when processing the request, please try again.";
-
-/// Error message for insufficient model capacity that should be retried
-const INSUFFICIENT_MODEL_CAPACITY_MESSAGE: &str = "I am experiencing high traffic, please try again shortly.";
-
-/// Status codes that indicate service overload/unavailability and should be retried
-const SERVICE_OVERLOAD_STATUS_CODES: &[u16] = &[
-    429, // Too Many Requests - throttling with insufficient model capacity
-    500, // Internal Server Error - requires specific message check for high load conditions
-    503, // Service Unavailable - server is temporarily overloaded or under maintenance
-];
+const SERVICE_UNAVAILABLE_EXCEPTION: &str = "ServiceUnavailableException";
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/api_client/retry_classifier.rs
```diff
@@ -33,112 +21,58 @@ impl QCliRetryClassifier {
...
+    fn is_service_overloaded_error(ctx: &InterceptorContext, body_str: &str) -> bool {
         let Some(resp) = ctx.response() else {
             return false;
         };
 
-        let status_code = resp.status().as_u16();
-
-        // Fail fast: if status code is not in our list, return false immediately
-        if !SERVICE_OVERLOAD_STATUS_CODES.contains(&status_code) {
+        if resp.status().as_u16() != 500 {
```
brandonskiser, line 47: What about throttling errors, with a 429 status code?

crates/chat-cli/src/api_client/retry_classifier.rs
```diff
@@ -8,22 +8,10 @@ use aws_smithy_runtime_api::client::retries::classifiers::{
...
-/// Error message for insufficient model capacity that should be retried
-const INSUFFICIENT_MODEL_CAPACITY_MESSAGE: &str = "I am experiencing high traffic, please try again shortly.";
-
-/// Status codes that indicate service overload/unavailability and should be retried
-const SERVICE_OVERLOAD_STATUS_CODES: &[u16] = &[
-    429, // Too Many Requests - throttling with insufficient model capacity
-    500, // Internal Server Error - requires specific message check for high load conditions
-    503, // Service Unavailable - server is temporarily overloaded or under maintenance
-];
+const SERVICE_UNAVAILABLE_EXCEPTION: &str = "ServiceUnavailableException";
```
brandonskiser, line 14: is this a new check we need in the `send_message` error handler for our API client?

crates/chat-cli/src/api_client/retry_classifier.rs
kkashilk: The rust aws-sdk handles that automatically

kensave: Is this safe to assume? Which other error codes might come here? 

Are we covering all permutations with this?

kensave (approved):

crates/chat-cli/src/api_client/retry_classifier.rs
brandonskiser: Just checked and still seems fine

crates/chat-cli/src/api_client/retry_classifier.rs
kkashilk: I ran this test locally to verify that rust aws-sdk treats that error correctly

PR #2565, opened 2025-08-12
feat: implement hierarchical agent config discovery

jbbbriant:
# Implement hierarchical agent config discovery

## Overview

This pull request modifies local agent discovery to look for `.amazonq/cli-agents/` not just in the current directory but in
all directories above. 

## Problem Statement

The AWS documentation for Q CLI states, under [Managing custom agents](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-custom-agents-management.html) that "Project-level custom agents" are "Available only within the specific project directory and its subdirectories."

However this is not _quite_ correct. Project-level custom agents are available _only_ in the specific project directory, and not in its subdirectories.

This change fixes the behavior to match the documentation.

## Key Changes Made

1. **`crates/chat-cli/src/cli/agent/mod.rs`** - Agent Management
   - modified `Agents::load` to search for local agents starting with the current working directory and then iterating upwards

2. **`crates/chat-cli/src/util/directories.rs`** - Directory Utils
   - Add chat_relative_agent_dir helper function for relative path resolution

### Build Verification ✅
- ✅ **Successful compilation**: `cargo build --release` completes without errors
- ✅ **Clean warnings**: Only expected warnings about unused helper functions
- ✅ **No breaking changes**: All existing functionality preserved

### Manual Testing Performed
- Created ./amazonq/cli-agents/MyAgent.json
- Verified that `chat_cli chat --agent=MyAgent` worked in directory with .amazonq/... and subdirectories.
- Verified that named agent not found in parent of that directory.

## Backward Compatibility
- ✅ All existing functionality preserved
- ✅ No API changes
- ✅ No configuration changes required
- ✅ No breaking changes to existing workflows

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -421,29 +421,37 @@ impl Agents {
                 },
             }
 
-            let Ok(path) = directories::chat_local_agent_dir(os) else {
-                break 'local Vec::<Agent>::new();
-            };
-            let Ok(files) = os.fs.read_dir(path).await else {
+            let mut agents = Vec::<Agent>::new();
+
+            let Ok(mut current_dir) = os.env.current_dir() else {
                 break 'local Vec::<Agent>::new();
             };
 
-            let mut agents = Vec::<Agent>::new();
-            let results = load_agents_from_entries(files, os, &mut global_mcp_config).await;
-            for result in results {
-                match result {
-                    Ok(agent) => agents.push(agent),
-                    Err(e) => {
-                        load_metadata.load_failed_count += 1;
-                        let _ = queue!(
-                            output,
-                            style::SetForegroundColor(Color::Red),
-                            style::Print("Error: "),
-                            style::ResetColor,
-                            style::Print(e),
-                            style::Print("\n"),
-                        );
-                    },
+            loop {
+                if let Ok(maybe_config_dir) = directories::chat_relative_agent_dir(current_dir.clone()) {
+                    if let Ok(files) = os.fs.read_dir(&maybe_config_dir).await {
+                        let results = load_agents_from_entries(files, os, &mut global_mcp_config).await;
+                        for result in results {
+                            match result {
+                                Ok(agent) => agents.push(agent),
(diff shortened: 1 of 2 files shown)


crates/chat-cli/src/util/directories.rs
```diff
@@ -165,6 +165,11 @@ pub fn chat_local_agent_dir(os: &Os) -> Result<PathBuf> {
     Ok(cwd.join(WORKSPACE_AGENT_DIR_RELATIVE))
 }
 
+/// Directory for agent config relative to given path
+pub fn chat_relative_agent_dir(dir: PathBuf) -> Result<PathBuf> {
```
brandonskiser, line 169: probably remove `chat_local_agent_dir` in favor of using this function

brandonskiser (changes requested):
`Agent::get_agent_by_name` would also need to be updated to perform this hierarchical check.

Ideally we just have a single function that does the hierarchical loading, returning a list of agents (and probably a list of errors encountered during the search, e.g. invalid agent configs found)

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -421,29 +421,37 @@ impl Agents {
...
+                                        style::Print(e),
+                                        style::Print("\n"),
+                                    );
+                                },
+                            }
+                        }
+                    }
+                }
+
+                if !current_dir.pop() {
```
dingfeli, line 453: This keeps popping until path len is zero. 
This is probably not what we want right?

dingfeli: Now that I think about it some more maybe it's not that big a of a deal to just go all the way. We are after all only invoking this per launch, and we are never going to traverse back down to anything.

Something else I had just thought about: what happens if as we are traversing upwards we encounter more than one such config dir? Here we are aggregating all of them. Is that the behavior we want? Should we short circuit? Of the two I am not sure which one is less surprising of a behavior. 

In addition to this I think this also exposes something that was missing from the current implementation. We don't have a mechanism for handling agent conflicts in the same scope.

PR #2573, opened 2025-08-12
fix(tools): show all native tools in /tools command

omansour:
Always load native tools regardless of agent configuration so users can see complete tool list with permission status. Fixes empty /tools output when tools have settings but aren't in allowedTools.

SCR-20250812-movk

### DIFF
--- a/crates/chat-cli/src/cli/chat/tool_manager.rs
+++ b/crates/chat-cli/src/cli/chat/tool_manager.rs
@@ -897,11 +897,13 @@ impl ToolManager {
                 serde_json::from_str::<HashMap<String, ToolSpec>>(include_str!("tools/tool_index.json"))?
                     .into_iter()
                     .filter(|(name, _)| {
+                        use crate::cli::chat::tools::NATIVE_TOOLS;
                         name == DUMMY_TOOL_NAME
                             || is_allow_all
                             || is_allow_native
                             || tool_list.contains(name)
                             || tool_list.contains(&format!("@builtin/{name}"))
+                            || NATIVE_TOOLS.contains(&name.as_str())
                     })
                     .collect::<HashMap<_, _>>();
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/tool_manager.rs
```diff
@@ -897,11 +897,13 @@ impl ToolManager {
                 serde_json::from_str::<HashMap<String, ToolSpec>>(include_str!("tools/tool_index.json"))?
                     .into_iter()
                     .filter(|(name, _)| {
+                        use crate::cli::chat::tools::NATIVE_TOOLS;
```
brandonskiser, line 900: We shouldn't include native tools if the user hasn't configured the use of one

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -711,7 +711,16 @@ impl Agents {
         if tool_trusted || self.trust_all_tools {
             format!("* {}", "trusted".dark_green().bold())
         } else {
-            self.default_permission_label(tool_name)
+            // For custom agents, only show defaults if using the built-in default agent
```
brandonskiser, line 714: There's a larger issue with what we're printing here in the first place that I'd like to address in a single PR (not necessarily putting this on you of course!)

We have hard coded permissions for some tools so whatever we print here doesn't make any sense for some tools - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/tools/mod.rs#L112

In my mind, default permission label shouldn't even exist

omansour: problem is coming from the absence of the tool entry in the agent configuration. I will close this PR as there is no problem: no tools activated => no tools listed.

PR #2604, opened 2025-08-13
fix:  Add client-side modelName mapping for backward compatibility 

evanliu048:
*Issue #, if available:*

*Description of changes:*
The service side has changed the model name from` claude-4-sonnet` to `claude-sonnet-4`. Add a client-side mapping to ensure backward compatibility when invoking Q or updating q settings for chat.defaultModel with either model. This should also remain consistent with the plugin.

### Invoke chat with `claude-4-sonnet`:
image

### Invoke chat with `claude-sonnet-4`:
Screenshot 2025-08-13 at 2 49 19 PM

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -16,6 +16,7 @@ mod server_messenger;
 mod skim_integration;
 mod token_counter;
 pub mod tool_manager;
+use crate::cli::chat::cli::model::find_model;
 pub mod tools;
 pub mod util;
 use std::borrow::Cow;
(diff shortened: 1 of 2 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -16,6 +16,7 @@ mod server_messenger;
 mod skim_integration;
 mod token_counter;
 pub mod tool_manager;
+use crate::cli::chat::cli::model::find_model;
```
brandonskiser: nit - weird import formatting here, above `pub mod` lines

brandonskiser (approved):

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Changed the use statement to be just under `use crate::cli::chat::cli::SlashCommand`; Thanks for pointing that out!

kensave (approved):

PR #2606, opened 2025-08-14
fix(agent): tool permission override

dingfeli:
*Issue #, if available:*
Premise

Currently in the agent config we require a tool to be included in “allowedTools” in order for its corresponding entry in “toolsSettings” to take effect. 

Feedback 1

If the toolsSettings is permission related, this would result in a stricter permission for the tool than if it were to have no toolsSettings at all. From users feedback, the consensus is that this is unintuitive. 

For example, config A is looser than config B:
```json
// Config A
{
  "allowedTools": [
    "fs_read",
    "execute_bash",
    "fs_write"
  ],
  "toolsSettings": {}
}
```
```json
// Config B
{
  "allowedTools": [
    "fs_read",
    "execute_bash",
    "fs_write"
  ],
  "toolsSettings": {
    "execute_bash": {
      "allowedCommands": ["git status"] // now only git status is permitted
    }
  }
}
```

Feedback 2

Having an entry for a given tool in toolsSettings without also including them in allowedTools would potentially break “trust tool” command. This is because trust tool effectively includes the tool in allowedTools, which in this scenario practically “activates” the entry in toolsSettings.

For example, consider the following:

1. User erroneously include a permission related toolsSettings entry for execute_bash without having included it in allowedTools. 
2. User goes to invoke said tool. When prompted, user replies with “t” to trust the tool for the rest of the session with the expecation that they should not be prompted again for execute_bash invocation.
3. The previously “dormant” toolsSettings now gets activated with the newly added entry of execute_bash in allowedTools. As mentioned above, this actually narrows the list of allowed commands.
4. User is once again prompted for execute_bash the next time it runs should the command not match what is stated in allowedCommands in the toolsSettings entry for execute_bash.

*Description of changes:*
Iff a tool appears in both allowedTools & toolSettings show this warning on launch

* You have trusted the “execute_bash” tool, which will override the toolSettings that have been defined: <dump toolSettings JSON>

Tool permission evaluation:

1. If tool invocation is rejected by a toolSetting, deny it
2. If tool is in allowedTools list, invoke it
3. If tool is [t]rusted at runtime, invoke it
4. If tool invocation conforms to toolSetting, invoke

image

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -254,37 +254,37 @@ impl Agent {
                 // currently we only have four native tools that offers tool settings
                 match allowed_tool.as_str() {
                     "fs_read" => {
-                        if let Some(overriden_settings) = FsRead::allowable_field_to_be_overriden(settings) {
+                        if let Some(overridden_settings) = FsRead::allowable_field_to_be_overridden(settings) {
                             queue_permission_override_warning(
                                 allowed_tool.as_str(),
-                                overriden_settings.as_str(),
+                                overridden_settings.as_str(),
                                 output,
                             )?;
                         }
                     },
                     "fs_write" => {
-                        if let Some(overriden_settings) = FsWrite::allowable_field_to_be_overriden(settings) {
+                        if let Some(overridden_settings) = FsWrite::allowable_field_to_be_overridden(settings) {
                             queue_permission_override_warning(
                                 allowed_tool.as_str(),
-                                overriden_settings.as_str(),
+                                overridden_settings.as_str(),
                                 output,
                             )?;
                         }
                     },
                     "use_aws" => {
-                        if let Some(overriden_settings) = UseAws::allowable_field_to_be_overriden(settings) {
+                        if let Some(overridden_settings) = UseAws::allowable_field_to_be_overridden(settings) {
                             queue_permission_override_warning(
                                 allowed_tool.as_str(),
-                                overriden_settings.as_str(),
+                                overridden_settings.as_str(),
                                 output,
                             )?;
                         }
                     },
                     name if name == execute_name => {
-                        if let Some(overriden_settings) = ExecuteCommand::allowable_field_to_be_overriden(settings) {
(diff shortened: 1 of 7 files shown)


evanliu048 (commented):
looks like one ut failed.

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -238,6 +242,58 @@ impl Agent {
             }
         }
 
+        stderr.flush()?;
+
+        Ok(())
+    }
+
+    pub fn validate_tool_settings(&self, output: &mut impl Write) -> Result<(), AgentConfigError> {
```
brandonskiser, line 250: nit - `validate` isn't a clear name for what is happening here, would prefer `print_warning_for_overriden_tool_settings`, and specify this function just checks allowed tools and warns the user for any tool settings that would be overriden with a doc comment.

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -1412,7 +1413,17 @@ mod tests {
         }))
         .unwrap();
 
-        let res = tool.eval_perm(&agent);
+        let res = tool_one.eval_perm(&agent);
+        assert!(matches!(
```
brandonskiser, line 1417: nit - can these assertions be split into multiple lines? For readability, just asserting it's denied and that the two denied paths are in the deny list.

docs/agent-format.md
```diff
@@ -160,6 +160,7 @@ Unlike the `tools` field, the `allowedTools` field does not support the `"*"` wi
 ## ToolsSettings Field
 
 The `toolsSettings` field provides configuration for specific tools. Each tool can have its own unique configuration options.
+Note that specifications that configure allowable patterns will be overridden if the tool is also included in `allowedTools`.
```
evanliu048, line 164: nit: Could we document the exact precedence? e.g., denies > trust (allowedTools) > toolsSettings allow rules

evanliu048 (approved):

brandonskiser (approved):

PR #2607, opened 2025-08-14
feat: add delay tracking interceptor for retry notifications

kkashilk:
- Track delays between retry attempts using request timestamps
- Notify users with two thresholds: 2s (rate limiting) and 5s (suggest model switch)
- Use attempt number from ConfigBag to provide contextual retry information

Screenshot 2025-08-13 at 5 27 28 PM
Screenshot 2025-08-13 at 5 29 00 PM


crates/chat-cli/src/api_client/delay_interceptor.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl Intercept for DelayTrackingInterceptor {
+    fn name(&self) -> &'static str {
+        "DelayTrackingInterceptor"
+    }
+
+    fn read_before_transmit(
+        &self,
+        _context: &BeforeTransmitInterceptorContextRef<'_>,
```
kensave: Any reason to keep this arguments if not used? I'd just remove them.

kensave: Oh that's the signature of base. Approving.

kensave (approved):

crates/chat-cli/src/api_client/delay_interceptor.rs
```diff
@@ -0,0 +1,92 @@
...
+
+            if delay >= self.major_delay_threshold {
+                Self::print_warning(format!(
+                    "Auto Retry #{} delayed by {:.1}s. Service is under heavy load - consider switching models.",
+                    attempt_number,
+                    delay.as_secs_f64()
+                ));
+            } else if delay >= self.minor_delay_threshold {
+                Self::print_warning(format!(
+                    "Auto Retry #{} delayed by {:.1}s due to rate limiting.",
```
brandonskiser: How do we know what the actual retry error is based on the delay? So if it's within 2-5 seconds, we know it's just rate limiting?

kkashilk: Updated the message to outline the issue maybe transient

PR #2619, opened 2025-08-15
fix(agent): tool permission

dingfeli:
*Issue #, if available:*
Fixing https://github.com/aws/amazon-q-developer-cli/pull/2618

*Description of changes:*
Internal discussion ongoing. Will update accordingly. 

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -254,37 +254,37 @@ impl Agent {
                 // currently we only have four native tools that offers tool settings
                 match allowed_tool.as_str() {
                     "fs_read" => {
-                        if let Some(overriden_settings) = FsRead::allowable_field_to_be_overriden(settings) {
+                        if let Some(overridden_settings) = FsRead::allowable_field_to_be_overridden(settings) {
                             queue_permission_override_warning(
                                 allowed_tool.as_str(),
-                                overriden_settings.as_str(),
+                                overridden_settings.as_str(),
                                 output,
                             )?;
                         }
                     },
                     "fs_write" => {
-                        if let Some(overriden_settings) = FsWrite::allowable_field_to_be_overriden(settings) {
+                        if let Some(overridden_settings) = FsWrite::allowable_field_to_be_overridden(settings) {
                             queue_permission_override_warning(
                                 allowed_tool.as_str(),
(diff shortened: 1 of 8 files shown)


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -238,6 +242,58 @@ impl Agent {
             }
         }
 
+        stderr.flush()?;
+
+        Ok(())
+    }
+
+    pub fn validate_tool_settings(&self, output: &mut impl Write) -> Result<(), AgentConfigError> {
```
brandonskiser: name this `print_overridden_permissions`?

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -135,7 +141,11 @@ impl FsRead {
                 let allow_set = {
                     let mut builder = GlobSetBuilder::new();
                     for path in &allowed_paths {
-                        if let Ok(glob) = Glob::new(path) {
+                        let path = path.strip_suffix('/').unwrap_or(path.as_str());
```
brandonskiser: Why do we need to strip `'/'`? Won't that cause issues between relative and absolute paths?

See we do this a lot - can we have a single function that performs this logic? Since it seems a bit obscure

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -102,6 +102,12 @@ impl FsRead {
         }
     }
 
+    pub fn allowable_field_to_be_overridden(settings: &serde_json::Value) -> Option<String> {
```
brandonskiser: This API seems a bit unintuitive to me, no idea how this `pub fn` is being used without checking references and seeing why we're returning a string formatted like this (it's also an associated method w/ no `self`). Could we instead maybe just iterate through builtin tools and call a `print_settings_warnings` or something instead? Or just do the entire logic inside `validate_tool_settings`?

crates/chat-cli/src/cli/chat/tools/fs_read.rs
dingfeli: This is stripping the suffix. 
This is so that when the model pass in some/path/ it would be treated the same as some/path

brandonskiser (approved):

PR #2637, opened 2025-08-18
feat(agent): hot swap

dingfeli:
*Issue #, if available:*
Currently, q chat is not able to freely swap between agents during a session. 
This PR enables the swapping of agents while preserving the conversation.

*Description of changes:*
- Changes prompts to be cached in the orchestrator task that handles backgrounds async events. (These are events such as fetching tools and prompts).
- Adds a slash command that swaps agents within a session. 

Shown below is swapping of agents within a session. Observe the relevant resources are updated (tools, context, and prompts). 

https://github.com/user-attachments/assets/fd90a0d3-739f-4dcb-95f1-e85882e40d6b


crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -134,29 +142,63 @@ impl PathCompleter {
...
+        // will not panic but nothing will be fetched because the thread that is doing
+        // try_recv is also the thread that is supposed to be doing the sending.
+        let mut attempts = 0;
+        let max_attempts = 5;
+        let query_res = loop {
+            match new_receiver.try_recv() {
+                Ok(result) => break result,
+                Err(_e) if attempts < max_attempts - 1 => {
+                    attempts += 1;
+                    std::thread::sleep(std::time::Duration::from_millis(100));
```
evanliu048, line 181: Just curious: how did we decide on sleeping 100ms here?

dingfeli: Trial and error but I think a wide range of values work here.

PR #2639, opened 2025-08-18
feat: added mcp admin level configuration with GetProfile

evanliu048:
*Issue #, if available:*

*Description of changes:*
This PR introduces MCP admin-level configuration support by integrating the `GetProfile` API. It enables administrators to centrally control whether MCP functionality is available to end users.

*Key Changes*
- Added `is_mcp_enabled` check in `ApiClient` using `GetProfile`.
- Extended `ConversationState` with an `mcp_enabled` flag.
- Updated Agent, Agents, and CLI commands (agent, chat, tools, mcp, profile, persist) to respect the MCP enablement setting.
- Display warnings in CLI when MCP is disabled by administrators.

*UX*

https://github.com/user-attachments/assets/d7d0c0f6-15cb-4a56-b392-c34cab18a3f2

tested on my side, MCP enable check API call (500ms) + Agents::load (400ms) took 974.82 ms


crates/chat-cli/src/cli/mcp.rs
```diff
@@ -416,7 +416,14 @@ impl StatusArgs {
 async fn get_mcp_server_configs(os: &mut Os) -> Result<BTreeMap<Scope, Vec<(String, Option<McpServerConfig>, bool)>>> {
     let mut results = BTreeMap::new();
     let mut stderr = std::io::stderr();
-    let agents = Agents::load(os, None, true, &mut stderr).await.0;
+    let mcp_enabled = match os.client.is_mcp_enabled().await {
```
kensave, line 419: `nit`: Here we are accessing directly client all times to obtain is_mcp_enabled, making it very verbose and directly accessing members of client. Wondering if it makes sense to encapsulate this logic, so warning is consistent across all accesses and the check is less verbose.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -251,9 +251,19 @@ impl ChatArgs {
         let conversation_id = uuid::Uuid::new_v4().to_string();
         info!(?conversation_id, "Generated new conversation id");
 
+        // Check MCP status once at the beginning of the session
```
kensave, line 254: Same, this logic seems duplicated in all access to client.

crates/chat-cli/src/cli/agent/root_command_args.rs
```diff
@@ -74,9 +74,16 @@ pub struct AgentArgs {
...
+            Ok(enabled) => enabled,
+            Err(err) => {
+                tracing::warn!(?err, "Failed to check MCP configuration, defaulting to enabled");
+                true
+            },
+        };
         match self.cmd {
             Some(AgentSubcommands::List) | None => {
-                let agents = Agents::load(os, None, true, &mut stderr).await.0;
+                let agents = Agents::load(os, None, true, &mut stderr, mcp_enabled).await.0;
```
kensave, line 86: `nit`: Not relevant right now, I assume this MCP governance will evolve to more than is_enabled. Eventually we would like a data structure which holds this metadata.

crates/chat-cli/src/cli/mcp.rs
evanliu048: Thanks for pointing this out! For the chat subcommands we already gate everything through
if `!session.conversation.mcp_enabled `{ ... }, so the MCP governance check is consistent there.

The places where I still call the API directly are in some q subcommands like `q agent` or `q mcp (status, list)`, since those commands are intended to interact with MCP itself and not in a chat session so need the direct API.

So in practice most chat flows are covered by the session.conversation.mcp_enabled guard, and only the MCP-related subcommands still query the client directly. Let me know if you have any concerns or advice!

crates/chat-cli/src/cli/agent/root_command_args.rs
evanliu048: Good point! If the service returns more metadata in the future, I’ll adjust the implementation accordingly. For now I’m just starting with `is_enabled` since this is the api response.

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -807,6 +839,11 @@ fn default_schema() -> String {
     "https://raw.githubusercontent.com/aws/amazon-q-developer-cli/refs/heads/main/schemas/agent-v1.json".into()
 }
 
+pub fn is_mcp_ref(s: &str) -> bool {
+    // Any tool reference starting with '@' is considered MCP (e.g., "@git", "@git/git_status").
```
brandonskiser: Is this really true though? what about `@builtin`? https://github.com/aws/amazon-q-developer-cli/blob/main/docs/agent-format.md#tools-field

evanliu048: updated the filter logic here to keep `@builtin`. Nice catch Brandon!

brandonskiser (approved):

PR #2647, opened 2025-08-20
feat: implement agent-scoped knowledge base and  context-specific search

kensave:
- fix: Add search_context() method to AsyncSemanticSearchClient for searching specific contexts. If context id is avaiable we should search on specific KB.
- Add unified search_context() method to ContextManager supporting both BM25 and semantic contexts
- Update KnowledgeStore.search() to use context_id parameter when provided
- Refactor knowledge base to use agent-scoped storage with migration support
- Fix SearchError variant usage and resolve clippy warnings
- Ensure all get_async_instance calls properly use agent names
- UX improvement: Sets UX consistency by matching Context UI before merging them both in new /resources

image


crates/chat-cli/src/util/knowledge_store.rs
```diff
@@ -91,91 +92,108 @@ impl AddOptions {
...
+                    Arc::new(Mutex::new(store))
+                })
+                .await
+                .clone())
+        }
     }
 
     /// Migrate legacy knowledge base from old location if needed
-    async fn migrate_legacy_knowledge_base(knowledge_dir: &PathBuf) {
+    async fn migrate_legacy_knowledge_base(agent_dir: &PathBuf, current_agent_name: &str) -> bool {
```
dingfeli: I guess this is just something that we would need to keep doing until the end of the time. 
Food for thought: do we perhaps want to introduce a version here so that migration can be abstracted via a generic so we would be able to have different impl of this based on what the generic is? Not a requirement for this PR I don't think but it would allow us with more liberty to change things.

crates/chat-cli/src/cli/chat/tools/mod.rs
```diff
@@ -120,6 +120,7 @@ impl Tool {
         os: &Os,
         stdout: &mut impl Write,
         line_tracker: &mut HashMap<String, FileLineTracker>,
+        agent: Option<&crate::cli::agent::Agent>,
```
dingfeli, line 123: Not entirely sure how to feel about this. On one hand we already have a param that's not needed by every tool (line_tracker) so we are really not doing anything new here by adding one more. On the other hand if we keep doing this we are going to end up having to pass a bunch of stuff for each tool call, only a fraction of which is actually needed for a given call.  

Should tools be aware of what agent is active currently? That might not be a bad idea. Since we already derive tools from its deserialized primitives in tool manager, what do you think about "decorating" this field there? https://github.com/aws/amazon-q-developer-cli/blob/71c00814247c3c2d6e134c3cbd0f23f6745b1466/crates/chat-cli/src/cli/chat/tool_manager.rs#L774

dingfeli: Although on second thought, if we are saying we want to make tool aware of agent then maybe this should be a part of invoke.

crates/chat-cli/src/cli/chat/cli/knowledge.rs
```diff
@@ -254,14 +236,17 @@ impl KnowledgeSubcommand {
 
     async fn handle_add(
         os: &Os,
+        session: &mut ChatSession,
```
dingfeli, line 239: Is this to future-proof the api? It looks like all we're extracting from it is the agent name.

crates/chat-cli/src/util/knowledge_store.rs
kensave: I think we can continue to make this process more elaborated. With Agents this schema will be easier as in schema configuration we can define the version itself there.

crates/chat-cli/src/cli/chat/tools/mod.rs
kensave: Yeah, I was thinking maybe other tools might need that too.

crates/chat-cli/src/cli/chat/cli/knowledge.rs
kensave: Any suggestion alternative here? As we have the session, we I was thinking we can then inject more information if needed.

crates/chat-cli/src/util/directories.rs
```diff
@@ -190,6 +190,11 @@ pub fn knowledge_bases_dir(os: &Os) -> Result<PathBuf> {
     Ok(home_dir(os)?.join(".aws").join("amazonq").join("knowledge_bases"))
 }
 
+/// The directory for agent-specific knowledge base storage
+pub fn agent_knowledge_dir(os: &Os, agent_name: &str) -> Result<PathBuf> {
+    Ok(knowledge_bases_dir(os)?.join(agent_name))
```
dingfeli: We also allow local agents and we currently don't stipulate that the names have to be globally unique. i.e. you could run into a situation where agents of the same names exists in different scope. It looks like `knowledge_bases_dir` is a global path? How are we resolving the conflicts there?

kensave: The deduplication is based on agent name. Is there any unique identifier we can build for agent?

So it's knowledge_dir/agent_name

kensave: How will it work now if the agent name is used under agents folder? My impression is that the identifier selected for the name of the schema must be also unique.

dingfeli: There is a path field on the agent. What do you think about deriving a key with the path associated with the agent somehow (that's unique for sure).

kensave: Yup that works, we can generate a hash based on it and add the agent name as a suffix.

Issue #2659, opened 2025-08-21
bug: q login always prompts for start url and region even if they are provided via --identity-provider and --region


rostcheck:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

MacOS 15.6

### Expected behaviour

When "q login" is supplied with --license pro, --identity-provider with value, and --region with value, q presents the confirmation URL without requiring the user to enter the identity provider URL and the region.

### Actual behaviour

"q login" still asks for the values for the identity provider and region - they are pre-filled, but still require user confirmation.

### Steps to reproduce

1. Run q logout (if previously logged in)
2. Run q login --license pro --identity-provider "https://your-identity-center.awsapps.com/start" --region us-east-1 --use-device-flow
3. Observe that the command prompts for confirmation despite providing the parameters. They are pre-filled, but still require user confirmation.

### Environment

```yaml
[q-details]
version = "1.14.0"
hash = "c67e487d5e31d50590e8e2e416af4d73b1070168"
date = "2025-08-15T19:11:26.645417Z (5d ago)"
variant = "full"

[system-info]
os = "macOS 15.6.0 (24G84)"
chip = "Apple M3"
total-cores = 8
memory = "16.00 GB"

[environment]
cwd = "/Users/USER/Documents/Source/q_cli_investigation/amazon-q-cli"
cli-path = "/Users/USER/Documents/Source/q_cli_investigation/amazon-q-cli"
os = "Mac"
shell-path = "/bin/bash"
shell-version = "3.2.57"
terminal = "macOS"
install-method = "unknown"

[env-vars]
[...]

PR #2660, opened 2025-08-21
Fix: bug: q login always prompts for start url and region even if the…

rostcheck:
…y are provided via --identity-provider and --region #2659

*Issue #, if available:* 2659

*Description of changes:* Small change to crates/chat-cli/src/cli/user.rs around line 117 to not call input() for values we already have supplied via command line arguments

### DIFF
--- a/crates/chat-cli/src/cli/user.rs
+++ b/crates/chat-cli/src/cli/user.rs
@@ -108,22 +108,21 @@ impl LoginArgs {
                 let (start_url, region) = match login_method {
                     AuthMethod::BuilderId => (None, None),
                     AuthMethod::IdentityCenter => {
-                        let default_start_url = match self.identity_provider {
-                            Some(start_url) => Some(start_url),
-                            None => os.database.get_start_url()?,
+                        let start_url = match self.identity_provider {
+                            Some(url) => url,  // CLI arg provided - use directly
+                            None => {
+                                // No CLI arg - prompt with database default
+                                let default = os.database.get_start_url()?;
+                                input("Enter Start URL", default.as_deref())?
+                            }
                         };
-                        let default_region = match self.region {
-                            Some(region) => Some(region),
-                            None => os.database.get_idc_region()?,
-                        };
-
-                        let start_url = match default_start_url {
-                            Some(url) => url,
-                            None => input("Enter Start URL", None)?,
-                        };
-                        let region = match default_region {
-                            Some(r) => r,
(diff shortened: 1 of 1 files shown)


brandonskiser (changes requested):
This wouldn't work since we're loading the default from the database as a fall-back, you'd want to only prompt if there isn't a start url or region provided as CLI args.

Also, `q login` is actually implemented here: https://github.com/aws/amazon-q-developer-cli-autocomplete/blob/main/crates/q_cli/src/cli/user.rs#L254

Sorry, this is pretty confusing with the current setup - we have `q` and `qchat` as separate binaries that currently support logging in, so we'd want to make the same changes across both.

rostcheck:
Hmm... I'm pretty sure it does work; I've actually been using the modified binary in production for the last week with my secure-agent install (https://github.com/rostcheck/secure-agent) to avoid it always asking for login. The workflow is "q login" then "q chat" (both started in a batch), "q login" saves the  login URL and region to the database, [crates/chat-cli/src/cli/user.rs](https://github.com/aws/amazon-q-developer-cli/pull/2660/commits/d45abbdb5174823ce8b27ec424cc0331d7e11e23#diff-876234879468a74de51218479c02e3cb0233857bde12725ce110430140546f85), right after my changed lines, and then  it's then picked up by "q chat's database lookup as you mentioned. Is there a different workflow I should also be accounting for? Thanks!

brandonskiser:
@rostcheck  the problem is I only want to allow skipping these auto-entered fields if they were passed as CLI args. They're prefilled when loaded from the database more as a convenience.

rostcheck:
Ok, I went through this and I think I've followed your comments and made the changes you requested: I changed the fix to only use the values for identity provider and region if they are provided on the CLI as arguments, otherwise prompt the user (with defaults from the DB). I then put in a [parallel issue #654 ](https://github.com/aws/amazon-q-developer-cli-autocomplete/issues/654) to amazon-q-developer-cli-autocomplete and a [PR #655 ](https://github.com/aws/amazon-q-developer-cli-autocomplete/pull/655) for it. Both were tested to validate that: "q logout" logs the user out; "q login" without arguments requires user data entry for the identity provider and region (pre-filling with the default values from the DB if available), "q login --identity-provider <provider> --region <region>" logs the user in directly (presenting the validation URL), and "q chat" either chats (if logged in) or asks the user to log in (if not logged in). If I misunderstood anything or there is more required, please let me know. Thanks!

PR #2680, opened 2025-08-23
Proposal: Add a `/reply` command

nikomatsakis:
*Description of changes:*

I'm not sure if there is better process for this, but this is a small feature I've been wanting! In addition to `/editor`, you can now do `/reply`. It begins the editor with the assistant's last comments, quoted with `>`. I find this very useful particularly when Claude has given me a list of clarifications and I want to answer all of them.

*Concerns:*

I couldn't find any existing infra to unit-test this with. I was thinking that it'd be nice to have a way to "mock out" the editor, but that doesn't seem to be a thing. Maybe I missed it?

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/mod.rs
+++ b/crates/chat-cli/src/cli/chat/cli/mod.rs
@@ -11,6 +11,7 @@ pub mod model;
 pub mod persist;
 pub mod profile;
 pub mod prompts;
+pub mod reply;
 pub mod subscribe;
 pub mod tangent;
 pub mod todos;
@@ -31,6 +32,7 @@ use model::ModelArgs;
 use persist::PersistSubcommand;
 use profile::AgentSubcommand;
 use prompts::PromptsArgs;
+use reply::ReplyArgs;
 use tangent::TangentArgs;
 use todos::TodoSubcommand;
 use tools::ToolsArgs;
@@ -71,6 +73,8 @@ pub enum SlashCommand {
     /// Open $EDITOR (defaults to vi) to compose a prompt
     #[command(name = "editor")]
     PromptEditor(EditorArgs),
+    /// Open $EDITOR with the most recent assistant message quoted for reply
+    Reply(ReplyArgs),
     /// Summarize the conversation to free up context space
     Compact(CompactArgs),
     /// View tools and permissions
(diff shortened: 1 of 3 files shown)


crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -740,11 +740,8 @@ impl ConversationState {
         self.append_transcript(format!("> {}", message.replace("\n", "> \n")));
     }
 
-    pub fn append_assistant_transcript(&mut self, message: &AssistantMessage) {
-        let tool_uses = message.tool_uses().map_or("none".to_string(), |tools| {
-            tools.iter().map(|tool| tool.name.clone()).collect::<Vec<_>>().join(",")
-        });
-        self.append_transcript(format!("{}\n[Tool uses: {tool_uses}]", message.content()));
+    pub fn append_assistant_transcript_for_reply(&mut self, message: &AssistantMessage) {
```
brandonskiser: Any reason why tool uses are being removed here? I would prefer to keep `append_assistant_transcript` the same

brandonskiser:
Thank you for the contribution! Had one question but otherwise looks good to me.

There unfortunately isn't any easy way to unit test this with the current code base so I wouldn't worry about it for now. I'm working on drafting something for a future v2 refactor that would otherwise alleviate this.

nikomatsakis:
I removed it because the goal is to answer questions the AI has posed to you and tool use is unlikely to be one of those questions. I can put it back if you prefer but I think for this use case it doesn’t seem useful. Is this method used elsewhere? My memory was that it was only used here but maybe that’s wrong, I could always add a Boolean parameter.

nikomatsakis:
@brandonskiser do you want me to make any further changes here?

brandonskiser:
@nikomatsakis apologies for the late reply - can you revert that change to `append_assistant_transcript` and keep it as it was before?

> Is this method used elsewhere? 

It's used for all new messages, the "tool uses" is just populated with "none" in those cases. Would rather keep with the existing paradigm at the moment since every command goes through this same flow.

For this case, it would always just populate tool uses with "none" but that seems simpler than creating a new method for this imo

nikomatsakis:
@brandonskiser done.

nikomatsakis:
@brandonskiser <https://github.com/brandonskiser> gentle ping =)

PR #2690, opened 2025-08-25
first round changes for agent generate for workshopping

xianwwu:
- **adding in agent contribution metric**
- **chore: fix lints**
- **agent generate where user is prompted to include or exclude entire MCP servers**
- **rebasing changes to newest api for conversation.rs**

*Issue #, if available:*

*Description of changes:*


crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -83,6 +93,23 @@ pub enum AgentSubcommand {
     Swap { name: Option<String> },
 }
 
+fn prompt_mcp_server_selection(servers: &[McpServerInfo]) -> Result<Vec<&McpServerInfo>, Box<dyn std::error::Error>> {
```
brandonskiser: use `eyre::Result` instead of returning a `Box<dyn Error>`

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -184,6 +211,47 @@ impl AgentSubcommand {
...
             },
+
+            Self::Generate { name, description } => {
+                use std::io::{self, Write};
+
+                let agent_name = match name {
+                    Some(n) => n,
+                    None => {
+                        print!("Enter agent name: ");
+                        io::stdout().flush()?;
```
brandonskiser: Write to `session.stderr` instead of directly to stdout

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -184,6 +211,47 @@ impl AgentSubcommand {
...
+                        let mut input = String::new();
+                        io::stdin().read_line(&mut input)?;
```
brandonskiser: Use `session.read_user_input` instead of reading a line directly from stdin here (same for all other instances)

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -184,6 +211,47 @@ impl AgentSubcommand {
...
+
+                let mcp_servers = ConversationState::get_enabled_mcp_servers(os).await.map_err(|e| ChatError::Custom(e.to_string().into()))?;
+
+                let selected_servers = if !mcp_servers.is_empty() {
+                    prompt_mcp_server_selection(&mcp_servers).map_err(|e| ChatError::Custom(e.to_string().into()))?
+                } else {
+                    Vec::new()
+                };
+
+                println!("Selected servers: {:?}", selected_servers);
```
brandonskiser: Remove these debug `print` statements

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -201,6 +219,76 @@ impl ConversationState {
         }
     }
 
+    /// Searches all configuration sources for MCP servers and returns a deduplicated list.
+    /// Priority order: Agent configs > Workspace legacy > Global legacy
+    pub async fn get_all_available_mcp_servers(os: &mut Os) -> Result<Vec<McpServerInfo>> {
```
brandonskiser: This shouldn't be in ConversationState, this can be moved to `profile.rs` for now

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -201,6 +219,76 @@ impl ConversationState {
...
+
+        // Convert to sorted vector
+        let mut result: Vec<McpServerInfo> = servers.into_values().collect();
+        result.sort_by(|a, b| a.name.cmp(&b.name));
+        
+        Ok(result)
+    }
+
+    /// Get only enabled MCP servers (excludes disabled ones)
+    pub async fn get_enabled_mcp_servers(os: &mut Os) -> Result<Vec<McpServerInfo>> {
```
brandonskiser: Move this to `profile.rs` as well

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1539,6 +1543,292 @@ impl ChatSession {
...
+        }
+
+        Ok(ChatState::PromptUser {
+            skip_printing_tools: true,
+        })
+    }
+
+    // Helper method to parse and validate the generated agent config
+    fn parse_and_validate_agent_config(
+        &self,
```
brandonskiser: this looks like `self` is unused? If so, this can be moved to a standalone function

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -184,6 +218,65 @@ impl AgentSubcommand {
...
+                    match session.read_user_input("Enter agent description: ", false) {
+                        Some(input) => input.trim().to_string(),
+                        None => {
+                            return Ok(ChatState::PromptUser {
+                                skip_printing_tools: true,
+                            });
+                        },
+                    }
+                } else {
+                    description.join(" ")
```
brandonskiser: Doesn't look like this else branch is necessary, just leave description empty

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -184,6 +218,65 @@ impl AgentSubcommand {
...
+                };
+
+                let mcp_servers = get_enabled_mcp_servers(os)
+                    .await
+                    .map_err(|e| ChatError::Custom(e.to_string().into()))?;
+
+                let selected_servers = if !mcp_servers.is_empty() {
+                    prompt_mcp_server_selection(&mcp_servers).map_err(|e| ChatError::Custom(e.to_string().into()))?
+                } else {
+                    Vec::new()
```
brandonskiser, line 257: This else branch also looks unnecessary, we'll get a new `Vec` returned from `prompt_mcp_server_selection`

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -311,3 +405,78 @@ fn highlight_json(output: &mut impl Write, json_str: &str) -> eyre::Result<()> {
...
+/// Priority order: Agent configs > Workspace legacy > Global legacy
+pub async fn get_all_available_mcp_servers(os: &mut Os) -> Result<Vec<McpServerInfo>> {
+    let mut servers = HashMap::<String, McpServerInfo>::new();
+    let mut seen_names = HashSet::<String>::new();
+    let mut seen_commands = HashSet::<String>::new();
+
+    // 1. Load from agent configurations (highest priority)
+    let mut stderr = std::io::stderr();
+    let (agents, _) = Agents::load(os, None, true, &mut stderr, true).await;
+    print!("Loaded {} agents from agent configurations", agents.agents.len());
```
brandonskiser: Should print to `session.stderr` for any user facing input

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1539,6 +1543,270 @@ impl ChatSession {
...
+            self.stderr,
+            style::SetForegroundColor(Color::Green),
+            style::Print(format!("✓ Generated agent config for '{}':\n\n", agent_name)),
+            style::SetForegroundColor(Color::Reset)
+        )?;
+
+        let formatted_json = serde_json::to_string_pretty(&initial_agent_config)
+            .map_err(|e| ChatError::Custom(format!("Failed to format JSON: {}", e).into()))?;
+
+        execute!(
```
brandonskiser: Can we instead prompt the user for editing the agent instead of opening editor for them? This should only be done if EDITOR is set


Also - open editor is already a general function, we can use the current implementation instead https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/cli/editor.rs#L86
(just make this `pub`)

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1539,6 +1543,270 @@ impl ChatSession {
         }
     }
 
+    /// Generates a custom agent configuration (system prompt and tool config) based on user input.
+    /// Uses an LLM to create the agent specifications from the provided name and description.
+    async fn generate_agent_config(
```
brandonskiser, line 1552: It looks like this implementation will always save the agent to the local agents directory. Can we instead pass an optional path parameter? Similar to `/agent create`

`/agent create` also makes use of this function for saving the config which can be reused here - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/agent/root_command_args.rs#L295

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3408,3 +3676,39 @@ mod tests {
         }
     }
 }
+
+// Helper method to parse and validate the generated agent config
+fn parse_and_validate_agent_config(response: &str, schema: &str) -> Result<Agent, String> {
```
brandonskiser: Can we reuse `Agent::load` instead?

crates/chat-cli/src/cli/chat/cli/profile.rs
xianwwu: I tried to do as you said, but then I ran into an error in the case where user has no discoverable MCP servers. the multiselect in prompt_mcp_server_selection needs to have some servers....

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1539,6 +1544,241 @@ impl ChatSession {
...
+            style::SetForegroundColor(Color::Reset)
+        )?;
+
+        let formatted_json = serde_json::to_string_pretty(&initial_agent_config)
+            .map_err(|e| ChatError::Custom(format!("Failed to format JSON: {}", e).into()))?;
+
+        let edited_content = open_editor(Some(formatted_json))?;
+
+        // Parse and validate the edited config
+        let final_agent_config = match serde_json::from_str::<Agent>(&edited_content) {
```
brandonskiser, line 1744: followup - let the user address malformed config without immediately failing. It's pretty likely that users are going to run into scenarios where they mess up editing somehow, and then have to go through the regeneration process again just to get their generated agent back

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -311,3 +419,77 @@ fn highlight_json(output: &mut impl Write, json_str: &str) -> eyre::Result<()> {
...
+    let mut seen_names = HashSet::<String>::new();
+    let mut seen_commands = HashSet::<String>::new();
+
+    // 1. Load from agent configurations (highest priority)
+    let mut null_writer = NullWriter;
+    let (agents, _) = Agents::load(os, None, true, &mut null_writer, true).await;
+
+    for (_, agent) in agents.agents {
+        for (server_name, server_config) in agent.mcp_servers.mcp_servers {
+            if !seen_names.contains(&server_name) && !seen_commands.contains(&server_config.command) {
```
dingfeli: So you are deduping on both how the server is named and the starting command? Why not just the starting command? Also what happens if the rest of the server config differs? e.g. different env list, different args, etc.

crates/chat-cli/src/cli/chat/cli/profile.rs
```diff
@@ -311,3 +419,77 @@ fn highlight_json(output: &mut impl Write, json_str: &str) -> eyre::Result<()> {
 
     Ok(execute!(output, style::ResetColor)?)
 }
+
+/// Searches all configuration sources for MCP servers and returns a deduplicated list.
+/// Priority order: Agent configs > Workspace legacy > Global legacy
+pub async fn get_all_available_mcp_servers(os: &mut Os) -> Result<Vec<McpServerInfo>> {
+    let mut servers = HashMap::<String, McpServerInfo>::new();
+    let mut seen_names = HashSet::<String>::new();
+    let mut seen_commands = HashSet::<String>::new();
```
dingfeli: I feel like we can perhaps get away with just using the servers HashMap here.

crates/chat-cli/src/cli/chat/cli/profile.rs
dingfeli: I think it's because `prompt_mcp_server_selection` needs to return an array regardless. Right now if you pass in an empty list of servers it just errors out because no selection can be made.

PR #2699, opened 2025-08-26
Add telemetry support for agent contribution tracking

kkashilk:
- Refactor send_cw_telemetry to handle multiple event types
- Add AgentContribution event handling that sends ChatInteractWithMessageEvent
- Track accepted line count from agent contributions as AgenticCodeAccepted interaction type

### DIFF
--- a/crates/chat-cli/src/telemetry/mod.rs
+++ b/crates/chat-cli/src/telemetry/mod.rs
@@ -14,6 +14,8 @@ use std::str::FromStr;
 
 use amzn_codewhisperer_client::types::{
     ChatAddMessageEvent,
+    ChatInteractWithMessageEvent,
+    ChatMessageInteractionType,
     IdeCategory,
     OperatingSystem,
     TelemetryEvent,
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/telemetry/mod.rs
```diff
@@ -557,55 +559,95 @@ impl TelemetryClient {
...
+                };
+
+                let event = TelemetryEvent::ChatAddMessageEvent(chat_add_message_event);
+                debug!(
+                    ?event,
+                    ?user_context,
+                    telemetry_enabled = self.telemetry_enabled,
+                    "Sending cw telemetry event"
+                );
+                if let Err(err) = codewhisperer_client
```
brandonskiser, line 605: very minor nit - return the `codewhisperer_client.send_telemetry_event` args in the match arms and have a single code path that invokes the `send_telemetry_event` func, just to keep the error printing from being duplicated

brandonskiser (approved):

kensave (approved):

crates/chat-cli/src/telemetry/mod.rs
kkashilk: Yeah, I thought about it, but will pick up some of this refactoring when we add a few more events that get published to CW Telemetry

PR #2700, opened 2025-08-26
chore(mcp): migrate to rmcp

dingfeli:
*Issue #, if available:*

*Description of changes:*
This PR migrates us to use the RMCP crate by anthropic. 
More specifically, this means the following changes:
- Removal of our own implementation of the MCP client
- Dependency of the RMCP crate
- Implementation of `Service<R>` trait from RMCP. 
- The spawning of child processes associated with a MCP is now done in the background. This is because RMCP leverages `TokioChildProcess` that does extra work which does not allow it to resolve immediately (and more importantly it is async which provided opportunity for racing conditions). This delay altered the control flow. Hence the move. Instead of storing initialized clients (which was what we had done previously) we store an enum of `Pending(JoinHandle)` and `Ready(InitializedClient)` (akin to future's Either). The state of this enum is flipped lazily (though because `JoinHandle` is spawned from a task it is polled actively in the background) when the server needs to be accessed. 

This change will serve as a foundation for support of different transport. We are incentivized to migrate because RMCP already supports all three transports (stdio, streamable http, and sse).  


crates/chat-cli/src/cli/chat/cli/prompts.rs
```diff
@@ -273,39 +285,12 @@ impl PromptsSubcommand {
                 });
             },
         };
-        if let Some(err) = prompts.error {
-            // If we are running into error we should just display the error
-            // and abort.
-            let to_display = serde_json::json!(err);
```
kensave, line 279: Interesting, why this was necessary, any why it's not needed now? Do we rely on the crate validations now?

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -280,23 +285,53 @@ impl ConversationState {
 
     /// Appends a collection prompts into history and returns the last message in the collection.
     /// It asserts that the collection ends with a prompt that assumes the role of user.
-    pub fn append_prompts(&mut self, mut prompts: VecDeque<Prompt>) -> Option<String> {
+    pub fn append_prompts(&mut self, mut prompts: VecDeque<PromptMessage>) -> Option<String> {
+        fn stringify_prompt_message_content(prompt_msg_content: PromptMessageContent) -> String {
+            match prompt_msg_content {
+                PromptMessageContent::Text { text } => text,
+                PromptMessageContent::Image { image } => {
+                    let length = image.raw.data.len();
```
kensave: Whats the behavior here? can an image be sent? Or we block? Otherwise it would seem we processed but didn't

crates/chat-cli/src/cli/chat/cli/prompts.rs
dingfeli: Yea. Previously the api I had extended from the client returns a lower level jrpc response object. Here the api called already nicely wrapped it in a Result<TypedResponse, Error>.

crates/chat-cli/src/cli/chat/tools/custom_tool.rs
```diff
@@ -64,175 +56,49 @@ pub fn default_timeout() -> u64 {
...
-            Ok(mut de_result) => {
-                for content in &mut de_result.content {
-                    if let MessageContent::Image { data, .. } = content {
-                        *data = format!("Redacted base64 encoded string of an image of size {}", data.len());
-                    }
+        let mut resp = self.client.call_tool(params).await?;
+
+        if resp.is_error.is_none_or(|v| !v) {
+            for content in &mut resp.content {
+                if let RawContent::Image(content) = &mut content.raw {
```
brandonskiser: Why do we redact image content?

crates/chat-cli/src/cli/chat/tools/custom_tool.rs
```diff
@@ -64,175 +56,49 @@ pub fn default_timeout() -> u64 {
...
 
-        match serde_json::from_value::<ToolCallResult>(result.clone()) {
-            Ok(mut de_result) => {
-                for content in &mut de_result.content {
-                    if let MessageContent::Image { data, .. } = content {
-                        *data = format!("Redacted base64 encoded string of an image of size {}", data.len());
-                    }
+        let mut resp = self.client.call_tool(params).await?;
+
+        if resp.is_error.is_none_or(|v| !v) {
```
brandonskiser, line 82: We don't log the source error? Or is a source error message not available?

crates/chat-cli/src/cli/chat/server_messenger.rs
```diff
@@ -1,48 +1,54 @@
...
     MessengerError,
-    PromptsListResult,
-    ResourceTemplatesListResult,
-    ResourcesListResult,
-    ToolsListResult,
+    MessengerResult,
+    Result,
 };
 
 #[allow(dead_code)]
```
brandonskiser, line 24: still required?

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -1,1140 +1,443 @@
...
-            Ok(result) => result,
-            Err(e) => {
-                let msg = format!("Failed to deserialize tool result from {}: {:?}", client.server_name, e);
-                break 'tool_list_result Err(eyre::eyre!(msg));
-            },
-        };
-        Ok::<ToolsListResult, eyre::Report>(tool_list_result)
-    };
+#[derive(Debug)]
+pub enum InitializedMcpClient {
```
brandonskiser, line 393: Why does `InitializedMcpClient` have a pending (ie, uninitialized client) as a valid value? Shouldn't these be split into separate types?

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -1,1140 +1,443 @@
...
-impl From<(tokio::time::error::Elapsed, String)> for ClientError {
-    fn from((error, context): (tokio::time::error::Elapsed, String)) -> Self {
-        ClientError::RuntimeError { source: error, context }
-    }
-}
+pub type RunningService = rmcp::service::RunningService<RoleClient, UninitMcpClient>;
 
 #[derive(Debug)]
-pub struct Client<T: Transport> {
+pub struct UninitMcpClient {
```
brandonskiser: Doc comments would help explain the role these structs play

crates/chat-cli/src/mcp_client/client.rs
```diff
@@ -1,1140 +1,443 @@
...
-    // Check the jrpc version.
-    // Currently we are only proceeding if the versions are EXACTLY the same.
-    let jrpc_version = ser_cap.jsonrpc.as_u32_vec();
-    let client_jrpc_version = JsonRpcVersion::default().as_u32_vec();
-    for (sv, cv) in jrpc_version.iter().zip(client_jrpc_version.iter()) {
-        if sv != cv {
-            return Err(ClientError::NegotiationError(
-                "Incompatible jrpc version between server and client".to_owned(),
-            ));
+impl Service<RoleClient> for UninitMcpClient {
```
brandonskiser: We're implementing a service trait for handling notifications from the MCP client on `UninitMcpClient`? Seems pretty unintuitive

brandonskiser (approved):

crates/chat-cli/src/mcp_client/client.rs
dingfeli: I struggled with the naming here. It's technically just saying "this contains a command that is run, it may or may not be ready".

crates/chat-cli/src/cli/chat/tools/custom_tool.rs
dingfeli: This was the case historically. I can enable it have it sent to the model as base64 encoded string but it will blow up the context in most cases.

crates/chat-cli/src/cli/chat/tools/custom_tool.rs
dingfeli: The resp is the error in this case. is_error is a field that signals whether or not the resp here is an error.

crates/chat-cli/src/cli/chat/conversation.rs
dingfeli: it has been blocked historically. My experience is that base64 encoded images are often too big and blows up the context window.

crates/chat-cli/src/cli/chat/server_messenger.rs
dingfeli: Yea there are variants in this that we are not using atm

crates/chat-cli/src/cli/chat/conversation.rs
dingfeli: Although on second thought there is really no harm in trying to send this. I have corrected this to actually send the data.

PR #2711, opened 2025-08-28
feat: add /experiment slash command for toggling experimental features

kensave:
- Add new /experiment command
- Toggle selection interface with colored On/Off status indicators
- Include experiment descriptions and disclaimer about experimental features

image

*Description of changes:*

image

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/experiment.rs
+++ b/crates/chat-cli/src/cli/chat/cli/experiment.rs
@@ -61,9 +61,9 @@ async fn select_experiment(os: &mut Os, session: &mut ChatSession) -> Result<Opt
 
         current_states.push(is_enabled);
         // Create clean single-line format: "Knowledge    [ON]   - Description"
-        let status_indicator = if is_enabled { 
+        let status_indicator = if is_enabled {
             style::Stylize::green("[ON] ")
-        } else { 
+        } else {
             style::Stylize::grey("[OFF]")
         };
         let label = format!(
--- a/crates/chat-cli/src/cli/chat/cli/experiment.rs
+++ b/crates/chat-cli/src/cli/chat/cli/experiment.rs
@@ -60,8 +60,12 @@ async fn select_experiment(os: &mut Os, session: &mut ChatSession) -> Result<Opt
         let is_enabled = os.database.settings.get_bool(experiment.setting_key).unwrap_or(false);
 
         current_states.push(is_enabled);
-        // Create clean single-line format: "Knowledge    [✅ ON]   - Description"
-        let status_indicator = if is_enabled { "[✅ ON] " } else { "[❌ OFF]" };
+        // Create clean single-line format: "Knowledge    [ON]   - Description"
+        let status_indicator = if is_enabled { 
+            style::Stylize::green("[ON] ")
+        } else { 
+            style::Stylize::grey("[OFF]")
+        };
         let label = format!(
             "{:<18} {} - {}",
             experiment.name,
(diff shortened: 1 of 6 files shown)


kensave:
We to consider tools spec to. We either ask user to restart or handle it.

crates/chat-cli/src/cli/chat/cli/experiment.rs
```diff
@@ -0,0 +1,151 @@
...
+        name: "Knowledge",
+        description: "Enables persistent context storage and retrieval across chat sessions (/knowledge)",
+        setting_key: Setting::EnabledKnowledge,
+    },
+    Experiment {
+        name: "Thinking",
+        description: "Enables complex reasoning with step-by-step thought processes",
+        setting_key: Setting::EnabledThinking,
+    },
+];
```
kkashilk, line 42: Is there a way to tie the `experimental` status of a setting directly in the settings module, instead of maintaining a separate list here?

crates/chat-cli/src/cli/chat/cli/experiment.rs
```diff
@@ -0,0 +1,151 @@
...
+    let mut experiment_labels = Vec::new();
+    let mut current_states = Vec::new();
+
+    for experiment in AVAILABLE_EXPERIMENTS {
+        let is_enabled = os.database.settings.get_bool(experiment.setting_key).unwrap_or(false);
+
+        current_states.push(is_enabled);
+        // Create clean single-line format: "Knowledge    [✅ ON]   - Description"
+        let status_indicator = if is_enabled { "[✅ ON] " } else { "[❌ OFF]" };
+        let label = format!(
```
kkashilk: instead of the emojis, can we use color coded `ON` and `OFF` to show the status. 
Or use smaller emojis?

crates/chat-cli/src/cli/chat/cli/experiment.rs
```diff
@@ -0,0 +1,151 @@
+// ABOUTME: Implements the /experiment slash command for toggling experimental features
```
brandonskiser, line 1: Use `//!` for module-level doc comments

crates/chat-cli/src/cli/chat/cli/experiment.rs
brandonskiser: `/knowledge` is currently using these same emojis in a couple lines, so we could either change all of them or just address it later

crates/chat-cli/src/cli/chat/cli/experiment.rs
brandonskiser: `Settings` could return a list of these experiment structs pretty easily

Although we're already pretty coupled to manually-written documentation already about these experimental features so don't think it's too terrible to just have it defined here.

brandonskiser (approved):

crates/chat-cli/src/cli/chat/cli/experiment.rs
kensave: I felt its easier for development, we can evolve it if it becomes hard to maintain.

crates/chat-cli/src/cli/chat/cli/experiment.rs
kensave: This could be UX rework by itself.

kensave: Lowered the tone.

evanliu048:
There are some minor comments and will be solved in a separate pr

PR #2726, opened 2025-08-28
fix: experiment selection bounds

kensave:
*Issue #, if available:*

Layoud issue in /experiment

image


brandonskiser (approved):

abhraina-aws (changes requested):
Good catch.

PR #2755, opened 2025-09-02
Make /quit command alias /exit visible

sebglazebrook:
*Issue #, if available:*

N/A

*Description of changes:*

When using Amazon Q I've found myself trying to exit the chat via the command `/exit` and then seeing no autocompletion so I search for help and see that the command is actually `/quit`. 

I wanted to add new functionality to create `/exit` as an alias for `/quit` but it turns out the alias already existed but was not visible to the user. I've made it visible both in the help and autocompletion so it's easily discoverable by users. 

NOTE: I didn't add any changes for the other alias `q` as autocompletion of the letter `q` will automatically show `quit` it didn't need necessary.

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/mod.rs
+++ b/crates/chat-cli/src/cli/chat/cli/mod.rs
@@ -32,6 +32,7 @@ use prompts::PromptsArgs;
 use tangent::TangentArgs;
 use todos::TodoSubcommand;
 use tools::ToolsArgs;
+use color_print::cstr;
 
 use crate::cli::chat::cli::subscribe::SubscribeArgs;
 use crate::cli::chat::cli::usage::UsageArgs;
@@ -49,8 +50,8 @@ use crate::os::Os;
 #[derive(Debug, PartialEq, Parser)]
 #[command(color = clap::ColorChoice::Always, term_width = 0, after_long_help = EXTRA_HELP)]
 pub enum SlashCommand {
-    /// Quit the application
-    #[command(aliases = ["q", "exit"])]
+    /// Quit the application (alias /exit)
+    #[command(visible_aliases = ["q", "exit"], override_usage = cstr!("<bold>/quit</bold> or <bold>/exit</bold>"))]
     Quit,
     /// Clear the conversation history
     Clear(ClearArgs),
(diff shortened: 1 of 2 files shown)


crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -530,6 +531,54 @@ mod tests {
         assert!(completions.is_empty());
     }
 
+    #[test]
+    fn test_chat_completer_quit_completion() {
```
brandonskiser: We seem to already have other tests here for completer - instead of adding a ton of new lines just for testing this alias, I'd prefer just not having a test for this (considering we don't have tests for all of the others), or creating an actual parameterized test instead

crates/chat-cli/src/cli/chat/cli/mod.rs
```diff
@@ -49,8 +50,8 @@ use crate::os::Os;
 #[derive(Debug, PartialEq, Parser)]
 #[command(color = clap::ColorChoice::Always, term_width = 0, after_long_help = EXTRA_HELP)]
 pub enum SlashCommand {
-    /// Quit the application
-    #[command(aliases = ["q", "exit"])]
+    /// Quit the application (alias /exit)
+    #[command(visible_aliases = ["q", "exit"], override_usage = cstr!("<bold>/quit</bold> or <bold>/exit</bold>"))]
```
brandonskiser, line 54: Just `#[command(visible_aliases = ["q", "exit"])` should be sufficient here, no need to change override_usage or the doc comment.

crates/chat-cli/src/cli/chat/prompt.rs
sebglazebrook: I've updated the tests to iterate through some example inputs and check for expected outputs. I can remove it if you prefer or add a more comprehensive list. I also look into different ways to parameterize test in Rust but that would require a crate which I didn't want to add unless you're keen for it.

crates/chat-cli/src/cli/chat/cli/mod.rs
sebglazebrook: I can remove the override_usage but I think the updated comment really helps to highlight that the alias exists. I'm comparing this to claude code where they reference the alias in their autocomplete command descriptions. 

But happy to remove if you prefer.

PR #2769, opened 2025-09-03
feat: implement persistent CLI history with file storage

kkashilk:
- Add Drop trait to InputSource for automatic history saving
- Replace DefaultHistory with FileHistory for persistence
- Store history in ~/.aws/amazonq/cli_history
- Refactor ChatHinter to use rustyline's built-in history search
- Remove manual history tracking in favor of rustyline's implementation
- Add history loading on startup with error handling
- Clean up unused hinter history update methods


kkashilk:
> This is great! Can you please confirm the behavior in the PR description?
> 
>     1. This is just as a back up when closing and starting cli?

We load/save history when starting/ending a session from a file. But while a session is ongoing, we will continue appending to an in-memory history list.

>     2. Is the whole conversation history thats getting transferred or is this just execute bash command history what is the history needed for?

This history is a list of all messages sent by the user. We already have this feature, we just don't persist it across sessions rn.

>     3. Or are we keeping the past discussion separate from current session history but giving access to the history nevertheless?
Not sure I follow, having this history will allow users to use `up/down` arrow keys and ctrl+r -> reverse search on past messages/commands that they have sent

crates/chat-cli/src/cli/chat/input_source.rs
```diff
@@ -97,6 +117,18 @@ impl InputSource {
         }
     }
 
+    fn should_append_history(line: &str) -> bool {
+        let trimmed = line.trim();
```
brandonskiser: Can we also convert to lowercase here, since `"Y"` is also a valid input for approving tool uses

brandonskiser (approved):

crates/chat-cli/src/cli/chat/input_source.rs
kkashilk: updated

PR #2811, opened 2025-09-08
fix: update dangerous patterns for execute bash to include $

brandonskiser:
*Description of changes:*
- Updating the `execute_bash` dangerous patterns list to include any references to `$`. Added a test as to why this is required.


brandonskiser:
`cargo-deny` is failing because bm25 depends on fxhash, which is apparently unmaintained - https://rustsec.org/advisories/RUSTSEC-2025-0057

Raised an issue to see if a fix can be applied in bm25 - https://github.com/Michael-JB/bm25/issues/25

PR #2815, opened 2025-09-08
feat: add model field support to custom agent format

mr-lee:
- Add optional 'model' field to Agent struct for specifying model per agent
- Update JSON schema and documentation with model field usage
- Integrate agent model into model selection priority:
  1. CLI argument (--model)
  2. Agent's model field (new)
  3. User's saved default model
  4. System default model
- Add proper fallback when agent specifies unavailable model
- Include comprehensive unit tests for model field functionality
- Maintain backward compatibility with existing agent configurations

*Description of changes:*

The current `agent-format.md` is missing a "model" argument to enable a user to set which model to use when creating a custom agent.  This fixes that gap.

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -161,6 +161,9 @@ pub struct Agent {
     /// you configure in the mcpServers field in this config
     #[serde(default)]
     pub use_legacy_mcp_json: bool,
+    /// The model ID to use for this agent. If not specified, uses the default model.
+    #[serde(default)]
+    pub model: Option<String>,
     #[serde(skip)]
     pub path: Option<PathBuf>,
 }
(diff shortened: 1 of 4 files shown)


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -161,6 +161,9 @@ pub struct Agent {
     /// you configure in the mcpServers field in this config
     #[serde(default)]
     pub use_legacy_mcp_json: bool,
+    /// The model ID to use for this agent. If not specified, uses the default model.
+    #[serde(default)]
```
dingfeli, line 165: you probably don't need the annotation if the field is already typed with Option.

mr-lee: Oh good call, let me fix that

Issue #2818, opened 2025-09-09
Agent toolsSettings relative path patterns fail to normalize causing inconsistent security behavior


yayami3:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

Linux

### Expected behaviour

Agent toolsSettings with relative path patterns should consistently enforce access restrictions regardless of how paths are specified. Both `./denied/test.txt` and `denied/test.txt` should be treated identically since they refer to the same file location.

### Actual behaviour

The `canonicalizes_path()` function only performs environment variable expansion (`$HOME`, `~`) but doesn't normalize relative paths to absolute paths. This causes inconsistent pattern matching:

**Test Configuration:**
```json
{
 "name": "test-agent",
 "tools": ["*"],
 "toolsSettings": {
   "fs_write": {
     "deniedPaths": ["./denied/**"]
   }
 }
}
```

**Results:**
- Input: `./denied/test.txt` → **DENY** ✅ (correctly blocked)
- Input: `denied/test.txt` → **ASK** ❌ (should be blocked but isn't)

This creates a security inconsistency where the same file can be accessed or blocked depending on path format.

### Steps to reproduce

1. Create agent config with relative path restrictions:
  ``` json
  {
    "name": "test-agent",
    "tools": ["*"],
    "toolsSettings": {
      "fs_write": {
        "deniedPaths": ["./denied/**"]
      }
    }
  }
```

2. Test with explicit relative path:
  ``` bash
  q chat --agent test-agent
  # Try: fs_write tool with path "./denied/test.txt"
  # Result: Correctly denied
```

3. Test with implicit relative path:
  ``` bash
  q chat --agent test-agent
  # Try: fs_write tool with path "denied/test.txt"
  # Result: Asks for permission (should be denied)
```

### Root Cause

The canonicalizes_path() function in crates/chat-cli/src/util/
directories.rs
 only expands environment variables but doesn't resolve relative paths to absolute paths:

``` rust
pub fn canonicalizes_path(os: &Os, path_as_str: &str) -> Result<String> {
   let context = |input: &str| Ok(os.env.get(input).ok());
   let home_dir = || os.env.home().map(|p| p.to_string_lossy().to_string());

[...]

PR #2820, opened 2025-09-09
fix: normalize relative paths in agent toolsSettings

yayami3:
Fixes an issue where relative path patterns in agent toolsSettings (e.g., './denied/**') would not match files accessed with different relative path formats (e.g., 'denied/file.txt').

Changes:
- Enhanced canonicalizes_path() to properly normalize relative paths
- Added path component resolution for '.' and '..' segments
- Ensures consistent path matching for both fs_read and fs_write tools

This resolves a security issue where deniedPaths could be bypassed by using different relative path formats.

*Issue #2818

*Description of changes:*

### DIFF
--- a/crates/chat-cli/src/util/directories.rs
+++ b/crates/chat-cli/src/util/directories.rs
@@ -186,7 +186,7 @@ pub fn canonicalizes_path(os: &Os, path_as_str: &str) -> Result<String> {
     let home_dir = || os.env.home().map(|p| p.to_string_lossy().to_string());
 
     let expanded = shellexpand::full_with_context(path_as_str, home_dir, context)?;
-    
+
     // Convert all relative paths to absolute paths (including glob patterns)
     if !expanded.starts_with("/") && !expanded.starts_with("~") {
         let current_dir = os.env.current_dir()?;
@@ -198,7 +198,7 @@ pub fn canonicalizes_path(os: &Os, path_as_str: &str) -> Result<String> {
                 // If canonicalize fails (e.g., path doesn't exist), do manual normalization
                 let normalized = normalize_path(&absolute_path);
                 Ok(normalized.to_string_lossy().to_string())
-            }
+            },
         }
     } else {
         Ok(expanded.to_string())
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/util/directories.rs
```diff
@@ -185,7 +185,44 @@ pub fn canonicalizes_path(os: &Os, path_as_str: &str) -> Result<String> {
     let context = |input: &str| Ok(os.env.get(input).ok());
     let home_dir = || os.env.home().map(|p| p.to_string_lossy().to_string());
 
-    Ok(shellexpand::full_with_context(path_as_str, home_dir, context)?.to_string())
+    let expanded = shellexpand::full_with_context(path_as_str, home_dir, context)?;
+    
+    // Convert all relative paths to absolute paths (including glob patterns)
+    if !expanded.starts_with("/") && !expanded.starts_with("~") {
```
erbenmo, line 191: ~ is already expanded I think?

erbenmo:
I noticed a same issue in https://github.com/aws/amazon-q-developer-cli/pull/2926/files

My change only applies normalization to fs_read and only to the file being read. I think we should probably use your change since it covers fs_write as well as the allowed_path/denied_path pattern.

Can you add some testing in your change? Feel free to take mine

crates/chat-cli/src/util/directories.rs
erbenmo: Consider using Path::new(&expanded).is_relative

crates/chat-cli/src/util/directories.rs
```diff
@@ -185,7 +185,44 @@ pub fn canonicalizes_path(os: &Os, path_as_str: &str) -> Result<String> {
...
+        let current_dir = os.env.current_dir()?;
+        let absolute_path = current_dir.join(expanded.as_ref());
+        // Normalize the path to remove ./ and ../ components
+        match absolute_path.canonicalize() {
```
erbenmo, line 195: I am wondering should we just always use the normalize_path() method for code simplicity and consistency?

brandonskiser (approved):

kkashilk:
This doesn't currently handle this test - 
```
let result = canonicalizes_path(&test_os, "/absolute/./../path").unwrap();
assert_eq!(result, "/path");
```

PR #2824, opened 2025-09-09
Reduce default fs_read trust permission to current working directory only

erbenmo:
*Description of changes:*

    Reduce default fs_read trust permission to current working directory only
    
    Previously by default fs_read is trusted to read any file on user's file
    system. This PR reduces the fs_read permission to CWD only. This means user
    can still access any file under CWD without prompt. But if user needs to
    access file outside CWD, she will be prompted for explicit approval.
    
    User can still explicitly add fs_read to trusted tools in chat / agent
    definition so fs_read can read any file without prompt.
    
    This change essentially adds a layer of defense against prompt injection
    by following the least-privilege principle.

## Testing
1. launching Q CLI for CWD
```
~/Documents/Work/2025/Project/amazon-q-developer-cli $ cargo run --bin chat_cli
   Compiling chat_cli v1.15.0 (/Users/moerben/Documents/Work/2025/Project/amazon-q-developer-cli/crates/chat-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 13.57s
     Running `target/debug/chat_cli`
```

2. Reading file outside CWD
```
> read ~/.super-secret

> I'll read the ~/.super-secret file for you.


🛠️  Using tool: fs_read
 ⋮ 
 ● 
Allow this action? Use 't' to trust (always allow) this tool for the session. [y/n/t]:

> n

Reading file: ~/.super-secret, all lines> I understand you denied the file read request. Could you clarify what you'd like me to do instead? Are you:

• Looking to create a file at that path?
• Wanting to check if the file exists?
• Trying to read a different file?
• Testing how I handle file access requests?

What specific action would you like me to take?
```

3. Reading files under CWD
```
> read read crates/chat-cli/src/cli/chat/tools/fs_read.rs

🛠️  Using tool: fs_read (trusted)
 ⋮ 
 ● Reading file: /Users/moerben/Documents/Work/2025/Project/amazon-q-developer-cli/crates/chat-cli/src/cli/chat/tools/fs_read.rs, all lines
 ✓ Successfully read 56561 bytes from /Users/moerben/Documents/Work/2025/Project/amazon-q-developer-cli/crates/chat-cli/src/cli/chat/tools/fs_read.rs

```

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/fs_read.rs
+++ b/crates/chat-cli/src/cli/chat/tools/fs_read.rs
@@ -109,12 +109,8 @@ impl FsRead {
             allowed_paths: Vec<String>,
             #[serde(default)]
             denied_paths: Vec<String>,
-            #[serde(default = "default_allow_read_only")]
+            #[serde(default)]
             allow_read_only: bool,
-        }
-
-        fn default_allow_read_only() -> bool {
-            false
         }
 
         let is_in_allowlist = matches_any_pattern(&agent.allowed_tools, "fs_read");
--- a/crates/chat-cli/src/cli/chat/tools/fs_read.rs
+++ b/crates/chat-cli/src/cli/chat/tools/fs_read.rs
@@ -114,23 +114,31 @@ impl FsRead {
         }
 
         fn default_allow_read_only() -> bool {
-            true
+            false
         }
 
         let is_in_allowlist = matches_any_pattern(&agent.allowed_tools, "fs_read");
-        match agent.tools_settings.get("fs_read") {
-            Some(settings) => {
+        let settings = agent.tools_settings.get("fs_read").cloned()
+            .unwrap_or_else(|| serde_json::json!({}));
+        
+        {
                 let Settings {
-                    allowed_paths,
+                    mut allowed_paths,
                     denied_paths,
                     allow_read_only,
-                } = match serde_json::from_value::<Settings>(settings.clone()) {
+                } = match serde_json::from_value::<Settings>(settings) {
                     Ok(settings) => settings,
                     Err(e) => {
                         error!("Failed to deserialize tool settings for fs_read: {:?}", e);
                         return PermissionEvalResult::Ask;
                     },
                 };
+                
+                // Always add current working directory to allowed paths
+                if let Ok(cwd) = os.env.current_dir() {
+                    allowed_paths.push(cwd.to_string_lossy().to_string());
+                }
+                
                 let allow_set = {
                     let mut builder = GlobSetBuilder::new();
                     for path in &allowed_paths {
@@ -259,10 +267,7 @@ impl FsRead {
                         PermissionEvalResult::Ask
                     },
                 }
-            },
-            None if is_in_allowlist => PermissionEvalResult::Allow,
-            _ => PermissionEvalResult::Ask,
-        }
+            }
     }
 
     pub async fn invoke(&self, os: &Os, updates: &mut impl Write) -> Result<InvokeOutput> {
@@ -862,6 +867,7 @@ fn format_mode(mode: u32) -> [char; 9] {
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
+    use std::path::PathBuf;
 
     use super::*;
     use crate::cli::agent::ToolSettingTarget;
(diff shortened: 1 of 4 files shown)


crates/chat-cli/src/util/directories.rs
erbenmo, line 1: I actually have a separate PR for this change: https://github.com/aws/amazon-q-developer-cli/pull/2817/

But I guess since that PR isn't merged, this change is also showing here

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -259,10 +267,7 @@ impl FsRead {
                         PermissionEvalResult::Ask
                     },
                 }
-            },
-            None if is_in_allowlist => PermissionEvalResult::Allow,
-            _ => PermissionEvalResult::Ask,
-        }
+            }
     }
```
erbenmo, line 267: Note that I am also changing the behavior for the case that "fs_read" is not explicitly defined in agent.tools_settings to re-use the big chunk of code above as if the fs_read is an empty json (no allowed path, no denied path, etc)

evanliu048:
Thanks for contributing this! I have a few questions:

- What was the motivation behind this change? Have we seen user complaints about `fs_read` being trusted by default?

- If we restrict trust to only the current working directory, do we anticipate pushback from users who may get interrupted by prompts when reading files outside CWD? For example, users with automation workflows might find the additional confirmation disruptive.

- This feels a bit like a product decision as well—we might also want to loop in PM to discuss whether this is the right default experience

erbenmo:
> Thanks for contributing this! I have a few questions:
> 
> * What was the motivation behind this change? Have we seen user complaints about `fs_read` being trusted by default?
> * If we restrict trust to only the current working directory, do we anticipate pushback from users who may get interrupted by prompts when reading files outside CWD? For example, users with automation workflows might find the additional confirmation disruptive.
> * This feels a bit like a product decision as well—we might also want to loop in PM to discuss whether this is the right default experience


Hi @evanliu048 

- This idea is to follow the principles of least privileged - by default we only give CLI the permission to read CWD. If CLI becomes unaligned with user for whatever reason, and starts reading user's secret files outside CWD, the user gets notified and can prevent that from happening.
- Users who don't want to get interrupted can explicitly add the fs_read to allowed tools (either by "--trust-tools fs_read" or through agent configuration) to get the old behavior.
- "we might also want to loop in PM" - Yes let me know how to get that started. I believe this is the right thing to do. Previously I have chatted with Matt Lee and Matt Schrage about this change and they are onboarded. Talking to PM would also help!

h16rkim:
After reading about the recent [security incident in nx](https://www.stepsecurity.io/blog/supply-chain-security-alert-popular-nx-build-system-package-compromised-with-data-stealing-malware), I started to worry about the risks of prompt injection. Personally, I’ve configured my agent prompt not to read files outside the CWD directory. From a security perspective, I think this PR could be helpful.

On the other hand, it might also be a good idea to add an additional option in fs_read’s [toolSettings](https://github.com/aws/amazon-q-developer-cli/blob/46ddc72ac2025e007368d869005aa8d0f10192fb/docs/built-in-tools.md?plain=1#L41-L63) to “trust files under the current CWD.”

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -114,23 +114,31 @@ impl FsRead {
         }
 
         fn default_allow_read_only() -> bool {
-            true
+            false
         }
```
evanliu048: nit: we don’t need an extra method here — if we want the default to be `false`, that’s already the default for bool.

erbenmo: removed in latest rev @evanliu048

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -3,47 +3,18 @@ use std::fs::Metadata;
 use std::io::Write;
 
 use crossterm::queue;
```
mr-lee: nit - can you undo the formatting changes?

erbenmo: fixed now. sorry I was using Kiro IDE for a quick change and I didn't notice it autoformat the file.

Issue #2210, opened 2025-07-02
Q CLI no longer supports ctrl+f to move forward a character (standard readline functionality)


natsirtguy:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.5.0 (24F74)

### Expected behaviour

Pressing ctrl+f within `q chat` moves forward one character.

### Actual behaviour

Pressing ctrl-f within `q chat` does nothing.

### Steps to reproduce

Open a chat with `q chat`, type some characters, go to the beginning of the line with ctrl+a, press ctrl-f.

### Environment

```yaml
<This will be visible to anyone. Do not include personal or sensitive information>

[q-details]
version = "1.12.2"
hash = "b9100eb751546269f095585705a0e7fc3ea55419"
date = "2025-06-30T22:40:01.105251Z (2d ago)"

[system-info]
os = "macOS 15.5.0 (24F74)"
chip = "Apple M1 Pro"
total-cores = 10
memory = "32.00 GB"

[environment]
cwd = "/Users/USER"
cli-path = "/Users/USER"
install-method = "unknown"

[env-vars]
PATH = "/Users/USER/.nvm/versions/node/v22.14.0/bin:/Users/USER/.asdf/shims:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Users/USER/.local/bin:/Users/USER/.toolbox/bin:/Users/USER/.cargo/bin"
QTERM_SESSION_ID = "934d1ae167674bb080ccf2872b08ff03"
Q_SET_PARENT_CHECK = "1"
Q_TERM = "1.12.2"
SHELL = "/bin/zsh"
TERM = "xterm-256color"
__CFBundleIdentifier = "com.apple.Terminal"
```

konippi:
@brandonskiser 
[...]

PR #2825, opened 2025-09-09
Change autocomplete shortcut from ctrl-f to ctrl-g

erbenmo:
*Description of changes:*

The reason is ctrl-f is the standard shortcut in UNIX for moving cursor forward by 1 character. You can find it being supported everywhere... in your browser, your terminal, etc.

## Test

Manually tested ctrl-g and ctrl-f in the CLI

Also tested overriding the shortcut through config:
```
~/Documents/Work/2025/Project/amazon-q-developer-cli/crates/chat-cli $ cargo run --bin chat_cli -- settings chat.autocompletionKey o
   Compiling chat_cli v1.15.0 (/Users/moerben/Documents/Work/2025/Project/amazon-q-developer-cli/crates/chat-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 11.86s
     Running `/Users/moerben/Documents/Work/2025/Project/amazon-q-developer-cli/target/debug/chat_cli settings chat.autocompletionKey o`
```

verified ctrl-o works in the above example ^^

### DIFF
--- a/crates/chat-cli/src/cli/chat/prompt.rs
+++ b/crates/chat-cli/src/cli/chat/prompt.rs
@@ -479,9 +479,13 @@ pub fn rl(
         EventHandler::Simple(Cmd::Insert(1, "\n".to_string())),
     );
 
-    // Add custom keybinding for Ctrl+g to accept hint (like fish shell)
+    // Add custom keybinding for autocompletion hint acceptance (configurable)
+    let autocompletion_key_char = match os.database.settings.get_string(Setting::AutocompletionKey) {
+        Some(key) if key.len() == 1 => key.chars().next().unwrap_or('g'),
+        _ => 'g', // Default to 'g' if setting is missing or invalid
+    };
     rl.bind_sequence(
-        KeyEvent(KeyCode::Char('g'), Modifiers::CTRL),
+        KeyEvent(KeyCode::Char(autocompletion_key_char), Modifiers::CTRL),
         EventHandler::Simple(Cmd::CompleteHint),
     );
 
--- a/crates/chat-cli/src/cli/chat/prompt.rs
+++ b/crates/chat-cli/src/cli/chat/prompt.rs
@@ -473,19 +473,19 @@ pub fn rl(
         EventHandler::Simple(Cmd::Insert(1, "\n".to_string())),
     );
 
-    // Add custom keybinding for Ctrl+J to insert a newline
+    // Add custom keybinding for Ctrl+j to insert a newline
     rl.bind_sequence(
         KeyEvent(KeyCode::Char('j'), Modifiers::CTRL),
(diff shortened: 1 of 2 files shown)


crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -473,19 +473,19 @@ pub fn rl(
...
     rl.bind_sequence(
         KeyEvent(KeyCode::Char('j'), Modifiers::CTRL),
         EventHandler::Simple(Cmd::Insert(1, "\n".to_string())),
     );
 
-    // Add custom keybinding for Ctrl+F to accept hint (like fish shell)
+    // Add custom keybinding for Ctrl+g to accept hint (like fish shell)
     rl.bind_sequence(
-        KeyEvent(KeyCode::Char('f'), Modifiers::CTRL),
+        KeyEvent(KeyCode::Char('g'), Modifiers::CTRL),
```
abhraina-aws: We have precedent to make shortcuts customizable. We should probably follow suit there.

erbenmo: Yeah that feels like a separate feature request. Looking at the "TangentModeKey" code below we already have a mechanism to configure the key. My goal in this PR is to just make ctrl-f available.

abhraina-aws: Yeah we can just copy the mechanism there and the whole debate about the right key ends. We had the same discussion during tangent mode.

erbenmo: I think the problem of setting up a good default value will not go away though, even when user can configure everything in the future. I am trying to solve the default value problem here.

mr-lee: https://fishshell.com/docs/current/interactive.html

Arguably the shortcut comment here is no longer accurate, as a `fish` user would expect Ctrl+f.  Rather than just change the default key, I think we may want to just spend the time to build the config feature.

If we're strongly set on changing the key, we should remove that comment on L482, as it's not accurate.

erbenmo: In the latest PR I am allowing user to configure the key for autocompletion through:

```
cargo run --bin chat_cli -- settings chat.autocompletionKey o
```

This is following the existing config system built for tangent mode, fuzzy search, etc.

I am also changing the default from f to g

natsirtguy: I don't think the issue is that Ctrl+f is a bad default, the issue is that Ctrl-f incorrectly attempts to autocomplete when an autocomplete suggestion isn't available. fish doesn't shadow the ability to move forward under normal circumstances. So the fundamental issue is a bug in the way Ctrl+f autocomplete behavior is implemented, not the default key.

natsirtguy:
Resolves #2210 (kind of)

PR #2828, opened 2025-09-10
feat(use_aws): add configurable autoAllowReadonly setting

mr-lee:
- Add auto_allow_readonly field to use_aws Settings struct (defaults to false)
- Update eval_perm method to use auto_allow_readonly setting instead of hardcoded behavior
- Default behavior: all AWS operations require user confirmation (secure by default)
- Opt-in behavior: when autoAllowReadonly=true, read-only operations are auto-approved
- Add comprehensive tests covering all scenarios

*Description of changes:*


mr-lee:
> This looks good to me. But I think recently IDE has also aligned with CLI in terms of config. Does this change anything there?

I don't think so - the tool implementation and config is already slightly different - at some point we may decide to align those, but that doesn't block us adding additional options to the spec.

Issue #1917, opened 2025-05-20
Q CLI should auto-trust chained/piped read-only commands


blakehurd:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.4.1 (24E263)

### Expected behaviour

<This will be visible to anyone. Do not include personal or sensitive information>

Q CLI should automatically trust chained or piped commands that only perform read operations without requiring explicit approval each time. The trust system should analyze the entire command chain to determine if it's read-only.

### Actual behaviour

<This will be visible to anyone. Do not include personal or sensitive information>

Currently, Q CLI requires explicit approval for chained or piped read-only commands (like "cat file.txt | grep pattern" or "echo file.txt | xargs grep pattern"), even though these operations don't modify any files.

```
[chat-transcript]
You're welcome! I'm glad I could help you report the issue with the Q CLI trusted commands configuration. The development team should now be aware of the problem and will work on addressing it.

If you have any other questions about Amazon Q CLI or need assistance with anything else, feel free to ask anytime.
[Too

> Is xargs grep readonly?

The `xargs grep` command itself is not inherently read-only. It's a combination of two commands:

1. `xargs` - Takes input from standard input and executes a command using that input as arguments
2. `grep` - Searches for patterns in files

When used together as `xargs grep`, the command is typically used to search for patterns across multiple files, but it doesn't modify the files by default. It only reads their contents to find matching patterns.

However, it's important to note that while the basic usage is read-only (just searching and displaying results), you could potentially use `xargs` with other commands that do modify files. For example, if you piped the output to something that writes to files, it would no longer be read-only.

In the context of our earlier discussion about trusted commands in Amazon Q CLI, adding `xargs` and `grep` to the trusted commands list should be safe from a read-only perspective when used in their standard pattern-matching capacity.
[Tool uses: none]

> Can you please report an issue about this not being trusted by default?

I'll help you report an issue about `xargs grep` not being trusted by default in the Amazon Q CLI.
[Tool uses: report_issue]





I've submitted a new issue report about `xargs grep` not being trusted by default in the Q CLI. The report includes:

- Title: xargs grep should be trusted by default in Q CLI
[...]

PR #2829, opened 2025-09-10
feat: Allow readonly `find -exec` commands without permission prompt

h16rkim:
## Summary

When using the Amazon Q Developer CLI, the LLM Agent often execute readonly file exploration commands such as:

```bash
find . -name '*.rs' -exec grep -l "ExecuteCommand" {} \;
```

Even with the agent config `toolSettings.execute_bash.allowReadonly = true`, the CLI still asks user for permission when executing this type of command. ([related documents](https://github.com/aws/amazon-q-developer-cli/blob/46ddc72ac2025e007368d869005aa8d0f10192fb/docs/built-in-tools.md?plain=1#L27))

## Root Cause

1. Any command using `find + -exec` is currently treated as permission-required commands.
2. Commands containing the character `;` are treated as dangerous and require permission for security reasons.

## Changes

1. Readonly `find -exec` handling
* Modified logic so that when using `find` + `-exec`, if the command following exec is readonly, no permission prompt is required.
* This also works correctly when multiple -exec clauses are present.

2. Semicolon (`;`) handling
* Updated logic so that `;` is only considered dangerous when it appears in the middle of a command.
* If `;` only appears at the end, it is no longer treated as a dangerous command.

## Related Issue

https://github.com/aws/amazon-q-developer-cli/issues/1917

## Review Notes

* Please review whether this change introduces any potential security concerns.
* The intent is to reduce unnecessary permission prompts while preserving security guarantees.

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
+++ b/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
@@ -72,7 +72,10 @@ impl ExecuteCommand {
         };
         const DANGEROUS_PATTERNS: &[&str] = &["<(", "$(", "`", ">", "&&", "||", "&", ";", "${", "\n", "\r", "IFS"];
 
-        if DANGEROUS_PATTERNS.iter().any(|p| contains_in_middle(&self.command, p)) {
+        if DANGEROUS_PATTERNS
+            .iter()
+            .any(|p| contains_but_not_ends_with(&self.command, p))
+        {
             return true;
         }
 
@@ -104,15 +107,15 @@ impl ExecuteCommand {
                 // Special casing for `find` so that we support most cases while safeguarding
                 // against unwanted mutations
                 Some(cmd)
-                if cmd == "find"
-                    && cmd_args.iter().any(|arg| {
-                    arg.contains("-delete")
+                    if cmd == "find"
+                        && cmd_args.iter().any(|arg| {
+                            arg.contains("-delete")
                         || arg.contains("-ok") // includes -okdir
                         || arg.contains("-fprint") // includes -fprint0 and -fprintf
-                }) =>
-                    {
-                        return true;
-                    },
+                        }) =>
+                {
+                    return true;
+                },
                 // Check -exec commands separately to allow readonly commands
                 Some(cmd) if cmd == "find" && cmd_args.iter().any(|arg| arg.contains("-exec")) => {
                     // Find all -exec arguments and check if ANY command is non-readonly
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -444,37 +481,37 @@ mod tests {
...
-        .unwrap();
+            .unwrap();
         let res = tool_allow_exact_should_ask.eval_perm(&os, &agent);
         assert!(matches!(res, PermissionEvalResult::Ask));
 
         let tool_allow_exact_should_allow = serde_json::from_value::<ExecuteCommand>(serde_json::json!({
             "command": "allow_exact",
         }))
-        .unwrap();
+            .unwrap();
```
h16rkim, line 514: This change comes from running the lint (cargo clippy) command.

crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -72,9 +72,9 @@ impl ExecuteCommand {
         };
         const DANGEROUS_PATTERNS: &[&str] = &["<(", "$(", "`", ">", "&&", "||", "&", ";", "${", "\n", "\r", "IFS"];
 
-        if args
+        if DANGEROUS_PATTERNS
             .iter()
-            .any(|arg| DANGEROUS_PATTERNS.iter().any(|p| arg.contains(p)))
+            .any(|p| contains_but_not_ends_with(&self.command, p))
```
h16rkim, line 77: If DANGEROUS_PATTERNS only appears at the end, it is no longer treated as a dangerous command.
Because these commands cannot appear at the end, or if they do, they are not dangerous commands (e.g., `;`).

crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -109,14 +109,28 @@ impl ExecuteCommand {
...
-                                || arg.contains("-ok") // includes -okdir
-                                || arg.contains("-fprint") // includes -fprint0 and -fprintf
+                            arg.contains("-delete")
+                        || arg.contains("-ok") // includes -okdir
+                        || arg.contains("-fprint") // includes -fprint0 and -fprintf
                         }) =>
                 {
                     return true;
                 },
+                // Check -exec commands separately to allow readonly commands
```
h16rkim, line 119: * when using `find + -exec`, if the command following exec is readonly, no permission prompt is required.
* this also works correctly when multiple `-exec` clauses are present.

PR #2833, opened 2025-09-10
feat: add auto-announcement feature with /whatsnew command

abhraina-aws:
Implement automatic announcement system to keep users informed about new features and updates. Shows on startup for new versions or up to 4 times per version.

- Implement auto-announcement on startup for new versions or first 4 shows
- Add /whatsnew command to display latest features and updates
- Integrate whatsnew content into introspect tool for Q self-awareness


crates/chat-cli/src/database/settings.rs
```diff
@@ -77,6 +77,10 @@ pub enum Setting {
     ChatEnableHistoryHints,
     #[strum(message = "Enable the todo list feature (boolean)")]
     EnabledTodoList,
+    #[strum(message = "Last version that showed whats-new announcement (string)")]
```
brandonskiser: This would be better suited for SQLite rather than settings - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/database/mod.rs

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1089,6 +1095,84 @@ impl ChatSession {
...
+        if should_show {
+            // Display the whats-new content
+            let whatsnew_content = include_str!("../../../../../docs/whatsnew.md");
+
+            execute!(self.stderr, style::Print("\nWhat's New in Amazon Q CLI\n\n"),)?;
+
+            // Parse and display the markdown content
+            let mut last_was_section = false;
+
+            for line in whatsnew_content.lines() {
```
brandonskiser: Can we try to use the existing markdown for rendering this? https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/parse.rs

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -432,8 +434,11 @@ const WELCOME_TEXT: &str = color_print::cstr! {"<cyan!>
 const SMALL_SCREEN_WELCOME_TEXT: &str = color_print::cstr! {"<em>Welcome to <cyan!>Amazon Q</cyan!>!</em>"};
 const RESUME_TEXT: &str = color_print::cstr! {"<em>Picking up where we left off...</em>"};
 
+// Maximum number of times to show the What's New announcement per version
+const WHATS_NEW_MAX_SHOW_COUNT: i64 = 4;
```
brandonskiser: 4 times seems a bit excessive maybe, what do you think?

crates/chat-cli/src/cli/chat/mod.rs
brandonskiser: This is also duplicating the same content as above? Should be extracted into the same function.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -1089,6 +1095,119 @@ impl ChatSession {
 
         Ok(())
     }
+
+    /// Render whatsnew content with manual formatting (using parser's styling constants)
+    fn render_whatsnew_content(output: &mut impl Write) -> Result<()> {
```
brandonskiser: Functions that don't take `self` don't need to be an associated method

brandonskiser: Can this be moved to `util/ui.rs` instead? No reason for this to be added to `ChatSession`

brandonskiser (approved):

crates/chat-cli/src/util/ui.rs
```diff
@@ -0,0 +1,101 @@
...
+    execute!(
+        output,
+        style::Print("\nRun `/changelog` anytime to see the latest updates and features!\n\n")
+    )?;
+    Ok(())
+}
+
+fn clean_pr_links(text: &str) -> String {
+    // Remove PR links like " - [#2711](https://github.com/aws/amazon-q-developer-cli/pull/2711)"
+    if let Some(pos) = text.find(" - [#") {
```
kensave, line 71: How deterministic is this?

.amazonq/cli-todo-lists/1757012331798.json
brandonskiser, line 1: don't commit your todos!

crates/chat-cli/build.rs
```diff
@@ -322,3 +327,56 @@ fn main() {
...
+            if let Err(e) = std::fs::write("src/cli/feed.json", result.stdout) {
+                println!("cargo:warning=Failed to write feed.json: {}", e);
+            } else {
+                println!("cargo:warning=Successfully downloaded latest feed.json");
+            }
+        },
+        Ok(result) => {
+            println!("cargo:warning=Failed to download feed.json: HTTP error");
+            if !result.stderr.is_empty() {
+                println!("cargo:warning=Error: {}", String::from_utf8_lossy(&result.stderr));
```
brandonskiser: We should be erroring out of the build if we're not able to download the feed

crates/chat-cli/src/util/ui.rs
```diff
@@ -0,0 +1,146 @@
...
+use eyre::Result;
+
+use crate::cli::feed::Feed;
+
+/// Render changelog content from feed.json with manual formatting
+pub fn render_changelog_content(output: &mut impl Write) -> Result<()> {
+    let feed = Feed::load();
+    let recent_entries = feed.get_all_changelogs()
+        .into_iter()
+        .take(2) // Show last 2 releases
```
brandonskiser, line 18: why not just show the last entry?

test_whatsnew.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+SETTINGS_FILE="$HOME/Library/Application Support/amazon-q/settings.json"
```
brandonskiser: We're updating state now, not settings

Also we shouldn't commit any scripts to the top level of the codebase, if we add anything let's keep it under `scripts/`

abhraina-aws: This is accidental I will remove these.

.amazonq/cli-todo-lists/1757012331798.json
abhraina-aws: This is accidental I will remove these.

crates/chat-cli/src/util/ui.rs
abhraina-aws: I thought its good for balance. A lot of features are connected it brings continuity.

PR #2838, opened 2025-09-11
feat: Add /tangent tail to preserve the last tangent conversation

abhraina-aws:
image

Users can now keep the final question and answer from tangent mode by using `/tangent tail` instead of `/tangent`. This preserves the last Q&A pair when returning to the main conversation, making it easy to retain helpful insights discovered during exploration.

- `/tangent` - exits tangent mode (existing behavior unchanged)
- `/tangent tail` - exits tangent mode but keeps the last Q&A pair

This enables users to safely explore topics without losing the final valuable insight that could benefit their main conversation flow.

### DIFF
--- a/crates/chat-cli/src/cli/chat/conversation.rs
+++ b/crates/chat-cli/src/cli/chat/conversation.rs
@@ -269,6 +269,27 @@ impl ConversationState {
         }
     }
 
+    /// Exit tangent mode and preserve the last conversation entry (user + assistant)
+    pub fn exit_tangent_mode_with_tail(&mut self) {
+        if let Some(checkpoint) = self.tangent_state.take() {
+            // Capture the last history entry from tangent conversation if it exists
+            // and if it's different from what was in the main conversation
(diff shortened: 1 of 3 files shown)


crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -269,6 +269,27 @@ impl ConversationState {
         }
     }
 
+    /// Exit tangent mode and preserve the last conversation entry (user + assistant)
+    pub fn exit_tangent_mode_with_tail(&mut self) {
+        if let Some(checkpoint) = self.tangent_state.take() {
+            // Capture the last history entry from tangent conversation if it exists
```
kensave, line 275: Just the last conversation entry? any ideas on addong a summary of it?

abhraina-aws: I took user feedback. Its hard to know what is the relevant information in the tangent. People want to just retain whats relevant at the end by themselves. (edited)

PR #2839, opened 2025-09-11
feat: add daily heartbeat telemetry

abhraina-aws:
Tracks daily active users by sending amazonqcli_dailyHeartbeat event once per day. Uses fail-closed logic to prevent spam during database errors.

### DIFF
--- a/crates/chat-cli/src/telemetry/core.rs
+++ b/crates/chat-cli/src/telemetry/core.rs
@@ -19,6 +19,7 @@ use crate::telemetry::definitions::metrics::{
     AmazonqMessageResponseError,
     AmazonqProfileState,
     AmazonqStartChat,
+    AmazonqcliDailyHeartbeat,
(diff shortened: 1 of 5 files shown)


crates/chat-cli/src/telemetry/core.rs
```diff
@@ -499,6 +499,7 @@ impl Event {
                 }
                 .into_metric_datum(),
             ),
+            EventType::DailyHeartbeat {} => None, // Will be handled by toolkit telemetry
```
brandonskiser: This doesn't seem correct - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/telemetry/mod.rs#L677

If it's None, then it won't be sent

abhraina-aws: Yes this is different will fix it.

Issue #1101, opened 2025-04-06
Error when read non-UTF8 sourcecode file


raihaku:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.3.2 (24D81)

### Expected behaviour

Q Developer can read or retry non-UTF8 sourcecode without errors.

### Actual behaviour

Q Developer occured the error when read non-UTF8 sourcecode.

```
Read from filesystem
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
Reading directory: /Users/kensuke/Desktop/myproject/source code with maximum depth of 0

Reading from filesystem...
[...]

PR #2850, opened 2025-09-12
fix(read none utf8 file)

itsamine27:
*Issue #1101 , if available:*
so when u q cli read a file that are not utf8 encoded the program will get confused cause he only see  ?????? or sthg like that 
*Description of changes:*
so instead of that i make it detect the encoding used on that file then update it to utf8 so with that the ai can understand what is in the file   (and there are test in it)


itsamine27:
@tobuck-aws @mattsb42-aws @abhraina-aws @kennvene-aws  Could you please approve the workflow and review the changes? Thanks!

PR #2859, opened 2025-09-13
feat: Path based loading for agent

mekza:
People can now do `q chat --agent team/myagent` in addition to the current `q chat --agent myagent`. This will enable better management of agents in Q CLI directory

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -10,10 +10,7 @@ use std::collections::{
     HashSet,
 };
 use std::ffi::OsStr;
-use std::io::{
-    self,
-    Write,
-};
+use std::io::Write;
 use std::path::{
     Path,
     PathBuf,
@@ -37,12 +34,12 @@ use serde::{
     Serialize,
 };
 use thiserror::Error;
-use tokio::fs::ReadDir;
 use tracing::{
     error,
     info,
     warn,
 };
+use walkdir;
 use wrapper_types::ResourcePath;
 pub use wrapper_types::{
     OriginalToolName,
@@ -212,6 +209,38 @@ impl Default for Agent {
 }
 
 impl Agent {
+    /// Calculate the path-based identifier for this agent
+    /// Returns the relative path from agent directory to file (without .json extension)
+    /// Example: "team/assistant" for file at agents/team/assistant.json
+    pub fn path_identifier(&self, os: &Os) -> Option<String> {
+        let full_path = self.path.as_ref()?;
+
+        // Extract just the filename without extension for fallback
+        let file_stem = full_path.file_stem()?.to_str()?;
+
+        // Try to get the actual directory paths using the proper functions
+        // Check local workspace directory first
+        if let Ok(local_dir) = directories::chat_local_agent_dir(os) {
+            if let Ok(rel_path) = full_path.strip_prefix(&local_dir) {
+                if let Some(path_str) = rel_path.with_extension("").to_str() {
+                    return Some(path_str.to_string());
+                }
+            }
+        }
+
+        // Check global directory
+        if let Ok(global_dir) = directories::chat_global_agent_path(os) {
+            if let Ok(rel_path) = full_path.strip_prefix(&global_dir) {
+                if let Some(path_str) = rel_path.with_extension("").to_str() {
+                    return Some(path_str.to_string());
+                }
+            }
+        }
+
+        // Fallback to just filename
+        Some(file_stem.to_string())
+    }
+
     /// This function mutates the agent to a state that is writable.
     /// Practically this means reverting some fields back to their original values as they were
     /// written in the config.
@@ -339,39 +368,23 @@ impl Agent {
         }
     }
 
-    /// Retrieves an agent by name. It does so via first seeking the given agent under local dir,
-    /// and falling back to global dir if it does not exist in local.
+    /// Retrieves an agent by name or path identifier. It does so via first seeking the given agent
+    /// under local dir, and falling back to global dir if it does not exist in local.
+    /// Supports both JSON name field lookup and path-based lookup (e.g., "team/assistant").
+    /// Load all agents first and filter by both JSON name and path identifier
     pub async fn get_agent_by_name(os: &Os, agent_name: &str) -> eyre::Result<(Agent, PathBuf)> {
-        let config_path: Result<PathBuf, PathBuf> = 'config: {
-            // local first, and then fall back to looking at global
-            let local_config_dir = directories::chat_local_agent_dir(os)?.join(format!("{agent_name}.json"));
-            if os.fs.exists(&local_config_dir) {
-                break 'config Ok(local_config_dir);
+        let mut stderr = std::io::stderr();
+        let (agents, _) = Agents::load(&mut os.clone(), None, true, &mut stderr, true).await;
+
+        for (_, agent) in agents.agents {
+            if agent.name == agent_name || agent.path_identifier(os).as_deref() == Some(agent_name) {
+                if let Some(path) = agent.path.clone() {
+                    return Ok((agent, path));
+                }
             }
-
-            let global_config_dir = directories::chat_global_agent_path(os)?.join(format!("{agent_name}.json"));
-            if os.fs.exists(&global_config_dir) {
-                break 'config Ok(global_config_dir);
-            }
-
-            Err(global_config_dir)
-        };
-
-        match config_path {
-            Ok(config_path) => {
-                let content = os.fs.read(&config_path).await?;
-                let mut agent = serde_json::from_slice::<Agent>(&content)?;
-                let legacy_mcp_config = if agent.use_legacy_mcp_json {
-                    load_legacy_mcp_config(os).await.unwrap_or(None)
-                } else {
-                    None
-                };
-                let mut stderr = std::io::stderr();
-                agent.thaw(&config_path, legacy_mcp_config.as_ref(), &mut stderr)?;
-                Ok((agent, config_path))
-            },
-            _ => bail!("Agent {agent_name} does not exist"),
         }
+
+        bail!("Agent {agent_name} does not exist")
     }
 
     pub async fn load(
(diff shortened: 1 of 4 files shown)


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -203,6 +201,38 @@ impl Default for Agent {
 }
 
 impl Agent {
+    /// Calculate the path-based identifier for this agent
+    /// Returns the relative path from agent directory to file (without .json extension)
+    /// Example: "team/assistant" for file at agents/team/assistant.json
```
dingfeli, line 214: So this is introducing namespacing into the agent directory? Right now the directory is flat (you can put folders in there if you want to but it's not going to look at anything in this folder). 

I haven't gone through the rest of the PR yet but what's the expectation there? Are these agents nested in these sub folders expected to be discoverable?

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -203,6 +201,38 @@ impl Default for Agent {
...
+        let full_path = self.path.as_ref()?;
+
+        // Extract just the filename without extension for fallback
+        let file_stem = full_path.file_stem()?.to_str()?;
+
+        // Try to get the actual directory paths using the proper functions
+        // Check local workspace directory first
+        if let Ok(local_dir) = directories::chat_local_agent_dir(os) {
+            if let Ok(rel_path) = full_path.strip_prefix(&local_dir) {
+                if let Some(path_str) = rel_path.with_extension("").to_str() {
```
dingfeli, line 225: Why are we calling `with_extension` with an empty string here?

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -203,6 +201,38 @@ impl Default for Agent {
...
+                    return Some(path_str.replace('\\', "/"));
```
dingfeli: Not entirely sure what this is doing. Are we replacing backslash with forward slash? Why?

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -285,16 +315,31 @@ impl Agent {
...
-    /// Retrieves an agent by name. It does so via first seeking the given agent under local dir,
+    /// Retrieves an agent by name or path identifier. It does so via first seeking the given agent under local dir,
     /// and falling back to global dir if it does not exist in local.
+    /// Supports both JSON name field lookup and path-based lookup (e.g., "team/assistant").
     pub async fn get_agent_by_name(os: &Os, agent_name: &str) -> eyre::Result<(Agent, PathBuf)> {
+        // First try direct file path lookup (for path-based identifiers like "team/assistant")
         let config_path: Result<PathBuf, PathBuf> = 'config: {
-            // local first, and then fall back to looking at global
+            // Try path-based lookup first (local)
+            let local_path_config = directories::chat_local_agent_dir(os)?.join(format!("{agent_name}.json"));
```
dingfeli: We do allow a `name` field in the agent config and that does not necessarily need to match the file name. The name supplied to this this function (i.e. `agent_name`) is the name assigned to the aforementioned field. So this wouldn't work for those cases.

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -285,16 +315,31 @@ impl Agent {
         Ok(serde_json::to_string_pretty(&agent_clone)?)
     }
 
-    /// Retrieves an agent by name. It does so via first seeking the given agent under local dir,
+    /// Retrieves an agent by name or path identifier. It does so via first seeking the given agent under local dir,
     /// and falling back to global dir if it does not exist in local.
+    /// Supports both JSON name field lookup and path-based lookup (e.g., "team/assistant").
     pub async fn get_agent_by_name(os: &Os, agent_name: &str) -> eyre::Result<(Agent, PathBuf)> {
```
dingfeli, line 375: If I am not mistaken here we are overloading `agent_name` with `$namespace/agent_name`. I think that can be confusing.

If we are already reading and parsing every json file in the local and global agent directory, we shouldn't need to reconstruct the associated path, it would mean that we already have a comprehensive list of agents. We should be able to just filter them by the namespace by looking through `path` on the agent.

dingfeli:
Implementation aside, I think this is a change that involves product level decisions. The way this feature is designed is fairly opinionated and is one that is based on a new concept / notion of a namespace in the agent directory, which is itself something that warrants some discussions. 

On that front, I'll bring this up with the team and I'll update it in the reply to this comment accordingly.

crates/chat-cli/src/cli/agent/mod.rs
mekza: That's exactly the idea yes. A flat directory makes it not really efficient to manage multiple agent and you end up prefixing file `myteam_myagent.json`. This PR address that by introducing a directory where you can group your file like a namespace would do.

crates/chat-cli/src/cli/agent/mod.rs
mekza: The idea is to strip the `.json` extension and convert `/path/to/agents/team/myagent.json` into `team/myagent`

crates/chat-cli/src/cli/agent/mod.rs
mekza: Path representations differ across platforms, the idea here is to allow to do `--agent team/myagent` regardless of whether they are on Windows, macOS or Linux. We could use `relative_path` crate [1] which provide portable relative paths support. Since it's not used anywhere I thought that would be much simpler but probably less elegant

[1] https://crates.io/crates/relative-path

crates/chat-cli/src/cli/agent/mod.rs
mekza: I think you are right but this function has a dual lookup approach. First it tries to find {agent_name}.json as a file path (for path-based identifiers like 'team/assistant') then if it fails it falls back to loading all agents and matches against both the JSON name field AND the path identifier and therefor it should not break any experience. Users will still be able to name their agent `team/myagent` or just `myagent`

crates/chat-cli/src/cli/agent/mod.rs
mekza: Good point, reconstructing path is clearly redundant. I will remove the direct path construction logic (lines 325-347), make sure we always load all agents and filter by both JSON name and path_identifier. This will avoid confusing dual semantics, bonus point it's more efficient. What do you think?

mekza:
@dingfeli I addressed all of your feedback, let me know what you think

dingfeli:
So I spoke with the team some more and the consensus is that this implementation is slightly premature and it would make more sense for when / if we have repositories for agents. I do think this is something we would want to eventually support but perhaps to compliment a better thought out agent distribution paradigm.

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -826,21 +842,33 @@ pub struct AgentsLoadMetadata {
     pub launched_agent: String,
 }
 
-async fn load_agents_from_entries(
-    mut files: ReadDir,
+async fn load_agents_from_directory_recursive(
```
dingfeli: Not sure if this is appropriate. This makes it seem like this function is creating another future to in calling itself. I would not make this change, especially when we are not preserving the previous version. There is no need to make this distinction.

crates/chat-cli/src/cli/agent/mod.rs
dingfeli: I see. Can you help me understand why we can't just return `path_str` here? Is `path_str` not already what we want?

dingfeli:
Revisiting this as per Matt's request. Left some more comments.

crates/chat-cli/src/cli/agent/mod.rs
mekza: I think your comment refers to line 220:221:
```rust
let normalized = RelativePath::new(path_str).normalize();
return Some(normalized.as_str().to_string());
```
If yes, I agree we should just return `return Some(path_str.to_string());`

crates/chat-cli/src/cli/agent/mod.rs
mekza: I think there might be a misunderstanding here. This isn't about async recursion or the function calling itself. It's actually a change in how we traverse directories to find agents. 

The old function `load_agents_from_entries` took a `ReadDir` iterator only looking at flat directory. `load_agents_from_directory_recursive` uses `walkdir` crate to walk through all of the subdirectories. I am not keeping the old version because we changed the behavior completely from flat to recursive. This is what enables the path-based loading feature.

Are you pointing out walkdir (sync) usage inside an async? 

If yes then probably the best approach would be to spawn a blocking task with `tokio::task::spawn_blocking` what do you think?

dingfeli: I don't think it matters all that much so I am okay with this if you insist on it. 

FWIW I am saying that the naming of the function suggest that the function is being called recursively (i.e. there is another invocation of `load_agents_from_directory_recursive` inside of it). I understand that you are actually describing the fact that we looking through all the directories and its children.

In rust, async function returns a `Pin<Box<dyn Future<Output = ...>>>`, hence my comment about it returning a future.  

> I am not keeping the old version because we changed the behavior completely from flat to recursive.

Right, so why specify this is recursive at all? It implies that there is a non-recursive variant of this function somewhere.

mekza: Done!

crates/chat-cli/Cargo.toml
```diff
@@ -75,6 +75,7 @@ r2d2.workspace = true
 r2d2_sqlite.workspace = true
 rand.workspace = true
 regex.workspace = true
+relative-path.workspace = true
```
dingfeli: Do we still need this?

Cargo.toml
```diff
@@ -82,6 +82,7 @@ r2d2_sqlite = "0.25.0"
 rand = "0.9.0"
 rayon = "1.10.0"
 regex = "1.7.0"
+relative-path = "1.9"
```
dingfeli: Do we still need this?

PR #2875, opened 2025-09-15
Support preToolUse and postToolUse hook

erbenmo:
feat: Add hook support for preToolUse and postToolUse
    
    With this change, customer can define arbitrary command to be run
    before and after tool use. Example definition:
    
    ```
      "hooks": {
        "preToolUse": [
          {
            "matcher": "execute_bash",
            "command": "cat >> /tmp/bash_audit_log && echo >> /tmp/bash_audit_log"
          }
        ],
        "postToolUse": [
          {
            "matcher": "fs_write",
            "command": "cargo fmt --all"
          }
        ]
      }
    ```
    
    Hook command receives input from STDIN in JSON format. Hook command is run as a
    separate process. It can return exit code and write to stdout / stderr.
    
    Overall life cycle of a tool call:
    system-level tool validation -> preToolUse -> user prompt for approving tool -> running tool -> postToolUse
    
    Hook input
    - hook_event_name
    - cwd
    - tool_name
    - tool_input
    - tool_response
    
    Hook output handling
    preToolUse
    - Exit code 0: nothing.
    - Exit code 2: tool run is blocked. the relevant context including hook stderr is sent back to LLM.
    - Other code: print stderr to screen. does NOT block tool.
    postToolUse:
    - Exit code 0: nothing.
    - Other code: print stderr to screen.

### DIFF
--- a/crates/chat-cli/src/cli/agent/hook.rs
+++ b/crates/chat-cli/src/cli/agent/hook.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::fmt::Display;
 
 use schemars::JsonSchema;
@@ -11,9 +10,6 @@ const DEFAULT_TIMEOUT_MS: u64 = 30_000;
 const DEFAULT_MAX_OUTPUT_SIZE: usize = 1024 * 10;
 const DEFAULT_CACHE_TTL_SECONDS: u64 = 0;
 
-#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
-pub struct Hooks(HashMap<HookTrigger, Hook>);
-
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, Eq, PartialEq, JsonSchema, Hash)]
 #[serde(rename_all = "camelCase")]
 pub enum HookTrigger {
@@ -21,6 +17,10 @@ pub enum HookTrigger {
     AgentSpawn,
     /// Triggered per user message submission
     UserPromptSubmit,
+    /// Triggered before tool execution
+    PreToolUse,
+    /// Triggered after tool execution
+    PostToolUse,
 }
 
 impl Display for HookTrigger {
@@ -28,6 +28,8 @@ impl Display for HookTrigger {
         match self {
             HookTrigger::AgentSpawn => write!(f, "agentSpawn"),
             HookTrigger::UserPromptSubmit => write!(f, "userPromptSubmit"),
+            HookTrigger::PreToolUse => write!(f, "preToolUse"),
+            HookTrigger::PostToolUse => write!(f, "postToolUse"),
         }
     }
 }
@@ -61,6 +63,11 @@ pub struct Hook {
     #[serde(default = "Hook::default_cache_ttl_seconds")]
     pub cache_ttl_seconds: u64,
 
+    /// Optional regex matcher for hook
+    /// Currently used for matching tool name of PreToolUse and PostToolUse hook
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub matcher: Option<String>,
+
     #[schemars(skip)]
     #[serde(default, skip_serializing)]
     pub source: Source,
@@ -73,6 +80,7 @@ impl Hook {
             timeout_ms: Self::default_timeout_ms(),
             max_output_size: Self::default_max_output_size(),
             cache_ttl_seconds: Self::default_cache_ttl_seconds(),
+            matcher: None,
             source,
         }
     }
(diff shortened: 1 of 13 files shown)


crates/chat-cli/src/cli/agent/hook.rs
```diff
@@ -61,6 +63,11 @@ pub struct Hook {
     #[serde(default = "Hook::default_cache_ttl_seconds")]
     pub cache_ttl_seconds: u64,
 
+    /// Optional regex matcher for hook
+    /// Currently used for matching tool name of PreToolUse and PostToolUse hook
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub matcher: Option<String>,
```
brandonskiser, line 69: If we're specifically using regex, can this be called `regex` instead?

Reason being is we already have globbing and regex being mixed in the agent config, so I think being explicit here is better.

crates/chat-cli/src/cli/chat/cli/hooks.rs
```diff
@@ -44,6 +44,31 @@ use crate::cli::chat::{
...
+
+/// Check if a hook matches a tool name based on its matcher pattern
+fn hook_matches_tool(hook: &Hook, tool_name: &str) -> bool {
+    match &hook.matcher {
+        None => true, // No matcher means the hook runs for all tools
+        Some(pattern) => {
+            // Convert single pattern to HashSet for matches_any_pattern
+            let mut patterns = std::collections::HashSet::new();
+            patterns.insert(pattern.clone());
+            matches_any_pattern(&patterns, tool_name)
```
brandonskiser: `matches_any_pattern` doesn't use regex, it uses glob, so we should decide on whether or not `matcher` should be a regex or glob

crates/chat-cli/src/cli/chat/cli/hooks.rs
```diff
@@ -44,6 +44,31 @@ use crate::cli::chat::{
     ChatSession,
     ChatState,
 };
+use crate::util::pattern_matching::matches_any_pattern;
+
+/// Hook execution result: (exit_code, output)
+/// Output is stdout if exit_code is 0, stderr otherwise.
+pub type HookOutput = (i32, String);
```
brandonskiser, line 53: Let's define a concrete struct here instead of a type alias so it's clearer what these fields are

```
struct HookOutput {
status_code: i32,
stdout: String,
stderr: String,
}

File: crates/chat-cli/src/cli/agent/hook.rs (hunk 1 above)
Comment (brandonskiser) [reply crates/chat-cli/src/cli/agent/hook.rs:69]: From checking the rest of the code, I guess matcher is fine - as long as we update this comment and any other references to this being a regex

**Review by brandonskiser — APPROVED** [review]

File: docs/agent-format.md [hunk 4]
```diff
@@ -256,19 +256,37 @@ Resources can include:
...
     ],
     "userPromptSubmit": [
       {
-        "command": "ls -la",
+        "command": "ls -la"
+      }
+    ],
+    "preToolUse": [
+      {
+        "matcher": "execute_bash",
```
kensave, line 278: Curious, what are your thoughts for this when you want to run hooks for certain inputs of certain tools? e.g Run fmt before git commit command is called.

docs/hooks.md
```diff
@@ -0,0 +1,138 @@
+# Hooks
+
+Hooks allow you to execute custom commands at specific points during agent lifecycle and tool execution. This enables security validation, logging, formatting, context gathering, and other custom behaviors.
```
kensave, line 3: Thanks for adding this, we could now add to introspect tool: https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/tools/introspect.rs#L65

kensave (approved):

crates/chat-cli/src/cli/agent/hook.rs
erbenmo: Updated comment from regex -> glob

crates/chat-cli/src/cli/chat/cli/hooks.rs
erbenmo: the intention is to use the same glob matching for "allowed_tools"

docs/agent-format.md
erbenmo: User can add add the filtering logic for "tool_input" in the hook command. The hook event looks like:

{"hook_event_name":"preToolUse","cwd":"/Users/moerben/Documents/Work/2025/Project/amazon-q-developer-cli","tool_name":"execute_bash","tool_input":{"command":"git commit -m \"update documentation based on reviewer's feedback\"","summary":"Create git commit with the specified message"}}

docs/hooks.md
erbenmo: Nice. I didn't know that!

PR #2890, opened 2025-09-16
feat: implement delegate tool for agent task management

abhraina-aws:
- Add delegate tool for launching and managing asynchronous agent processes
- Support agent validation with approval UI for security
- Enable status checking for individual agents or all agents
- Implement background process monitoring and execution tracking
- Store execution state in ~/.aws/amazonq/.subagents/
- Support both global (~/.aws/amazonq/cli-agents/) and local (.amazonq/cli-agents/) agent configurations
- Add experimental flag for delegate functionality
- Clean error handling with proper agent not found messages

The delegate tool allows users to offload tasks to specialized agents that run independently, enabling parallel work and better task organization. Agents require user approval before execution for security, and the tool provides comprehensive status tracking.

### DIFF
--- a/docs/experiments.md
+++ b/docs/experiments.md
@@ -108,6 +108,29 @@ Amazon Q CLI includes experimental features that can be toggled on/off using the
 
 **When enabled:** Use `/tangent` or the keyboard shortcut to create a checkpoint and explore tangential topics. Use the same command to return to your main conversation.
 
+### Delegate
+**Command:** `/delegate`
+**Keyboard Shortcut:** `Ctrl+D` (customizable via `q settings chat.delegateModeKey x`)
+**Description:** Launch and manage asynchronous task processes. Enables running Q chat sessions with specific agents in parallel to the main conversation.
+**Usage:**
+- `/delegate launch "Fix the bug in main.rs"` - Launch task with default agent
+- `/delegate launch --agent coding "Fix the bug in main.rs"` - Launch with specific agent (shows approval dialog)
+- `/delegate status` - Show summary of all tasks
+- `/delegate status abc12345` - Show status of specific task
+- `/delegate read abc12345` - Read output from completed task (triggers LLM analysis)
+- `/delegate delete abc12345` - Delete task and its files
+- `/delegate list` - List all tasks with timestamps and brief info
+**Agent Approval Flow:**
+When using `--agent`, you'll see an approval dialog:
+```
+Agent: coding
+Description: Coding assistant for software development
+Task: Fix the bug in main.rs
+Tools: fs_read, fs_write, execute_bash
+⚠️  This task will run with trust-all permissions and can execute commands or consume system/cloud resources. Continue? [y/N]:
+```
+**When enabled:** Use `/delegate` commands or `Ctrl+D` to spawn independent Q processes that work on tasks while you continue your main conversation. Tasks with agents require explicit approval and show agent details. Tasks without agents run with a warning about trust-all permissions. Once delegated, tasks work independently and you can check progress, read results, or delete them as needed.
+
 ### TODO Lists
 **Tool name**: `todo_list`
 **Command:** `/todos`  
--- a/crates/chat-cli/src/cli/chat/cli/delegate.rs
+++ b/crates/chat-cli/src/cli/chat/cli/delegate.rs
@@ -1,0 +1,543 @@
+use std::collections::HashMap;
+use std::io::{
+    Write,
+    stdin,
+    stdout,
+};
+use std::path::PathBuf;
+use std::process::Stdio;
+
+use crossterm::execute;
+use crossterm::style::{
+    Color,
+    Print,
+    SetForegroundColor,
+};
+use eyre::{
+    Result,
+    bail,
+    eyre,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use time::OffsetDateTime;
+use tokio::process::Command;
+use uuid::Uuid;
+
+use crate::cli::chat::{
+    ChatError,
+    ChatSession,
+    ChatState,
+};
+use crate::database::settings::Setting;
+use crate::os::Os;
+use crate::util::directories::home_dir;
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct SubagentHeader {
+    pub launched_at: String,
+    pub agent: Option<String>,
+    pub prompt: String,
+    pub status: String, // "active", "completed", "failed"
+    pub pid: u32,
+    pub completed_at: Option<String>,
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct SubagentContent {
+    pub output: String,
+    pub exit_code: Option<i32>,
+}
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct StatusFile {
+    pub subagents: HashMap<String, SubagentHeader>,
+    pub last_updated: String,
+}
+
+#[derive(Debug, PartialEq, clap::Subcommand)]
+pub enum DelegateArgs {
+    /// Show status of tasks
+    Status {
+        /// Specific task UUID (optional)
+        uuid: Option<String>,
+    },
+    /// Read output from a task
+    Read {
+        /// Task UUID
+        uuid: String,
+    },
+    /// Delete a task and its files
+    Delete {
+        /// Task UUID
+        uuid: String,
+    },
+    /// List all tasks
+    List,
+    /// Launch a new task
+    Launch {
+        /// Agent to use for the task
+        #[arg(long)]
+        agent: Option<String>,
+        /// Task description
+        #[arg(trailing_var_arg = true, allow_hyphen_values = true)]
+        prompt: Vec<String>,
+    },
+}
+
+impl DelegateArgs {
+    pub async fn execute(self, os: &mut Os, session: &mut ChatSession) -> Result<ChatState, ChatError> {
+        if !is_enabled(os) {
+            return Err(ChatError::Custom(
+                "Delegate feature is not enabled. Enable it with /experiment command.".into(),
+            ));
+        }
+
+        let result = match self {
+            DelegateArgs::Status { uuid } => show_status(os, uuid.as_deref()).await,
+            DelegateArgs::Read { uuid } => {
+                // For read command, we want to trigger LLM analysis
+                let full_uuid = match find_subagent_by_partial_uuid(os, &uuid).await {
+                    Ok(Some(uuid)) => uuid,
+                    Ok(None) => return Err(ChatError::Custom("Subagent not found".into())),
+                    Err(e) => return Err(ChatError::Custom(e.to_string().into())),
+                };
+
+                let content_path = match get_subagents_dir(os).await {
+                    Ok(dir) => dir.join(format!("{}.json", full_uuid)),
+                    Err(e) => return Err(ChatError::Custom(e.to_string().into())),
+                };
+
+                // Print the basic info first
+                match read_subagent(os, &uuid, session).await {
+                    Ok(message) => println!("{}", message),
+                    Err(e) => return Err(ChatError::Custom(e.to_string().into())),
+                }
+
+                // Return HandleInput state to trigger LLM analysis
+                return Ok(ChatState::HandleInput {
+                    input: format!(
+                        "Please read and summarize the subagent results from this file: {}\n\nProvide a concise 2-3 bullet point summary of what was accomplished.",
+                        content_path.display()
+                    ),
+                });
+            },
+            DelegateArgs::Delete { uuid } => {
+                let full_uuid = match find_subagent_by_partial_uuid(os, &uuid).await {
+                    Ok(Some(uuid)) => uuid,
+                    Ok(None) => return Err(ChatError::Custom("Task not found".into())),
+                    Err(e) => return Err(ChatError::Custom(e.to_string().into())),
+                };
+
+                delete_subagent(os, &full_uuid).await
+            },
+            DelegateArgs::List => list_subagents(os).await,
+            DelegateArgs::Launch { agent, prompt } => {
+                let prompt_str = prompt.join(" ");
+                if prompt_str.trim().is_empty() {
+                    return Err(ChatError::Custom("Please provide a prompt for the task".into()));
+                }
+
+                launch_subagent(os, session, agent.as_deref(), &prompt_str).await
+            },
+        };
+
+        match result {
+            Ok(output) => {
+                writeln!(session.stderr, "{}", output)?;
+            },
+            Err(e) => {
+                writeln!(session.stderr, "Error: {}", e)?;
+            },
+        }
+
+        Ok(ChatState::PromptUser {
+            skip_printing_tools: false,
+        })
+    }
+}
+
+fn is_enabled(os: &Os) -> bool {
+    os.database.settings.get_bool(Setting::EnabledDelegate).unwrap_or(false)
+}
+
+async fn get_subagents_dir(os: &Os) -> Result<PathBuf> {
+    let home = home_dir(os)?;
+    let dir = home.join(".aws").join("amazonq").join(".subagents");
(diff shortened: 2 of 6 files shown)


crates/chat-cli/src/cli/chat/tools/tool_index.json
```diff
@@ -379,5 +379,28 @@
...
+  "delegate": {
+    "name": "delegate",
+    "description": "Launch and manage asynchronous agent processes. This tool allows you to delegate tasks to agents that run independently in the background.\n\nOperations:\n- launch: Start a new task with an agent (requires task parameter, agent is optional)\n- status: Check agent status and get full output if completed. Agent is optional - defaults to 'all' if not specified\n\nIf no agent is specified for launch, uses 'default_agent'. Only one task can run per agent at a time. Files are stored in ~/.aws/amazonq/.subagents/\n\nIMPORTANT: If a specific agent is requested but not found, DO NOT automatically retry with 'default_agent' or any other agent. Simply report the error and available agents to the user.\n\nExample usage:\n1. Launch with agent: {\"operation\": \"launch\", \"agent\": \"rust-agent\", \"task\": \"Create a snake game\"}\n2. Launch without agent: {\"operation\": \"launch\", \"task\": \"Write a Python script\"}\n3. Check specific agent: {\"operation\": \"status\", \"agent\": \"rust-agent\"}\n4. Check all agents: {\"operation\": \"status\", \"agent\": \"all\"}\n5. Check all agents (shorthand): {\"operation\": \"status\"}",
```
brandonskiser, line 385: What about running multiple delegate tasks using the same agent config? This seems to only support one agent at a time

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -114,156 +114,156 @@ impl FsRead {
         }
 
         let is_in_allowlist = matches_any_pattern(&agent.allowed_tools, "fs_read");
-        let settings = agent.tools_settings.get("fs_read").cloned()
+        let settings = agent
```
brandonskiser, line 117: These only seem to be formatting changes, right?

crates/chat-cli/src/cli/chat/tools/delegate/process.rs
```diff
@@ -0,0 +1,104 @@
...
+    if agent != DEFAULT_AGENT {
+        cmd.arg("--agent").arg(agent);
+    }
+
+    cmd.arg(task)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .stdin(Stdio::null());
+
+    let child = cmd.spawn()?;
```
dingfeli: Haven't gone through the rest of the PR yet but I see here we are spawning the the child process without promoting them to process group leader. 

Any signals received by the main process is also going to be received by the child process. So a SIGINT from the main process is going to terminate this. Not entirely sure if that's what we want, just wanted to point that out.

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+
+const OPERATION_LAUNCH: &str = "launch";
+const OPERATION_STATUS: &str = "status";
+const DEFAULT_AGENT: &str = "default";
+const ALL_AGENTS: &str = "all";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Delegate {
+    /// Operation to perform: launch or status
+    pub operation: String,
```
brandonskiser: You can define operation as an enum, and use `strum` derive macros for defining it.

I think something like:
```
#[derive(strum::EnumString)]
#[strum(serialize_all = "lowercase")]
enum Operation {
Launch,
Status,
}
```

crates/chat-cli/src/cli/chat/tools/delegate/process.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+pub fn format_launch_success(agent: &str, task: &str) -> String {
+    format!(
+        "✓ Task launched with agent '{}'\nTask: {}\nAgent is now working independently. Use the delegate tool with 'status' operation to check progress.",
+        agent, task
+    )
+}
+
+async fn monitor_agent_execution(mut execution: AgentExecution, os: Os) {
```
brandonskiser: It doesn't seem clear as to what is this supposed to be doing? It seems that we're just launching another `q` instance to run the task.

`monitor_agent_execution` being launched on another task through `start_monitoring` would lead me to believe we're doing some loop and continually polling the status of an agent, but instead we're just launching the same task again as is done in `spawn_agent_process`? Or am I missing something

crates/chat-cli/src/cli/chat/tools/delegate/agent_paths.rs
```diff
@@ -0,0 +1,30 @@
...
+        if !agents_dir.exists() {
+            std::fs::create_dir_all(&agents_dir)?;
+        }
+
+        Ok(agents_dir)
+    }
+
+    pub async fn cli_agents_dir(os: &Os) -> Result<PathBuf> {
+        let home_dir = os.env.home().unwrap_or_default();
+        Ok(home_dir.join(".aws").join("amazonq").join("cli-agents"))
```
brandonskiser: We should have this already defined in `src/util/directories.rs`

crates/chat-cli/src/cli/chat/tools/delegate/agent_paths.rs
```diff
@@ -0,0 +1,30 @@
+use std::path::PathBuf;
+
+use eyre::Result;
+
+use crate::os::Os;
+
+pub struct AgentPaths;
+
+impl AgentPaths {
+    pub async fn subagents_dir(os: &Os) -> Result<PathBuf> {
```
brandonskiser: Should we add this to `src/util/directories.rs` instead for consistency with the rest of the code base? Common paradigm to keep constants like this inside `directories.rs` currently.

crates/chat-cli/src/cli/chat/tools/delegate/types.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub completed_at: Option<String>,
+    #[serde(default)]
+    pub pid: u32,
+    #[serde(default)]
+    pub exit_code: Option<i32>,
+    #[serde(default)]
+    pub output: String,
+}
+
+mod status_serde {
```
dingfeli: Why do we need this to be deserialized with a custom deserializer? It does not look like the logic here is doing anything all that special.

crates/chat-cli/src/cli/chat/tools/delegate/errors.rs
```diff
@@ -0,0 +1,27 @@
+pub struct AgentError;
```
brandonskiser: This isn't being used as an error? We should implement `std::error::Error` (easy with `thiserror::Error` derive macro) and use this as an error return value

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -114,156 +114,156 @@ impl FsRead {
         }
```
dingfeli, line 114: Is this file just formatting change?

crates/chat-cli/src/cli/chat/tools/delegate/types.rs
```diff
@@ -0,0 +1,190 @@
...
+    #[serde(default)]
+    pub agent: String,
+    #[serde(default)]
+    pub task: String,
+    #[serde(default, with = "status_serde")]
+    pub status: AgentStatus,
+    #[serde(default)]
+    pub launched_at: String,
+    #[serde(default)]
+    pub completed_at: Option<String>,
```
brandonskiser, line 42: what's the format of this timestamp? Could we use `chrono` crate instead?

e.g.
```
    #[serde(with = "chrono::serde::ts_seconds_option")]
    pub completed_at: Option<DateTime<Utc>>,
```

crates/chat-cli/src/cli/chat/tools/delegate/types.rs
```diff
@@ -0,0 +1,190 @@
+use std::str::FromStr;
+
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
```
brandonskiser: We should use `strum::EnumString` and `strum::Display` instead of having to manually implement serialize/deserialize/fromstr/as_str

crates/chat-cli/src/cli/chat/tools/delegate/process.rs
abhraina-aws: Thanks. I think I would want to keep the background process running if the main one is killed. Also another angle is that we are trying to keep this simple. But Matt probably is the best to comment on this.

crates/chat-cli/src/cli/chat/tools/delegate/types.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    pub fn is_active(&self) -> bool {
+        self.status == AgentStatus::Running
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub struct AgentConfig {
```
brandonskiser, line 76: What is this used for? Just deserializing a subset of the full agent config?

docs/experiments.md
```diff
@@ -79,6 +79,35 @@ Amazon Q CLI includes experimental features that can be toggled on/off using the
 **Settings:**
 - `chat.enableTodoList` - Enable/disable TODO list functionality (boolean)
 
+### Delegate
+
+**Command:** `/delegate`
```
brandonskiser, line 157: All of this is left unimplemented intentionally?

crates/chat-cli/src/cli/chat/tools/delegate/agent_paths.rs
abhraina-aws: I can move if that is the convention. Since this is definitely experimental don't want to have others accidentally relying on this code.

crates/chat-cli/src/cli/chat/tools/delegate/agent_paths.rs
abhraina-aws: Ok, I can check that out.

crates/chat-cli/src/cli/chat/tools/delegate/errors.rs
abhraina-aws: Need to discuss this.

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
abhraina-aws: Sure

crates/chat-cli/src/cli/chat/tools/delegate/process.rs
abhraina-aws: Let me check this out.

crates/chat-cli/src/cli/chat/tools/delegate/types.rs
abhraina-aws: Ok

docs/experiments.md
abhraina-aws: I will update it since this is legacy now.

crates/chat-cli/src/cli/chat/tools/tool_index.json
dingfeli: Also how would the model understand what agents are available to run? Is there a "query" or something equivalent where the model would learn about the list of available agents as well as their descriptions?

crates/chat-cli/src/cli/chat/tools/delegate/agent.rs
```diff
@@ -0,0 +1,132 @@
+use std::path::PathBuf;
+use eyre::Result;
+use serde_json;
+
+use crate::cli::chat::tools::delegate::types::{AgentConfig, AgentExecution};
+use crate::cli::chat::tools::delegate::ui::{display_agent_info, get_user_confirmation};
+use crate::os::Os;
+
+fn home_dir(os: &Os) -> Result<PathBuf> {
```
dingfeli: I think we already have a util for this in `directories.rs`

crates/chat-cli/src/cli/chat/tools/delegate/execution.rs
```diff
@@ -0,0 +1,126 @@
...
+        exit_code: None,
+        output: String::new(),
+    };
+
+    save_agent_execution(os, &execution).await?;
+    
+    // Start monitoring with the actual child process
+    tokio::spawn(monitor_child_process(child, execution.clone(), os.clone()));
+    
+    Ok(execution)
```
dingfeli: Are we planning to do something with the `execution` returned here? This has a completely separate lifetime as the one passed in to `monitor_child_process` and you really won't be able to derive anything about the child process with what is returned here.

crates/chat-cli/src/cli/chat/tools/delegate/agent.rs
```diff
@@ -0,0 +1,132 @@
...
+                            .collect())
+                })
+                .unwrap_or_else(|| vec!["No tools specified".to_string()]),
+        });
+    }
+    
+    Err(eyre::eyre!("Agent config not found"))
+}
+
+pub async fn list_available_agents(os: &Os) -> Result<Vec<String>> {
```
dingfeli: This is something we already do at the start of each session. We actually load up all available agents in scope. This is how are able to generate the drop down menu when we do `/agent swap`. 

In the future we would probably move away from deserializing them but we would still retain the list.

dingfeli: Also, not accounted for in this is the list of local agents in `cwd/.amazonq/`. 

I think with how `Tool` is designed we would have to just pass the list of active agents in `invoke` (we already pass active agent in there, we could instead pass `&Agents`.

crates/chat-cli/src/cli/chat/tools/delegate/execution.rs
```diff
@@ -0,0 +1,126 @@
...
+            let stdout = String::from_utf8_lossy(&output.stdout);
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            execution.output = if stderr.is_empty() {
+                stdout.to_string()
+            } else {
+                format!("STDOUT:\n{}\n\nSTDERR:\n{}", stdout, stderr)
+            };
+            
+            // Save to ~/.aws/amazonq/.subagents/{agent}.json
+            if let Err(e) = save_agent_execution(&os, &execution).await {
```
dingfeli, line 79: I see that we are saving the output (which captures `stdout` and `stderr` of the child process) to a file. 
I am a little confused as to what the intended use case for this feature is. From what I can see, it looks to be: 
1. User somehow prompts the model to invoke this tool, thus spawning a child process
2. User continues to chat with the model from which the child process is spawned
3. User periodically checks to see if the child process is done (from what I can see there are not mechanisms here for the model to check, and it was also not explicitly told to do so) 
4. If it just so happens that one of these checks concludes that the task is done the model would then take the output from the delegate agent and continue with the conversation. 

Does that sound like the use case we are aiming to accommodate here?

crates/chat-cli/src/cli/chat/tools/delegate/agent.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+pub async fn save_agent_execution(os: &Os, execution: &AgentExecution) -> Result<()> {
+    let file_path = agent_file_path(os, &execution.agent).await?;
+    let content = serde_json::to_string_pretty(execution)?;
+    os.fs.write(&file_path, content).await?;
+    Ok(())
+}
+
+async fn agent_file_path(os: &Os, agent: &str) -> Result<PathBuf> {
```
dingfeli: This function looks to be responsible for deriving the directory in which child process output is saved. 
There are nothing session specific about the path being created here. It's essentially `home/.aws/amazonq/.subagents/agent_name.json`. 

If my understanding correct, we could be clobbering this output if we have multiple sessions at the same time. Or if we run delegate with the same agent multiple times in the same session.

PR #2896, opened 2025-09-16
feat: Adds checkpointing functionality using Git CLI commands

evanliu048:
*Issue #, if available:*

*Description of changes:*
Introduce session-scoped checkpoints for the current repo. Q CLI now snapshots file changes into a shadow bare git repo and lets users list, expand, diff, and restore to any checkpoint. Errors are short and human-readable; empty diffs print “No differences.” The conversation history is also unwind when restoring a checkpoint

**Commands**
- `/checkpoint init` — Manually enable checkpoints if not in a git repo.
  - (Auto): If starting inside a git repo, checkpoints auto-enable (ephemeral, cleaned on session end).
- `/checkpoint list [--limit N]` — Turn-level checkpoints with file stats.
- `/checkpoint expand <tag> `— Show tool-level checkpoints under a turn, with “(+/~/-) files” badges.
- `/checkpoint diff <tag1> [tag2|HEAD]`
  - Prints header then diff, or “No differences.” if none.
  - Unknown tags: Checkpoint with tag '<t>' does not exist! Use /checkpoint list to see available checkpoints
- `/checkpoint restore [<tag>] [--hard]`
  - Default: revert tracked changes & deletions; do not delete files created after the checkpoint. 
  - --hard: make workspace exactly match the checkpoint; deletes tracked files created after it.
  - If `<tag>` omitted, shows interactive picker. 
- `/checkpoint clean` — Delete session shadow repo. 
image

**Implementation (high-level)**

Shadow storage: **bare** git repo under `get_shadow_repo_dir(conversation_id)`, operating on `--work-tree=.`.
Helper `run_git(...)` centralizes `--git-dir/--work-tree` and error handling.
File stats via `git diff --name-status`; treat `R/C` as modified; cache per-tag stats.
Previous-tag resolution for `X` and `X.Y` forms.
Soft vs hard restore wired through `git checkout` vs `git reset --hard`.

**Examples**
```
/checkpoint list
[0] 2025-09-18 14:00:00 - Initial checkpoint
[1] 2025-09-18 14:05:31 - add two_sum.py (+1 file)
[2] 2025-09-18 14:07:10 - add tests (modified 1)

/checkpoint expand 2
[2] 2025-09-18 14:07:10 - add tests
 └─ [2.1] fs_write: Add minimal test cases to two_sum.py (modified 1)

/checkpoint diff 2
Comparing current state with checkpoint [2]:
No differences.

/checkpoint restore 1 --hard
Restored checkpoint: 1
```


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2304,6 +2360,75 @@ impl ChatSession {
...
+                let enabled = os
+                    .database
+                    .settings
+                    .get_bool(Setting::EnabledCheckpoint)
+                    .unwrap_or(false);
+                if invoke_result.is_err() || !enabled {
+                    String::new()
+                }
+                // Take manager out temporarily to avoid borrow conflicts
+                else if let Some(mut manager) = self.conversation.checkpoint_manager.take() {
```
dingfeli, line 2378: Is there any invocation here that requires a `mut self` for manager? If not we can just do 
```rust
else if let Some(manager) = self.conversation.checkpoint_manager.as_mut() {
...
}
```

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2346,8 +2471,11 @@ impl ChatSession {
                         style::SetAttribute(Attribute::Bold),
                         style::Print(format!(" ● Completed in {}s", tool_time)),
                         style::SetForegroundColor(Color::Reset),
-                        style::Print("\n\n"),
                     )?;
+                    if !checkpoint_tag.is_empty() {
```
dingfeli: We can assign checkpoint_tag to an `Option<String>` and do this instead:
```rust
if let Some(tag) = checkpoint_tag {
...
}
```

And for branches where you return `String::new()` you can instead return `None`. We can save an allocation this way.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2304,6 +2360,75 @@ impl ChatSession {
...
+                            &tag,
+                            &description,
+                            history_len + 1,
+                            false,
+                            Some(tool.name.clone()),
+                        ) {
+                            debug!("Failed to create tool checkpoint: {}", e);
+                            String::new()
+                        } else {
+                            manager.tools_in_turn += 1;
```
dingfeli, line 2419: Do we ever increment this by more than one per turn? I see that we are resetting it to 0 every turn. 
If that's the case this can just be a `bool` then?

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Thanks for the advise! I do change manager later (`tools_in_turn` += 1 and `create_checkpoint` also mutates internal state)

evanliu048: Also the `take()` approach is here because I need to access both manager (mutably) and self.conversation.history() (immutably) in the same scope. Using as_mut() would cause a borrowing conflict when calling self.conversation.history().clone() inside create_checkpoint().

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Thanks! Just changed the code

crates/chat-cli/src/cli/chat/mod.rs
evanliu048: Yes we do increase this value
It's used to generate unique checkpoint tags like "3.1", "3.2" for each tool within a turn
Multiple tools can be executed within a single turn.
We need to distinguish between different tool checkpoints in the same turn

crates/chat-cli/src/cli/chat/cli/checkpoint.rs
```diff
@@ -0,0 +1,578 @@
...
+        /// Second checkpoint tag (defaults to current state)
+        #[arg(required = false)]
+        tag2: Option<String>,
+    },
+}
+
+impl CheckpointSubcommand {
+    pub async fn execute(self, os: &Os, session: &mut ChatSession) -> Result<ChatState, ChatError> {
+        // Check if in tangent mode - captures are disabled during tangent mode
+        if os
```
kensave: This is on me but we should have an experiment helper to query experiments in particular.

crates/chat-cli/src/cli/chat/cli/checkpoint.rs
```diff
@@ -0,0 +1,578 @@
...
+            .database
+            .settings
+            .get_bool(Setting::EnabledTangentMode)
+            .unwrap_or(false)
+        {
+            execute!(
+                session.stderr,
+                style::SetForegroundColor(Color::Yellow),
+                style::Print(
+                    "⚠️ Checkpoint is disabled while in tangent mode. Disable tangent mode with: q settings -d chat.enableTangentMode.\n\n"
```
kensave: Why disable checkpoint if tangent mode is enabled? I understand disabling it while in Tangent mode, but not following why we can't have both features enabled at same time.

crates/chat-cli/src/cli/chat/cli/experiment.rs
```diff
@@ -50,6 +50,11 @@ static AVAILABLE_EXPERIMENTS: &[Experiment] = &[
         description: "Enables Q to create todo lists that can be viewed and managed using /todos",
         setting_key: Setting::EnabledTodoList,
     },
+    Experiment {
+        name: "Checkpoint",
+        description: "Enables workspace checkpoints to snapshot, list, expand, diff, and restore files (/checkpoint)",
+        setting_key: Setting::EnabledCheckpoint,
```
kensave, line 60: If there is a dependency with Tangent mode, probably we need to explain here.

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -876,6 +883,20 @@ Return only the JSON configuration, no additional text.",
         self.transcript.push_back(message);
     }
 
+    /// Restore conversation from a checkpoint's history snapshot
+    pub fn restore_to_checkpoint(&mut self, checkpoint: &Checkpoint) -> Result<(), eyre::Report> {
+        // 1. Restore history from snapshot
+        self.history = checkpoint.history_snapshot.clone();
```
kensave, line 904: Just curious? Can a checkpoint corrupt conversation history by any means? Like a tool use without tool results and so on?

evanliu048: The checkpoint is created after the tool executes successfully and in handle_response, so that should be fine. Thanks for the callout — we can test it more thoroughly during the bug bash

kensave (approved):

crates/chat-cli/src/cli/chat/cli/experiment.rs
evanliu048: sure. explanation added ✅

brandonskiser (approved):

crates/chat-cli/src/cli/chat/cli/checkpoint.rs
```diff
@@ -0,0 +1,573 @@
...
+            });
+        }
+
+        // Check if in tangent mode - captures are disabled during tangent mode
+        if session.conversation.is_in_tangent_mode() {
+            execute!(
+                session.stderr,
+                style::SetForegroundColor(Color::Yellow),
+                style::Print(
+                    "⚠️ Checkpoint is disabled while in tangent mode. Disable tangent mode with: q settings -d chat.enableTangentMode.\n\n"
```
kkashilk, line 110: I think user just needs to exit tangent mode, right? Doesn't need to disable it completely?

evanliu048: yes. thanks for catching this. I will change the word here

PR #2898, opened 2025-09-16
feat: Add MCP elicitation support

joshrutkowski:
*Issue #, if available:*

*Description of changes:*
Updates MCP client support for elicitation feature per MCP spec: https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation

### DIFF
--- a/crates/chat-cli/src/cli/chat/server_messenger.rs
+++ b/crates/chat-cli/src/cli/chat/server_messenger.rs
@@ -1,4 +1,12 @@
+use crossterm::style::Color;
+use crossterm::{
+    execute,
+    style,
+};
 use rmcp::model::{
+    CreateElicitationRequestParam,
+    CreateElicitationResult,
+    ElicitationAction,
     ListPromptsResult,
     ListResourceTemplatesResult,
     ListResourcesResult,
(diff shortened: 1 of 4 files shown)


crates/chat-cli/src/cli/chat/server_messenger.rs
```diff
@@ -179,6 +187,90 @@ impl Messenger for ServerMessenger {
         });
     }
 
+    async fn handle_elicitation_request(
```
dingfeli: This is a server driven event right? We have no control over when this request would be made. 

I see we are printing directly to output and stdout as a response. Would this not potentially clobber whatever is being displayed at the time?

dingfeli: The role of the messenger here is to convey to messages to the orchestrator task that takes centralized control over what gets performed: https://github.com/aws/amazon-q-developer-cli/blob/37a77e0a6246dbf0ca5e71d16683a9969daa86d7/crates/chat-cli/src/cli/chat/tool_manager.rs#L1254

I think the proper logic flow would be `server request` -> `service handler` -> `messenger` -> `orchestrator`. The orchestrator would then need to handle this. 

Right now it looks like all this method does is returning an empty hashmap.

joshrutkowski: Good callout - addressing this to move out of the messenger

PR #2925, opened 2025-09-18
fix(mcp): oauth issues

dingfeli:
*Issue #, if available:*

*Description of changes:*
This PR addresses the following:
- Headers arg for http transport not being ingested (this was previously not added for P0) 
- Exposes scopes field for oauth. This provides a way for users to circumvent issues where sometimes oauth would fail if invalid scopes are not handled gracefully by auth provider. 


crates/chat-cli/src/cli/chat/tools/custom_tool.rs
```diff
@@ -43,17 +46,20 @@ impl Default for TransportType {
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug, Eq, PartialEq, JsonSchema)]
+#[serde(tag = "type", rename_all = "camelCase", deny_unknown_fields)]
```
brandonskiser: What is `tag = "type"` doing for the struct? It seems that this should be an enum for either stdio or HTTPStreamable config

dingfeli: Ahh good catch. This was left from when I was trying to make this config into an enum by transport type (the first commit in this). I decided against it because there is no easy way to do this without making type an optional field defaulting to stdio (I would need a custom deser and I just don't think the extra maintenance that comes with it is worth it).

PR #2926, opened 2025-09-18
Use absolute path for permission management for fs_read

erbenmo:
Problem: users are getting prompted to approve fs_read even though they are reading files under CWD (through relative path). and CWD is already added into allowed path.

The root cause is a latent software bug: If an allowed path is in aboslute format, and user is reading the same file in relative format, the file path would not actually match.

To fix the issue, I am now converting the relative path of the accessed file to absolute path, and only use absolute path for permission management.

I considered an alternative approach which is to add './' to allowed path. This would match all files in CWD through relative path. But The problem with that approach is it would also match file acces like './../../../super-secret' etc which is outside CWD.

Another way this latent bug could manifest is if user add an absolute path to deny list, it doesn't actually stop the CLI to access that file through relative path... That issue is also fixed now by converting the accessed file to absolute path.

Tested the change:

```
🛠️  Using tool: fs_read (trusted)
 ⋮ 
 ● Reading file: ./crates/chat-cli/src/util/directories.rs, from line 519 to 540
 ✓ Successfully read 939 bytes from ./crates/chat-cli/src/util/directories.rs

 ⋮ 
 ● Completed in 0.1s
```

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/fs_read.rs
+++ b/crates/chat-cli/src/cli/chat/tools/fs_read.rs
@@ -181,7 +181,8 @@ impl FsRead {
                             FsReadOperation::Line(FsLine { path, .. })
                             | FsReadOperation::Directory(FsDirectory { path, .. })
                             | FsReadOperation::Search(FsSearch { path, .. }) => {
-                                let Ok(path) = directories::canonicalizes_path(os, path) else {
+                                // always use absolute file path for permission check
+                                let Ok(path) = directories::canonicalizes_path_absolute(os, path) else {
                                     ask = true;
                                     continue;
                                 };
@@ -208,7 +209,8 @@ impl FsRead {
                                 let denied_match_set = paths
                                     .iter()
                                     .flat_map(|path| {
-                                        let Ok(path) = directories::canonicalizes_path(os, path) else {
+                                        // always use absolute file path for permission check
+                                        let Ok(path) = directories::canonicalizes_path_absolute(os, path) else {
                                             return vec![];
                                         };
                                         deny_set.matches(path.as_ref() as &str)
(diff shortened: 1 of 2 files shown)


erbenmo:
I noticed an earlier PR is trying to address the same issue. That PR is a more comprehensive fix than this one. https://github.com/aws/amazon-q-developer-cli/pull/2820. I think we should just take that one instead.

erbenmo:
Or we can merge my PR first to fix the CWD issue and I can work with the author of https://github.com/aws/amazon-q-developer-cli/pull/2820 to get his/her more comprehensive change added. That PR is more comprehensive because:
- it also handle the relative glob pattern
- it also apply the fix to fs_write

crates/chat-cli/src/cli/chat/tools/fs_read.rs
```diff
@@ -1537,4 +1541,52 @@ mod tests {
...
+            name: "test_agent".to_string(),
+            tools_settings: HashMap::new(), // No fs_read settings
+            ..Default::default()
+        };
+        
+        // Save original CWD and change to temp directory
+        let original_cwd = std::env::current_dir().unwrap();
+        std::env::set_current_dir(temp_path).unwrap();
+
+        // all of these files are under CWD through relative path. reading them should be allowed by default
```
mr-lee, line 1574: you want to test the `~` just in case?

mr-lee (approved):

crates/chat-cli/src/cli/chat/tools/fs_read.rs
erbenmo: @mr-lee I noticed that https://github.com/aws/amazon-q-developer-cli/pull/2820 is a better fix and the fix for this issue has now been merged into https://github.com/aws/amazon-q-developer-cli/pull/2933 (which fixed a bug in #2820 found by kkashilk@) . So I am going to discard this one. Sorry about the confusion.

erbenmo:
discarding this PR in favor of #2933

PR #2944, opened 2025-09-19
Fix dotfile indexing in knowledge base

aaronmarkham:
*Description of changes:*

- Treat dotfiles as text files instead of unknown file types
- Allow common config files (.bashrc, .vimrc, .raprc, etc.) through hidden file filters
- Add test cases for dotfile patterns (.bashrc, .vimrc, .raprc, .testrc, etc.)

Previously, dotfiles were classified as FileType::Unknown and skipped during directory processing, preventing configuration files from being indexed.

This change enables indexing of:
- Files ending in 'rc' (.bashrc, .vimrc, .raprc)
- Files ending in 'config' (.gitconfig, .npmconfig)
- Common dotfiles (.gitignore, .env, .dockerignore)

Fixes issue where configuration files showed 0 items after indexing.

Tested with .raprc and .testrc files - both now properly index content.

### DIFF
--- a/crates/semantic-search-client/src/client/background/file_processor.rs
+++ b/crates/semantic-search-client/src/client/background/file_processor.rs
@@ -45,10 +45,19 @@ impl FileProcessor {
                 .filter_map(|e| e.ok())
                 .filter(|e| e.file_type().is_file())
                 .filter(|e| {
-                    !e.path()
-                        .file_name()
-                        .and_then(|n| n.to_str())
-                        .is_some_and(|s| s.starts_with('.'))
+                    // Skip hidden files, but allow common configuration files
+                    if let Some(filename) = e.path().file_name().and_then(|n| n.to_str()) {
+                        if filename.starts_with('.') {
+                            // Allow common configuration files
+                            filename.ends_with("rc") 
+                                || filename.ends_with("config")
+                                || matches!(filename, ".gitignore" | ".env" | ".dockerignore")
+                        } else {
+                            true
+                        }
+                    } else {
+                        true
+                    }
                 })
                 .filter(|e| {
                     pattern_filter
--- a/crates/semantic-search-client/src/client/implementation.rs
+++ b/crates/semantic-search-client/src/client/implementation.rs
@@ -430,13 +430,18 @@ impl SemanticSearchClient {
         {
             let path = entry.path();
 
-            // Skip hidden files
-            if path
-                .file_name()
-                .and_then(|n| n.to_str())
-                .is_some_and(|s| s.starts_with('.'))
-            {
-                continue;
+            // Skip hidden files, but allow common configuration files
+            if let Some(filename) = path.file_name().and_then(|n| n.to_str()) {
+                if filename.starts_with('.') {
+                    // Allow common configuration files
+                    let is_config_file = filename.ends_with("rc") 
+                        || filename.ends_with("config")
+                        || matches!(filename, ".gitignore" | ".env" | ".dockerignore");
+                    
+                    if !is_config_file {
+                        continue;
+                    }
+                }
             }
 
             // Process the file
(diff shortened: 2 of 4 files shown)


crates/semantic-search-client/src/client/background/file_processor.rs
```diff
@@ -45,10 +45,19 @@ impl FileProcessor {
...
-                        .file_name()
-                        .and_then(|n| n.to_str())
-                        .is_some_and(|s| s.starts_with('.'))
+                    // Skip hidden files, but allow common configuration files
+                    if let Some(filename) = e.path().file_name().and_then(|n| n.to_str()) {
+                        if filename.starts_with('.') {
+                            // Allow common configuration files
+                            filename.ends_with("rc") 
+                                || filename.ends_with("config")
+                                || matches!(filename, ".gitignore" | ".env" | ".dockerignore")
```
kensave, line 54: Can we make this a CONSTANT so we we can use this collection in both places.

crates/semantic-search-client/src/client/implementation.rs
```diff
@@ -430,13 +430,18 @@ impl SemanticSearchClient {
...
-                .file_name()
-                .and_then(|n| n.to_str())
-                .is_some_and(|s| s.starts_with('.'))
-            {
-                continue;
+            // Skip hidden files, but allow common configuration files
+            if let Some(filename) = path.file_name().and_then(|n| n.to_str()) {
+                if filename.starts_with('.') {
+                    // Allow common configuration files
+                    let is_config_file = filename.ends_with("rc")
```
kensave, line 437: Constant or utility for this?

pgpbpadilla:
I ran into this issue looking for support for .org files (Emacs Org Mode) which are also plain text. 

Any chance /knowledge can process any text file regardless of the extension?

crates/semantic-search-client/src/client/background/file_processor.rs
pgpbpadilla, line 54: Why hard-code file extensions here? I think a better approach would be to extract the actual list of allowed file format/type/extensions into a configuration file. 

Are there any documented reasons not to support any plain text file type/extension?

crates/semantic-search-client/src/client/implementation.rs
```diff
@@ -430,13 +430,18 @@ impl SemanticSearchClient {
...
-                .is_some_and(|s| s.starts_with('.'))
-            {
-                continue;
+            // Skip hidden files, but allow common configuration files
+            if let Some(filename) = path.file_name().and_then(|n| n.to_str()) {
+                if filename.starts_with('.') {
+                    // Allow common configuration files
+                    let is_config_file = filename.ends_with("rc") 
+                        || filename.ends_with("config")
+                        || matches!(filename, ".gitignore" | ".env" | ".dockerignore");
```
pgpbpadilla, line 439: similarly here, I think a better approach would be to extract this to a configuration file instead of hardcoding each file type/extension

crates/semantic-search-client/src/processing/file_processor.rs
```diff
@@ -256,6 +264,13 @@ mod tests {
             ("Dockerfile", FileType::Text),
             ("LICENSE", FileType::Text),
             (".gitignore", FileType::Text),
+            (".bashrc", FileType::Text),
```
pgpbpadilla, line 267: why not just write a single test that is valid for any plain text file? this approach of adding each file type/extension one by one  is bound to require unnecessary maintenance for every new file type/extension that people want to use.

crates/semantic-search-client/src/client/background/file_processor.rs
kensave: You are right Pablo, we wanted to have guardrails at least on the supported extensions. But it can be definitely explored.

Issue #2961, opened 2025-09-22
Feature Request: Add File Injection Capability (@ Command for File Reference)


zy8848:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.6.0 (24G84)

### Requested Feature:
Add file injection capability using `@filepath` syntax to automatically inject the content of specified files into prompts. This is useful for asking questions about specific code, text, or collections of files without manual copy-pasting. This feature is already available in both Claude Code and Gemini CLI through their respective @ command implementations.
Examples:
- `Explain this text @path/to/your/file.txt `
- `Summarize the code in this directory. @src/my_project/ `
- `What is this file about? @README.md`

### Expected Behavior

Referring to [gemini-cli](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#at-commands-), when a user enters a prompt containing @filepath patterns(e.g. `Compare the file @package.json and @src/config.ts`), the processed prompt looks like:
```
Compare the file @package.json and @src/config.ts

--- Content from referenced files ---

Content from @package.json:
{
  "name": "my-project",
  "version": "1.0.0"
}

Content from @src/config.ts:
export const config = {
  port: 3000
};

--- End of content ---
```

### Actual behaviour

  Currently, when users want to reference a file in chat, they must:
  - Manually copy and paste file path into their prompts
  - Wait for Q-CLI to identify their intent and extract file paths
  - Wait for Q-CLI to use fs_read tool calls to retrieve file content

### Environment

```yaml
[q-details]
version = "1.16.2"
variant = "full"
[system-info]
os = "macOS 15.6.0 (24G84)"
chip = "Apple M2 Pro"
total-cores = 12
memory = "32.00 GB"
```

bergjaak:
you can press ctrl + s to use fuzzy find to look for `context add` and then fuzzy find is uses to look up files, similar to @ in Claude Code and Cursor
but it would be an improvement to have @ automatically trigger this fuzzy find, similar to Claude Code

PR #2967, opened 2025-09-22
feat: support @filepath syntax for injecting file content into prompts

zy8848:
*Issue #, if available:*
Issue #2961 

*Description of changes:*

Add file injection support for @ command

This change introduces the ability to reference and inject file contents using the @filepath syntax in chat commands. When users include @filepath in their input, the q cli automatically reads and appends the referenced file content to the message.

Key changes:
- Added inject_at_files() method to detect @filepath patterns using regex
- Added format_file_content() helper to read and format file contents
- File contents are appended with delimiters

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -3393,11 +3393,7 @@ impl ChatSession {
     async fn format_file_content(&self, path: &std::path::Path) -> Result<String, std::io::Error> {
         let content = tokio::fs::read_to_string(path).await?;
 
-        let formatted_content = format!(
-            "Content from @{}:\n{}",
-            path.to_string_lossy(),
-            content.trim()
-        );
+        let formatted_content = format!("Content from @{}:\n{}", path.to_string_lossy(), content.trim());
 
         Ok(formatted_content)
     }
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -1933,6 +1933,11 @@ impl ChatSession {
     async fn handle_input(&mut self, os: &mut Os, mut user_input: String) -> Result<ChatState, ChatError> {
         queue!(self.stderr, style::Print('\n'))?;
         user_input = sanitize_unicode_tags(&user_input);
+
+        if let Some(injected_input) = self.inject_at_files(user_input.trim()).await {
+            user_input = injected_input;
+        }
+
         let input = user_input.trim();
 
         // handle image path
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3350,6 +3355,48 @@ impl ChatSession {
                 .content,
         })
     }
```
kensave, line 3570: Thanks for submitting your PR, a couple of comments:

* @ already conflicts with some other injection mechanisms like prompts where syntax works like `@prompt arg1 arg2` and .
* The current mechanism doesn't have a way to sanitize path, see `sanitize_path_tool_arg` as a sample. And would also have broadly matches like `../../../etc/passwd`
* Current regest would capture e-mails too.

We already have slash commands to include context files. To avoid conflicts we could also have more explicit mechanisms such as `@file:path` What do you think about that?

zy8848: ### Regarding point 1&3 about @ symbol conflicts: 
I understand the concern, but my implementation doesn’t conflict with @prompt functionality. This is because the file injection mechanism only processes @ patterns that correspond to actual existing files on the filesystem (the checking logic is adapted from [gemini-cli](https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/ui/hooks/atCommandProcessor.ts#L196)). The logic is:

1. It uses regex `@([^\s]+)` to find potential file paths
2. **Crucially**, it only processes matches where `path.exists() && path.is_file()` returns true
3. If the path after @ doesn't exist as a file, it's simply ignored and left unchanged

This means:
- `@prompt arg1 arg2` → "prompt" isn't a file, so it's left untouched
- `@README.md` → if README.md exists, it gets processed
- `@nonexistent.txt` → ignored, left as-is
- `user_email@email.com` → "email.com" isn't a file, so ignored

The file injection runs as a preprocessing step before other @ mechanisms, but it only consumes @ patterns that are actual files, leaving all other @ syntax intact for downstream processing.

Points 2: I'll integrate `sanitize_path_tool_arg` for security.

I agree with using `@file:path` syntax for better clarity. Should I update to use that format?

### Regarding the conflict with `/context` command:

While both features involve files, they serve completely different use cases:

- **Context commands**: For persistent files that stay in context across conversations (design docs, knowledge base, etc.)
- **File injection**: For one-time, per-message file inclusion (editing specific files, explaining code snippets, etc.). I believe this is a feature developers are likely to use frequently in day-to-day workflows.

The injection mechanism is closer to “attach this file to my current message” rather than “add this to my persistent context.” Each conversation turn can include different injected files without overloading the persistent context.

kensave: If we are able to make it work so we avoid triggering @path on cases where not required, then would be easier from user perspective, one less thing for users to add to muscle memory. Also Tab file name auto-complete is already working for @.

One nit is that we could hightlight the text when there's a match so user can visually perceive that.

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3350,6 +3355,48 @@ impl ChatSession {
...
+
+    async fn inject_at_files(&self, input: &str) -> Option<String> {
```
kensave: nti: I'd move this to its own file file_injection.rs

crates/chat-cli/src/cli/chat/mod.rs
zy8848: In this new commit, I moved the `inject_at_files` function to `file_injection.rs` and highlighted the `@file` text when the file exists. Please check it out, thanks!

kensave: The highlight seems flaky when you are typing, it's necessary to delete a character so it does nightlight autocomplete highlight works pretty good.

PR #2986, opened 2025-09-25
[fix] Fixes issues with Tool Input parsing.

kensave:
####Summary
* Ocassionally the model will generate a tool use which parameters are not a valid json. When this happens it corrupts the conversation history.
* Here we first avoid storing the tool use and add the proper validation logic to the conversation history.

*Issue #, if available:*

*Description of changes:*

This makes sure not corrupted messages are stored into conversation history, by verifying format of the response. Also, adds the validation message for clear user understanding of the error and traceability in conversation history of the model.

Screenshot 2025-09-24 at 5 01 17 PM


crates/chat-cli/src/cli/chat/parser.rs
```diff
@@ -472,7 +472,20 @@ impl ResponseParser {
...
 
         let args = match serde_json::from_str(&tool_string) {
-            Ok(args) => args,
+            Ok(args) => {
+                // Ensure we have a valid JSON object
+                match args {
+                    serde_json::Value::Object(_) => args,
+                    _ => {
+                        error!("Received non-object JSON for tool arguments: {:?}", args);
+                        serde_json::json!({
```
brandonskiser: don't think this is necessary? If it's not an object, then will the below code that checks for MCP tools ever get ran? Otherwise parser will only return valid Objects despite there actually being an error

kensave: This is to avoid the object getting stored in conversation. So string can be parsed correctly, but it might not be a `serde_json::Value::Object`. Here I wanted two things to happen, avoid corrupting conversation history and provide proper validation.

kensave: Bellow code will check for orig arguments.

brandonskiser: but args will always be equal to orig_args, unless I'm missing something

kensave: I see what you mean. true

kensave: Avoiding actually, original values, that could have more side effects.

brandonskiser (approved):

dingfeli (approved):

PR #2987, opened 2025-09-25
feat: Typing "quit" or "exit" without slash also terminates session

jbnunn:
*Issue #, if available:*
None

*Description of changes:*
Added support for exit and quit commands (without the slash) to terminate the chat session. Keeps feature parity with other agents. Saves expensive call to model when users intends to quit the session but doesn't use the slash.

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -2031,6 +2031,8 @@ impl ChatSession {
             Ok(ChatState::PromptUser {
                 skip_printing_tools: false,
             })
+        } else if matches!(input.to_lowercase().as_str(), "exit" | "quit") {
+            return Ok(ChatState::Exit);
         } else if let Some(command) = input.strip_prefix("@") {
             let input_parts =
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2031,6 +2031,8 @@ impl ChatSession {
             Ok(ChatState::PromptUser {
                 skip_printing_tools: false,
             })
+        } else if matches!(input.to_lowercase().as_str(), "exit" | "quit") {
+            return Ok(ChatState::Exit);
```
kensave, line 2035: I have second thoughts about this one as user could legitimately want to use this as a prompt.

jbnunn: That's a possibility. It feels like "exit" though would be a stronger indicator than "quit" that the user wanted to end the session. I can update the PR to only look for "exit" if you agree. 

For "quit", could we look at data to verify? For example, what's the percentage of chats that have "quit" followed by "/quit" or "/exit"?

PR #2999, opened 2025-09-26
Fix bugs in execute_bash permission control

erbenmo:
Customer wants to allow some commands and deny all other commands.

To do that, today customer needs to add commands to allow list and use a negative regex in deny list to exclude other commands. That actually doesn't work.

The first issue is deny list regex we use in Rust doesn't work for such negative look around. When this happens the deny list doesn't match anything. The first change is to fallback deny list to "*" when regex compilation fails, instead of ignoring that failure.

The second issue is customer just want an easy way to deny by default. The 2nd change is to introduce a boolean flag called "denyByDefault" so anything not in allowed list is denied by default (instead of "ask" by default)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
+++ b/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
@@ -195,6 +195,8 @@ impl ExecuteCommand {
             allowed_commands: Vec<String>,
             #[serde(default)]
             denied_commands: Vec<String>,
+            #[serde(default)]
+            deny_by_default: bool,
             #[serde(default = "default_allow_read_only")]
             auto_allow_readonly: bool,
         }
@@ -211,6 +213,7 @@ impl ExecuteCommand {
                 let Settings {
                     allowed_commands,
                     denied_commands,
+                    deny_by_default,
                     auto_allow_readonly,
                 } = match serde_json::from_value::<Settings>(settings.clone()) {
                     Ok(settings) => settings,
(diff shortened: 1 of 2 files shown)


mfekadu (approved):
As _the customer_, I do wish the lookahead/lookbehinds could be supported but I accept the limitation given that `deny_by_default` is a reasonable workaround! 

The unit tests correctly show the [principle of least privilege](https://en.m.wikipedia.org/wiki/Principle_of_least_privilege) applied to `toolSettings` of `execute_bash`. Specifically, granting `ls` does not implicitly allow `pwd` (or anything more consequential). 

Thanks for this excellent security improving PR! 🚀

crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -195,6 +195,8 @@ impl ExecuteCommand {
             allowed_commands: Vec<String>,
             #[serde(default)]
             denied_commands: Vec<String>,
+            #[serde(default)]
+            deny_by_default: bool,
```
kensave, line 199: Thanks for raising this! Another approach is configure the behavior to either : `ask` or just `deny`, being the default behaviour ask, although this accomplishes same.

erbenmo: Yeah I didn't do this because the default setting for a tool is to `ask`.  so I am only adding `deny` support here. technically I can add `approve` to be default but there are other ways to do that already.

brandonskiser (approved):

PR #3017, opened 2025-09-27
fix: treat empty string as unset for chat.defaultAgent setting

mibeco:
When users run 'q settings chat.defaultAgent ""' to reset their default agent (as documented), the empty string was being treated as a valid agent name, causing an error message on every chat session start.

This change treats empty strings as 'no default set', allowing users to cleanly reset to the built-in default agent without error messages.

Fixes the misleading behavior described in the AWS documentation where setting an empty string should reset to built-in default silently. (The documentation is also being revised, but we should take care of this edge case anyway.)

Includes test to verify empty string handling behavior.

*Issue #, if available:*

*Description of changes:*


dingfeli:
But would user not just delete `chat.defaultAgent` if they don't want a default set?

PR #3018, opened 2025-09-27
fix(prompts): Allow MCP prompts with no arguments

jlhood:
*Issue #, if available:* n/a

*Description of changes:*

Allow MCP prompts with no arguments

- Fix argument processing logic in tool_manager.rs to handle prompts with optional/no arguments
- Add validation for required arguments before sending to MCP server
- Add MissingRequiredArguments error type for better error messages
- Update error handling in prompts.rs to display helpful messages
- Add comprehensive tests to verify the fix works correctly

Fixes issue where prompts like 'simple-greeting' with no required arguments would fail with 'Missing required arguments' error when called as @simple-greeting without any parameters.

### DIFF
--- a/crates/chat-cli/src/cli/chat/tool_manager.rs
+++ b/crates/chat-cli/src/cli/chat/tool_manager.rs
@@ -1080,22 +1080,42 @@ impl ToolManager {
                     let server_name = &bundle.server_name;
                     let client = self.clients.get_mut(server_name).ok_or(GetPromptError::MissingClient)?;
                     let PromptBundle { prompt_get, .. } = bundle;
-                    let arguments = if let (Some(schema), Some(value)) = (&prompt_get.arguments, &arguments) {
-                        let params = schema.iter().zip(value.iter()).fold(
-                            HashMap::<String, String>::new(),
-                            |mut acc, (prompt_get_arg, value)| {
-                                acc.insert(prompt_get_arg.name.clone(), value.clone());
-                                acc
-                            },
-                        );
-                        Some(
-                            params
-                                .into_iter()
-                                .map(|(k, v)| (k, serde_json::Value::String(v)))
-                                .collect(),
-                        )
-                    } else {
-                        None
+
+                    // Validate required arguments before processing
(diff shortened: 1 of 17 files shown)


dingfeli:
Thanks for doing this, James. 
Prompts can already be used without arguments. 
And error propagation from incorrect prompt fetch is also delegated to the mcp servers:
image

I have put up a PR for fixing tests that were invalidated from type changes that came with the upgrade of rmcp: https://github.com/aws/amazon-q-developer-cli/pull/3037

jlhood:
I just pulled latest main and I still see the bug:

Latest main branch:
Screenshot 2025-09-29 at 11 26 29

With my change:
Screenshot 2025-09-29 at 11 26 40

crates/chat-cli/src/cli/chat/tool_manager.rs
```diff
@@ -1080,22 +1080,42 @@ impl ToolManager {
...
+
+                        let provided_count = arguments.as_ref().map_or(0, |args| args.len());
+
+                        if !required_args.is_empty() && provided_count < required_args.len() {
+                            return Err(GetPromptError::MissingRequiredArguments {
+                                prompt_name: prompt_name.clone(),
+                                required_args: required_args.iter().map(|arg| arg.name.clone()).collect(),
+                            });
+                        }
+                    }
```
dingfeli: We don't need this. We should just let the error happen where the source of truth is, the servers.

crates/chat-cli/src/cli/chat/tool_manager.rs
```diff
@@ -2158,4 +2180,142 @@ mod tests {
         assert_eq!(params.name, "test-prompt"); // Not "example-server/test-prompt"
         assert_eq!(server_name, Some("example-server".to_string()));
     }
+
+    #[test]
+    fn test_argument_processing_logic() {
```
dingfeli: These tests are not really testing anything in the code path introduced. 

If we want to validate that the logic introduced to derive the arguments, we might want to put them in a function and #[inline] them.

dingfeli (approved):

PR #3020, opened 2025-09-27
fix:bug(prompts): fix error when truncating prompt description written in multibyte characters

h16rkim:
Issue #, if available: https://github.com/aws/amazon-q-developer-cli/issues/3117

## *Error Screenshot*

image

<br />

## *Description of changes:*

~This issue does not occur in the currently released version(1.16.2). However, **it can be reproduced on the current main branch**.~ **This error is occuring in the current released version (1.17.1)**

When executing the `/prompts list` command, there is a [function](https://github.com/h16rkim/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/cli/prompts.rs#L240-L251) that trims the description if it is too long. **However, if the trim happens at the boundary of a multibyte character, a panic occurs and the program terminates.**

This PR fixes an issue that may occur in multibyte language systems such as Korean, Japanese, Chinese, Vietnamese, Thai, Arabic, emojis, etc..

<br />

## *After Applying this PR (Screenshot)*

image

<br />


crates/chat-cli/src/cli/chat/tool_manager.rs
```diff
@@ -2112,7 +2112,9 @@ mod tests {
         // Create mock prompt bundles
         let prompt = rmcp::model::Prompt {
             name: "test_prompt".to_string(),
+            title: Some("Test Prompt".to_string()),
             description: Some("Test description".to_string()),
+            icons: None,
```
h16rkim, line 2117: This change is not related to this PR, but I fixed it because a syntax error was occurring (test data modification).

crates/chat-cli/src/cli/chat/util/mod.rs
```diff
@@ -12,10 +12,42 @@ use aws_smithy_types::{
...
+    UnicodeWidthStr,
+};
 
 use super::ChatError;
 use super::token_counter::TokenCounter;
 
+/// Truncates text to a maximum display width.
+/// Takes into account the display width of characters, where CJK characters are counted as 2 units
+/// wide. Returns the truncated text without ellipsis.
+pub fn truncate_safe_with_width(text: &str, max_width: usize) -> String {
```
h16rkim, line 25: In languages such as CJK, the width of a single character is twice of an alphabet letter. The code has been improved to correctly calculate the width even when the description is written in a CJK language.

image

PR #3024, opened 2025-09-29
feat: Add file:// URI support for agent prompts

rationull:
- Add file_uri utility module for resolving file:// URIs
- Support both relative and absolute file paths in agent prompts
- Resolve file URIs relative to agent config directory
- Update agent format documentation with file URI examples

The aim here is to improve the user experience when using longer, more complex prompts in custom agent configs.

(built with Q CLI)

*Issue #, if available:* 2802

*Description of changes:*

(See commit message). Additional context:

- I didn't file the original issue. This is my take on a good solution but if it doesn't align with the team's preferred direction then I won't be offended if the PR is canceled :)
- This is my first time using Rust; of course I am very open to feedback and learning!

This adds a new `file_uri` utility module. My original thought was that ideally this would be combined with existing `file://` URI processing (e.g. for "resources" in agent config). But, the behaviors are (should be) slightly different:
  - Glob matches don't make sense for prompt (unless multi-file concatenated prompts are a valid use case? Seems like a stretch to me and globs would be a bad way to implement that anyway due lack of fine control of ordering).
  - Relative paths are resolved against the user's working dir for resources, but for prompt they should be resolved against the agent config .json file's directory.

So I opted to go with a new module and avoid modifying any existing functionality. I'm happy to revise this (either here or in a followup PR) based on feedback.

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -69,6 +69,7 @@ use crate::util::{
     self,
     MCP_SERVER_TOOL_DELIMITER,
     directories,
+    file_uri,
 };
 
 pub const DEFAULT_AGENT_NAME: &str = "q_cli_default";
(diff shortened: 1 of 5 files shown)


kensave:
Thanks for your contribution,

Can you please the scenarios you used to test this?

rationull:
For manual testing, I used the following agent config and prompt file, and:

- Moved the prompt file to different paths, testing with absolute and relative paths.
- Used prompts like "tell me about butter" with Q CLI 1.16.2 and with my local build.
   - With 1.16.2 Q would see the prompt and try to the `butter.md` before answering.
   - With my local built Q would provide various answers about butter, adhering to the instructions in the prompt.

[butter.json](https://github.com/user-attachments/files/22624277/butter.json)
[butter.md](https://github.com/user-attachments/files/22624279/butter.md)

Issue #3034, opened 2025-09-29
Allow postToolUse Hook to run after the Agent is finished (instead of after all usages)


hugoncosta:
Hello,

I've been playing with the postToolUse (introduced in #2875) hook a bit, I think it's a good place for us to automatically run linters/tests, but as it sits, it will run every single time the Agent writes something. As you know, sometimes the Agent writes the whole file in a single shot, sometimes it just adds an import statement. Running linter/test every single time we write one line is going to take longer than what we'd save with a "run linter/test" prompt after it's done.

My proposal would be to allow us to run postToolUse when the Agent is ready to hand over the control to us. This would allow it to do everything it wants, and then we'd run the linter/tests, feeding the output (if wrong) back to the LLM. 

Let me know your thoughts on it

PR #3036, opened 2025-09-29
feat: Allow users to run postToolUse hooks only after a complete turn

hugoncosta:
*Issue #, if available:* #3034

*Description of changes:*
See #2875 for the initial implementation. Sometimes, the Agents run the tools several times in a particular turn (e.g. when writing code, they don't write the whole thing at once, especially when dealing with already existing files).

With this feature, you can set your hook to only run at the end of the turn, allowing you to only run e.g. formatting/tests when the whole file is written.

Tested locally, it's working as intended - only after doing all the fs_write did it do the formatting (I used the example currently in the hooks.md that runs `cargo fmt` after it's done).

### DIFF
--- a/crates/chat-cli/src/cli/agent/hook.rs
+++ b/crates/chat-cli/src/cli/agent/hook.rs
@@ -68,7 +68,7 @@ pub struct Hook {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub matcher: Option<String>,
 
-    /// If true, PostToolUse hooks are deferred until turn completion
+    /// If true, the hook is deferred until turn completion
     #[serde(default)]
     pub only_when_turn_complete: bool,
 
--- a/crates/chat-cli/src/cli/agent/hook.rs
+++ b/crates/chat-cli/src/cli/agent/hook.rs
@@ -68,6 +68,10 @@ pub struct Hook {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub matcher: Option<String>,
 
+    /// If true, PostToolUse hooks are deferred until turn completion
+    #[serde(default = "Hook::default_only_when_turn_complete")]
+    pub only_when_turn_complete: bool,
+
     #[schemars(skip)]
     #[serde(default, skip_serializing)]
     pub source: Source,
@@ -81,6 +85,7 @@ impl Hook {
             max_output_size: Self::default_max_output_size(),
             cache_ttl_seconds: Self::default_cache_ttl_seconds(),
             matcher: None,
+            only_when_turn_complete: Self::default_only_when_turn_complete(),
             source,
         }
     }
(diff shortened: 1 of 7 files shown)


crates/chat-cli/src/cli/agent/hook.rs
```diff
@@ -68,6 +68,10 @@ pub struct Hook {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub matcher: Option<String>,
 
+    /// If true, PostToolUse hooks are deferred until turn completion
+    #[serde(default = "Hook::default_only_when_turn_complete")]
```
erbenmo: nit: Consider just setting it to false instead of using a separate method

docs/hooks.md
```diff
@@ -133,6 +133,23 @@ Runs after tool execution with access to tool results.
...
+This is useful for formatting, cleanup tasks or for running tests that should run only once after all tools complete, 
+rather than after each individual tool.
+
+```json
+{
+  "postToolUse": [
+    {
+      "matcher": "fs_write",
+      "command": "cargo fmt --all",
+      "only_when_turn_complete": true
```
erbenmo, line 147: Maybe it would be a good idea to set cache_ttl_seconds in this case to avoid calling this multiple times?

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -4411,6 +4427,101 @@ mod tests {
...
+
+        use crate::cli::agent::hook::{
+            Hook,
+            HookTrigger,
+        };
+
+        let mut os = Os::new().await.unwrap();
+        os.client.set_mock_output(serde_json::json!([
+            [
+                "I'll create a file for you",
```
erbenmo: it seems this test doesn't actually verify if the hook is deferred or not since we are only calling fs_write tool once?

erbenmo:
Overall change looks good to me. It would be great if we can get review from the core team to verify that we are calling the deferred hooks in the right place as well!

docs/hooks.md
hugoncosta: Since the implementation is done on top of the postToolUse hook, that is enabled for those that choose to use. Personally, I think it might be better to not use it, as some turns take 10 seconds, some take 10 minutes.

docs/agent-format.md
```diff
@@ -286,7 +286,8 @@ For detailed information about hook behavior, input/output formats, and examples
     "postToolUse": [
       {
         "matcher": "fs_write",
-        "command": "cargo fmt --all"
+        "command": "cargo fmt --all",
+        "only_when_turn_complete": true
```
dk19y, line 290: `onlyWhenTurnComplete` -> Can you please  Camel case consistent with rest of the config ?

dk19y (changes requested):

hugoncosta:
Closing as we decided to go with a simpler approach featured in #3070

PR #3042, opened 2025-09-29
fix: make q settings delete functionality user friendly

nirajchowdhary:
*Issue #, if available:*: N/A

*Description of changes:*
* When you user `q settings --help`, the delete option help says "Delete a value" when it should be `delete a key` as delete does not take value.
* Going by Usage as mentioned in `q settings --help`, when you type out `q settings -d chat.defaultAgent "some-value"` The error message shown is confusing and contradictory. It prints out two contradicting statement. As seen below 
```
dev-dsk-fast-chownira-2c-402b637c % q settings --delete chat.defaultAgent my-preferred-agent
error: the argument '--delete' cannot be used with '[VALUE]'

Usage: q settings --delete <KEY> [VALUE] <--- this contradicts above

For more information, try '--help'.

```
* Fixed both the above problems, now by mistake if user uses value in delete , it will print a user friendly correcting message
```
dev-dsk-fast-chownira-2c-402b637c % cargo run --bin chat_cli -- settings -d chat.defaultAgent "some-val"
   Compiling chat_cli v1.16.3 (/local/home/chownira/AmazonQ/amazon-q-developer-cli/crates/chat-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 43.75s
     Running `target/debug/chat_cli settings -d chat.defaultAgent some-val`
error: the argument '--delete' cannot be used with '[VALUE]'
 Usage: q settings --delete chat.defaultAgent

``` 
## testing
Wrote a unit test and 
Ran
* cargo test
* cargo clippy
* cargo +nightly fmt

### DIFF
--- a/crates/chat-cli/src/cli/settings.rs
+++ b/crates/chat-cli/src/cli/settings.rs
@@ -89,7 +89,13 @@ impl SettingsArgs {
             None => {
                 let Some(key) = &self.key else {
                     if self.delete {
-                        return Err(eyre::eyre!("the argument {} requires a {}\n Usage: q settings {} {}", "'--delete'".yellow(), "<KEY>".green(), "--delete".yellow(), "<KEY>".green()));
+                        return Err(eyre::eyre!(
+                            "the argument {} requires a {}\n Usage: q settings {} {}",
+                            "'--delete'".yellow(),
+                            "<KEY>".green(),
+                            "--delete".yellow(),
+                            "<KEY>".green()
+                        ));
                     }
                     return Ok(ExitCode::SUCCESS);
                 };


crates/chat-cli/src/cli/settings.rs
```diff
@@ -92,6 +93,12 @@ impl SettingsArgs {
 
                 let key = Setting::try_from(key.as_str())?;
                 match (&self.value, self.delete) {
+                    (Some(_), true) => Err(eyre::eyre!(
+                        "the argument {} cannot be used with {}\n Usage: q settings {} {key}",
```
kkashilk, line 106: With this change, whats the behavior when you run `q settings -d` (without providing a key)

nirajchowdhary: @kkashilk  that's a good catch, when we run the same without my code Clap handles it, but it is still giving output which can be confusing, where it says [VALUE] is needed. I can handle this scenario for delete in next revision
```
error: the following required arguments were not provided:
  <KEY>

Usage: q settings --delete <KEY> [VALUE]

```

brandonskiser (approved):

PR #3054, opened 2025-09-30
feat: Adds experiment manager

kensave:
* Creates Experiment Manager
* Creates a master toggle to disable experiments.
* Replaces usage of settings in favor of ExperimentManager

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/tangent.rs
+++ b/crates/chat-cli/src/cli/chat/cli/tangent.rs
@@ -13,7 +13,10 @@ use crate::cli::chat::{
     ChatSession,
     ChatState,
 };
-use crate::database::settings::Setting;
+use crate::cli::experiment::experiment_manager::{
+    ExperimentManager,
+    ExperimentName,
+};
 use crate::os::Os;
 
 #[derive(Debug, PartialEq, Args)]
@@ -46,12 +49,7 @@ impl TangentArgs {
 
     pub async fn execute(self, os: &Os, session: &mut ChatSession) -> Result<ChatState, ChatError> {
         // Check if tangent mode is enabled
-        if !os
-            .database
-            .settings
-            .get_bool(Setting::EnabledTangentMode)
-            .unwrap_or(false)
-        {
+        if !ExperimentManager::is_enabled(os, ExperimentName::TangentMode) {
             execute!(
                 session.stderr,
                 style::SetForegroundColor(Color::Red),
@@ -66,12 +64,7 @@ impl TangentArgs {
         match self.subcommand {
             Some(TangentSubcommand::Tail) => {
                 // Check if checkpoint is enabled
-                if os
-                    .database
-                    .settings
-                    .get_bool(Setting::EnabledCheckpoint)
-                    .unwrap_or(false)
-                {
+                if ExperimentManager::is_enabled(os, ExperimentName::Checkpoint) {
                     execute!(
                         session.stderr,
                         style::SetForegroundColor(Color::Yellow),
@@ -123,12 +116,7 @@ impl TangentArgs {
                     )?;
                 } else {
                     // Check if checkpoint is enabled
-                    if os
-                        .database
-                        .settings
-                        .get_bool(Setting::EnabledCheckpoint)
-                        .unwrap_or(false)
-                    {
+                    if ExperimentManager::is_enabled(os, ExperimentName::Checkpoint) {
                         execute!(
                             session.stderr,
                             style::SetForegroundColor(Color::Yellow),
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -156,6 +156,10 @@ use crate::cli::chat::cli::prompts::{
 };
 use crate::cli::chat::message::UserMessage;
 use crate::cli::chat::util::sanitize_unicode_tags;
+use crate::cli::experiment::experiment_manager::{
+    ExperimentManager,
+    ExperimentName,
+};
 use crate::database::settings::Setting;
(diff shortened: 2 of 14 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2365,13 +2363,10 @@ impl ChatSession {
...
             // Handle checkpoint after tool execution - store tag for later display
             let checkpoint_tag: Option<String> = {
-                let enabled = os
-                    .database
-                    .settings
-                    .get_bool(Setting::EnabledCheckpoint)
-                    .unwrap_or(false)
-                    && !self.conversation.is_in_tangent_mode();
-                if invoke_result.is_err() || !enabled {
+                if invoke_result.is_err()
```
dingfeli, line 2366: Perhaps this was a question that should have been raised earlier but do we still need to check the negative conditions this way if checkpoint_manager is `Some`? Could we simplify this by just doing:
```rust
if let Some(mut manager) = self.conversation.checkpoint_manager.take() {
   ...
} else {
   None
}
```

And then perhaps if it's in tangent mode we just never assign checkpoint_manager to begin with?

crates/chat-cli/src/cli/chat/cli/tangent.rs
```diff
@@ -46,12 +49,7 @@ impl TangentArgs {
 
     pub async fn execute(self, os: &Os, session: &mut ChatSession) -> Result<ChatState, ChatError> {
         // Check if tangent mode is enabled
-        if !os
-            .database
-            .settings
-            .get_bool(Setting::EnabledTangentMode)
-            .unwrap_or(false)
-        {
+        if !ExperimentManager::is_enabled(os, ExperimentName::TangentMode) {
```
dingfeli, line 52: Yea this is much cleaner than explicitly querying the database.

crates/chat-cli/src/cli/experiment/experiment_manager.rs
```diff
@@ -0,0 +1,126 @@
...
+    },
+    Experiment {
+        experiment_name: ExperimentName::ContextUsageIndicator,
+        description: "Shows context usage percentage in the prompt (e.g., [rust-agent] 6% >)",
+        setting_key: Setting::EnabledContextUsageIndicator,
+        enabled: true,
+    },
+];
+
+pub struct ExperimentManager;
```
dingfeli, line 78: Do we just want to convert these to pure util functions? There are no states associated with this struct.

crates/chat-cli/src/cli/chat/mod.rs
kensave: Yeah, good point that piece becomes irrelevant, let me improve.

crates/chat-cli/src/cli/experiment/experiment_manager.rs
kensave: I still want to keep it as a manager in case we need to go state-full then its easier to migrate.

crates/chat-cli/src/cli/chat/mod.rs
kensave: I think because this is dynamic, we should keep it for prevention.

crates/chat-cli/src/cli/experiment/experiment_manager.rs
```diff
@@ -0,0 +1,126 @@
...
+        experiment_name: ExperimentName::Checkpoint,
+        description: "Enables workspace checkpoints to snapshot, list, expand, diff, and restore files (/checkpoint)\nNote: Cannot be used in tangent mode (to avoid mixing up conversation history)",
+        setting_key: Setting::EnabledCheckpoint,
+        enabled: true,
+    },
+    Experiment {
+        experiment_name: ExperimentName::ContextUsageIndicator,
+        description: "Shows context usage percentage in the prompt (e.g., [rust-agent] 6% >)",
+        setting_key: Setting::EnabledContextUsageIndicator,
+        enabled: true,
```
kkashilk, line 74: nit: call this `active` or `activated`

crates/chat-cli/src/cli/experiment/experiment_manager.rs
```diff
@@ -0,0 +1,126 @@
...
+    },
+];
+
+pub struct ExperimentManager;
+
+impl ExperimentManager {
```
kkashilk, line 80: Can the ExperimentManager now be used to provide auto-complete on commands like `/knowledge`, `/checkpoint` if an experiment is enabled?

kensave: Added. Though we might need a bigger refactor to maybe make commands more dynamic rather than just session only. Cntrl + S works perfect though.

crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -97,21 +101,72 @@ pub const COMMANDS: &[&str] = &[
...
+    }
+
+    // Add tangent commands if TangentMode experiment is enabled
+    if ExperimentManager::is_enabled(os, ExperimentName::TangentMode) {
+        commands.extend_from_slice(&["/tangent", "/tangent tail"]);
+    }
+
+    commands.sort();
+    commands
+}
```
kkashilk, line 160: Should this be moved to ExperimentManager, so that someone adding an experiment doesn't have to make updates across multiple files?

kkashilk (approved):

crates/chat-cli/src/cli/chat/prompt.rs
kensave: I was thinking about that, it seems strange the experiment manager telling the commands in experiment, but makes sense.

PR #3060, opened 2025-09-30
feat: Support Kiro CLI social login

evanliu048:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/crates/chat-cli/src/auth/mod.rs
+++ b/crates/chat-cli/src/auth/mod.rs
@@ -2,6 +2,7 @@ pub mod builder_id;
 mod consts;
 pub mod pkce;
 mod scope;
+pub mod social;
 
 use aws_sdk_ssooidc::error::SdkError;
 use aws_sdk_ssooidc::operation::create_token::CreateTokenError;
@@ -48,6 +49,10 @@ pub enum AuthError {
     OAuthCustomError(String),
     #[error(transparent)]
     DatabaseError(#[from] crate::database::DatabaseError),
+    #[error(transparent)]
+    Reqwest(#[from] reqwest::Error),
+    #[error("HTTP error: {0}")]
+    HttpStatus(reqwest::StatusCode),
 }
 
 impl From<aws_sdk_ssooidc::Error> for AuthError {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1319,6 +1319,7 @@ dependencies = [
  "typed-path",
  "unicode-width 0.2.1",
  "url",
+ "urlencoding",
  "uuid",
  "walkdir",
  "webpki-roots 0.26.8",
--- a/crates/chat-cli/Cargo.toml
+++ b/crates/chat-cli/Cargo.toml
@@ -118,6 +118,7 @@ winnow.workspace = true
 schemars.workspace = true
 jsonschema.workspace = true
 rmcp.workspace = true
+urlencoding = "2.1.3"
 
 [target.'cfg(unix)'.dependencies]
 nix.workspace = true
--- a/crates/chat-cli/src/auth/social.rs
+++ b/crates/chat-cli/src/auth/social.rs
@@ -1,0 +1,420 @@
+use std::fmt;
+
+// use aws_smithy_runtime_api::client::identity::http::Token;
+// use aws_smithy_runtime_api::client::identity::{
+//     Identity,
+//     IdentityFuture,
+//     ResolveIdentity,
+// };
+// use aws_smithy_types::config_bag::ConfigBag;
+// use aws_smithy_types::runtime_components::RuntimeComponents;
+use eyre::{
+    Result,
+    bail,
+};
+use reqwest::Client;
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use time::OffsetDateTime;
+use tokio::net::TcpListener;
+use tokio::sync::mpsc;
+use tracing::{
+    debug,
+    error,
+    info,
+    trace,
+};
+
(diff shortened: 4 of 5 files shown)


crates/chat-cli/src/auth/index.html
```diff
@@ -156,23 +156,51 @@ <h4>Request denied</h4>
...
-
-                function showErrorMessage(errorText) {
-                    document.getElementById('approved-auth').classList.add('hidden')
-                    document.getElementById('denied-auth').classList.remove('hidden')
-                    document.getElementById('errorMessage').innerText = errorText
+
+                // Only show success message if explicitly marked as successful
+                // This prevents showing success when token exchange might still fail
+                if (success === 'true') {
+                    const productName = 'KIRO CLI';
```
kensave: `nit`: This logic could be extracted to their own method.

crates/chat-cli/src/auth/mod.rs
```diff
@@ -73,3 +98,33 @@ impl From<SdkError<StartDeviceAuthorizationError>> for AuthError {
...
+impl ResolveIdentity for UnifiedBearerResolver {
+    fn resolve_identity<'a>(
+        &'a self,
+        _runtime_components: &'a RuntimeComponents,
+        _config_bag: &'a ConfigBag,
+    ) -> IdentityFuture<'a> {
+        IdentityFuture::new_boxed(Box::pin(async {
+            let database = Database::new().await?;
+
+            if let Ok(Some(token)) = builder_id::BuilderIdToken::load(&database).await {
```
kensave, line 108: Why do we need to check both all times?

crates/chat-cli/src/auth/pkce.rs
```diff
@@ -282,42 +283,83 @@ impl PkceRegistration {
         Ok(())
     }
 
-    async fn recv_code(listener: TcpListener, expected_state: String) -> Result<String, AuthError> {
+    async fn recv_code(listener: tokio::net::TcpListener, expected_state: String) -> Result<String, AuthError> {
+        Self::recv_code_with_extra_accepts(listener, expected_state, 0).await
+    }
+
+    // NEW: extended version that accepts N extra connections to serve /index.html again
```
kensave: `nit`: `NEW` not needed as part of the comment,

crates/chat-cli/src/auth/pkce.rs
```diff
@@ -282,42 +283,83 @@ impl PkceRegistration {
...
+            _ = tokio::time::sleep(DEFAULT_AUTHORIZATION_TIMEOUT) => {
+                Err(AuthError::OAuthTimeout)
+            }
+        };
+
+        // Small grace period so the browser can load /index.html after the 302
+        tokio::time::sleep(Duration::from_millis(200)).await;
+
+        // Let the server task finish (best-effort)
+        let _ = tokio::time::timeout(Duration::from_secs(2), server_handle).await;
```
kensave: Whats the reasoning behind the 2 seconds? why is this best effort?

crates/chat-cli/src/cli/mod.rs
```diff
@@ -137,7 +137,10 @@ impl RootSubcommand {
 
     pub async fn execute(self, os: &mut Os) -> Result<ExitCode> {
         // Check for auth on subcommands that require it.
-        if self.requires_auth() && !crate::auth::is_logged_in(&mut os.database).await {
+        if self.requires_auth()
+            && !crate::auth::is_logged_in(&mut os.database).await
```
kensave: Seems this two methods can be combined into one is_loged_in whuch checks is_builder_id_loged_in, is_social_loged_in.

crates/chat-cli/src/auth/mod.rs
evanliu048: The `UnifiedBearerResolver` needs to support two login paths — **Builder ID** and **Social**.
A user may complete login through either method. Therefore, during the resolution phase, we need attempt both methods

crates/chat-cli/src/auth/pkce.rs
evanliu048: The 2-second window is a small buffer for the browser to finish loading /index.html after the 302 redirect.
By the time we receive the code and state, authentication is already complete — this delay just prevents occasional “connection refused” or partial page load issues.
It’s best-effort only: even if it fails, the CLI flow remains successful (token exchanged).
The 2-second value is empirical — long enough for typical browser follow-up requests without delaying the CLI

crates/chat-cli/src/telemetry/mod.rs
```diff
@@ -809,7 +818,7 @@ mod test {
             .await
             .unwrap();
 
-        thread.send_user_logged_in().ok();
+        let _ = thread.send_user_logged_in(&database, None).await;
```
dingfeli: Do we not want this on the critical path anymore? I am not saying that we should but I am just comparing this to what was changed.

crates/chat-cli/src/auth/index.html
evanliu048: Sure. Since we’re handling social login in the auth portal, I didn’t make any further changes to that part.

crates/chat-cli/src/auth/pkce.rs
evanliu048: handle the callback in portal.rs and I don't change this part anymore

crates/chat-cli/src/telemetry/mod.rs
evanliu048: thanks for flagging this issue. reverted change here

PR #3070, opened 2025-10-01
feat: add Stop hook

hugoncosta:
*Issue #, if available:* #3034

*Description of changes:*
See https://github.com/aws/amazon-q-developer-cli/pull/2875 for the initial implementation. Sometimes, the Agents run the tools several times in a particular turn (e.g. when writing code, they don't write the whole thing at once, especially when dealing with already existing files).

With this feature, you can set your hook to only run at the end of the turn, allowing you to only run e.g. formatting/tests when the whole file is written.

Tested locally, it's working as intended - only after doing all the fs_write did it do the formatting (I used the example currently in the hooks.md that runs cargo fmt after it's done).

This is a simplified version of what I had originally proposed in #3036, more in line with industry expectations as well.

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/knowledge.rs
+++ b/crates/chat-cli/src/cli/chat/cli/knowledge.rs
@@ -7,10 +7,7 @@ use crossterm::style::{
     Color,
 };
 use eyre::Result;
-use semantic_search_client::{
-    OperationStatus,
-    SystemStatus,
-};
+use semantic_search_client::SystemStatus;
 
 use crate::cli::chat::tools::sanitize_path_tool_arg;
 use crate::cli::chat::{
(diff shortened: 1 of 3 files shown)


docs/hooks.md
```diff
@@ -133,6 +133,26 @@ Runs after tool execution with access to tool results.
 - **0**: Hook succeeded.
 - **Other**: Show STDERR warning to user. Tool already ran.
 
+### EndOfTurn
```
erbenmo: Can we update the "Available hook triggers" under https://github.com/aws/amazon-q-developer-cli/blob/main/docs/agent-format.md#hooks-field as well?

erbenmo (approved):

docs/hooks.md
dk19y: +1

hugoncosta: done, also renamed to Stop per our offline conversation

dk19y (approved):

kkashilk (approved):

PR #3073, opened 2025-10-02
feat(agent): delegate tool

dingfeli:
*Issue #, if available:*
This is a continuation of https://github.com/aws/amazon-q-developer-cli/pull/2890

*Description of changes:*
In the interest of time, a lot of corners were cut: 
- [ ] Filter output from delegate task to capture only the important parts
- [ ] Async UI for real time update
- [ ] Offer a way for completed delegate task results to be attached to the next message more seamlessly
- [ ] In memory subagent without persisting via FS
- [ ] Better identifier for delegate tasks (currently it's just named after the agent employed to perform the task)

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/delegate.rs
+++ b/crates/chat-cli/src/cli/chat/tools/delegate.rs
@@ -269,7 +269,7 @@ pub struct AgentExecution {
     pub task: String,
     #[serde(default)]
     pub status: AgentStatus,
-    #[serde(default)]
+    #[serde(default, with = "chrono::serde::ts_seconds")]
     pub launched_at: chrono::DateTime<chrono::Utc>,
     #[serde(default, with = "chrono::serde::ts_seconds_option")]
     pub completed_at: Option<chrono::DateTime<chrono::Utc>>,


crates/chat-cli/src/cli/chat/cli/delegate.rs
```diff
@@ -0,0 +1,206 @@
...
+        session.stderr.flush()?;
+
+        Ok(ChatState::PromptUser {
+            skip_printing_tools: false,
+        })
+    }
+}
+
+fn is_enabled(os: &Os) -> bool {
+    os.database.settings.get_bool(Setting::EnabledDelegate).unwrap_or(false)
```
kensave: We should use experiment manager here.

crates/chat-cli/src/cli/chat/cli/delegate.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+async fn gather_executions(os: &Os) -> Result<Vec<(AgentExecution, PathBuf)>> {
+    let mut dir_walker = os.fs.read_dir(subagents_dir(os).await?).await?;
```
kensave: This seems a bit of just trusting whatever that folder could potentially have.

crates/chat-cli/src/cli/chat/cli/delegate.rs
```diff
@@ -0,0 +1,206 @@
...
+                    &session.conversation.agents,
+                    &prompt_str,
+                )
+                .await
+            },
+        };
+
+        match result {
+            Ok(output) => {
+                crossterm::queue!(session.stderr, crossterm::style::Print(format!("{}\n", output)))?;
```
kensave: `nit`: `use crossterm::` as it's more consistent with code base.

crates/chat-cli/src/cli/chat/prompt_parser.rs
```diff
@@ -12,12 +13,27 @@ pub struct PromptComponents {
...
 pub fn parse_prompt_components(prompt: &str) -> Option<PromptComponents> {
     // Expected format: "[agent] 6% !> " or "> " or "!> " or "[agent] ↯ > " or "6% ↯ > " etc.
+    let mut delegate_notifier = None::<String>;
     let mut profile = None;
     let mut warning = false;
     let mut tangent_mode = false;
     let mut usage_percentage = None;
     let mut remaining = prompt.trim();
 
+    // Check for delegate notifier first
```
kensave, line 23: nit: Id prefer to encapsulate this into a function so we give meaning to it, for someone reading this outside this PR, seems like magic code.

crates/chat-cli/src/cli/chat/tools/delegate/execution.rs
```diff
@@ -0,0 +1,170 @@
...
+        Command::new("kill")
+            .args(["-0", &pid.to_string()])
+            .output()
+            .map(|output| output.status.success())
+            .unwrap_or(false)
+    }
+
+    #[cfg(not(unix))]
+    {
+        // For non-Unix systems, assume process is alive (fallback)
```
kensave: Will this be problematic?

crates/chat-cli/src/cli/experiment/experiment_manager.rs
```diff
@@ -108,6 +110,13 @@ static AVAILABLE_EXPERIMENTS: &[Experiment] = &[
         enabled: true,
         commands: &[],
     },
+    Experiment {
+        experiment_name: ExperimentName::Delegate,
+        description: "Enables launching and managing asynchronous subagent processes",
+        setting_key: Setting::EnabledDelegate,
+        enabled: true,
+        commands: &[],
```
kensave, line 118: Can we add the commands here?? so they are available in auto-complete and cntrl+s

kensave: I guess this ones:

 `/delegate launch "Fix the bug in main.rs"` - Launch task with default agent
- `/delegate launch --agent coding "Fix the bug in main.rs"` - Launch with specific agent (shows approval dialog)
- `/delegate status` - Show summary of all tasks
- `/delegate status abc12345` - Show status of specific task
- `/delegate read abc12345` - Read output from completed task (triggers LLM analysis)
- `/delegate delete abc12345` - Delete task and its files
- `/delegate list` - List all tasks with timestamps and brief info

.gitignore
```diff
@@ -8,6 +8,10 @@ xcuserdata/
 .DS_Store
 node_modules/
 .eslintcache
+
+# Conversation/data files
+/delegate
```
brandonskiser: what is this?

crates/chat-cli/src/cli/chat/tools/delegate/execution.rs
```diff
@@ -0,0 +1,170 @@
...
+                let _ = os.fs.remove_file(path).await;
+            }
+
+            Ok(execution.format_status())
+        },
+        None => Ok(format!("No execution found for agent '{}'", agent)),
+    }
+}
+
+pub async fn status_all_agents(os: &Os) -> Result<String> {
```
brandonskiser: Why do we conflate actual errors with a status of not ready? Would prefer to instead have some structure around this as it's a more intuitive api

crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -497,6 +502,17 @@ pub fn rl(
         }
     }
 
+    // Add custom keybinding for Ctrl+D to open delegate command (configurable)
+    let delegate_key_char = match os.database.settings.get_string(Setting::DelegateModeKey) {
```
brandonskiser: I think we should make this conditional based on if delegate is enabled.

Also  - if key.len() != 1, then maybe don't create the keybind? Reasoning is that I might be used to having ctrl+d default behavior, and I don't know what other keybind to come up with just to avoid having delegate override my normal readline keybinds

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+use crate::cli::agent::Agents;
+use crate::cli::chat::tools::{
+    InvokeOutput,
+    OutputKind,
+};
+use crate::database::settings::Setting;
+use crate::os::Os;
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+/// Launch and manage asynchronous agent processes. This tool allows you to delegate tasks to agents
```
brandonskiser: Duplicated with the index.json schema?

If the thought is to have the tool index schema be auto-generated by `JsonSchema` then I don't think that's the best behavior - the backend API is a restricted subset of the supported schema (resulting in validation exceptions for valid json schemas), and the generated output of JsonSchema is not considered breaking changes across version updates. Might just be better to keep this a one-time manual process

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// Launch a new agent with a specified task
+    Launch,
+    /// Check the status of a specific agent or all agents if None is provided
+    Status,
+    /// List all available agents
+    List,
+}
+
+#[allow(unused_imports)]
+pub use types::{
```
brandonskiser: re-export types below struct defs?

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    AgentConfig,
+    AgentExecution,
+    AgentStatus,
+};
+
+impl Delegate {
+    pub async fn invoke(&self, os: &Os, _output: &mut impl Write, agents: &Agents) -> Result<InvokeOutput> {
+        if !is_enabled(os) {
```
brandonskiser: I guess we want to use ExperimentManager here

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+use crate::cli::DEFAULT_AGENT_NAME;
+use crate::cli::agent::Agents;
+use crate::cli::chat::tools::{
+    InvokeOutput,
+    OutputKind,
+};
+use crate::database::settings::Setting;
+use crate::os::Os;
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
```
brandonskiser: nit - put derives below doc comments

crates/chat-cli/src/cli/chat/tools/delegate/types.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct AgentExecution {
+    #[serde(default)]
+    pub agent: String,
+    #[serde(default)]
+    pub task: String,
+    #[serde(default)]
+    pub status: AgentStatus,
+    #[serde(default)]
+    pub launched_at: String,
```
brandonskiser: I would prefer using chrono for saving timestamps
e.g. for saving UNIX epoch timestamps with second granularity:
```
    #[serde(with = "chrono::serde::ts_seconds_option")]
    pub timestamp: Option<DateTime<Utc>>,
```

crates/chat-cli/src/cli/chat/tools/delegate/agent.rs
brandonskiser, line 1: nit - don't think it's really necessary to split these into separate agent/execution files, responsibilities for each of these files isn't clear imo

crates/chat-cli/src/cli/chat/tools/delegate/agent.rs
```diff
@@ -0,0 +1,62 @@
...
+    os.fs.write(&file_path, content).await?;
+    Ok(())
+}
+
+pub async fn agent_file_path(os: &Os, agent: &str) -> Result<PathBuf> {
+    let subagents_dir = subagents_dir(os).await?;
+    Ok(subagents_dir.join(format!("{}.json", agent)))
+}
+
+pub async fn subagents_dir(os: &Os) -> Result<PathBuf> {
```
brandonskiser: If we're sticking with subagents id'd by agent name, then we probably want to use a local directory instead. Otherwise, it's actually impossible to have the same agent work on separate projects

crates/chat-cli/src/cli/chat/cli/delegate.rs
dingfeli: yea this is meant to be temporary solution for the interim. In the long term we plan to use a full fledged purpose built in memory solution for this.

crates/chat-cli/src/cli/chat/tools/delegate/execution.rs
dingfeli: Same as https://github.com/aws/amazon-q-developer-cli/pull/3073#discussion_r2399599966

crates/chat-cli/src/cli/experiment/experiment_manager.rs
dingfeli: ah yea good call. I missed this

crates/chat-cli/src/cli/chat/tools/delegate/mod.rs
dingfeli: ah yea we don't have to have it here. I was using jsonschema to generate part of the schema and pasting it in the tool index.

.gitignore
dingfeli: Oh I think these are left over from the original implementation. Let me clean this up.

crates/chat-cli/src/cli/experiment/experiment_manager.rs
mschrage: This was an intentional omission. Adding new slash commands creates a user-facing contract, which I'd like to avoid while the design space is still being explored.

When subagents are invoked/queried using natural language, we preserve more flexibility to modify the implementation and concept structure going forward.

kensave:
You could rebase, clippy errors are now gone.

Issue #3086, opened 2025-10-03
AmazonQ CLI crashes when entering a long Korean prompt.


yakpoong:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 26.0.0 (25A354)

### Expected behaviour

When entering a prompt in Korean, the CLI should not crash and must function normally.

### Actual behaviour

When entering a long Korean prompt, the q chat CLI process panics (crashes).

### Steps to reproduce

When long CJK text is entered as the prompt text, the app crashes.

### Environment

```yaml
[q-details]
version = "1.17.1"
hash = "b0ef39f91625c36c2ed6f065628e7165fae9012c"
date = "2025-09-30T22:01:25.726592Z (2d ago)"

[system-info]
os = "macOS 26.0.0 (25A354)"
chip = "Apple M1 Pro"
total-cores = 8
memory = "16.00 GB"
```

yakpoong:
Closing this issue since the PR that resolves it has been merged.

PR #3087, opened 2025-10-03
fix: ensure `sanitize_user_prompt` handles UTF-8 boundaries correctly

h16rkim:
*Issue #, if available:*

https://github.com/aws/amazon-q-developer-cli/issues/3086

*Description of changes:*

### Screen Capture

crash

### Description

When executing a hook, the user’s prompt is sanitized through the sanitize_user_prompt(prompt) function. This function trims the prompt with a substring if its length is too long (over 4096 bytes).

However, in the case of multibyte characters such as those in CJK languages, **if a character is cut across a UTF-8 boundary, a panic occurs and the program terminates.**

**This PR removes such crashes and fixes the sanitize_user_prompt function** so that it works correctly.

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/hooks.rs
+++ b/crates/chat-cli/src/cli/chat/cli/hooks.rs
@@ -378,8 +378,16 @@ impl HookExecutor {
 
 /// Sanitizes a string value to be used as an environment variable
 fn sanitize_user_prompt(input: &str) -> String {
-    // Limit the size of input to first 4096 characters
-    let truncated = if input.len() > 4096 { &input[0..4096] } else { input };
+    // Limit the size of input to first 4096 bytes, respecting character boundaries
+    let truncated = if input.len() > 4096 {
+        let mut end = 4096;
+        while end > 0 && !input.is_char_boundary(end) {
+            end -= 1;
+        }
+        &input[0..end]
+    } else {
+        input
+    };
 
     // Remove any potentially problematic characters
     truncated.replace(|c: char| c.is_control() && c != '\n' && c != '\r' && c != '\t', "")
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/cli/hooks.rs
```diff
@@ -378,8 +378,15 @@ impl HookExecutor {
 
 /// Sanitizes a string value to be used as an environment variable
 fn sanitize_user_prompt(input: &str) -> String {
-    // Limit the size of input to first 4096 characters
-    let truncated = if input.len() > 4096 { &input[0..4096] } else { input };
+    // Limit the size of input to first 4096 bytes, respecting character boundaries
+    const MAX_LEN: usize = 4096;
+
+    let truncated = if input.len() > MAX_LEN {
```
brandonskiser: We have a util function that does this - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/cli/chat/util/mod.rs#L19

Can this be used instead?

brandonskiser (changes requested):

crates/chat-cli/src/cli/chat/cli/hooks.rs
h16rkim: Thank you. I changed it to use `truncate_safe` function

h16rkim:
@kensave @brandonskiser 

There’s another PR that fixes a similar UTF-8 boundary bug — can you review it as well?

https://github.com/aws/amazon-q-developer-cli/pull/3020

h16rkim:
Thank you for your reviews. Can we merge this PR ?

brandonskiser:
Thank you!

Issue #3092, opened 2025-10-03
bug: AWS CLI Parameter Validation Error for All Describe Commands


sam-fakhreddine:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macos 26.0

### Expected behaviour

# Bug Report: AWS CLI Parameter Validation Error for All Describe Commands

## Description
The Q Developer CLI generates AWS CLI `describe-*` commands with incorrect parameter casing in filter specifications. All describe commands fail with parameter validation errors because lowercase parameter names (`name`, `values`) are used instead of the required Pascal-case format (`Name`, `Values`).

<img width="1177" height="707" alt="Image" src="https://github.com/user-attachments/assets/1293f325-0959-4acf-b2a3-54459f340020" />

## Example Command Executed
```bash
aws ec2 describe-subnets \
  --filters '[{"name":"tag:Name","values":["*app-a*"]}]' \
  --profile test \
  --region us-east-1
```

## Error Output
```
Execution failed after 0.341s:

Parameter validation failed:
Unknown parameter in Filters[0]: "name", must be one of: Name, Values
Unknown parameter in Filters[0]: "values", must be one of: Name, Values
```

## Expected Behavior
Q Developer CLI should generate AWS CLI commands with correctly cased filter parameters:
```bash
aws ec2 describe-subnets \
  --filters '[{"Name":"tag:Name","Values":["*app-a*"]}]' \
  --profile test \
  --region us-east-1
```

## Actual Behavior
All describe commands are generated with lowercase `name` and `values` keys, causing systematic parameter validation failures across all AWS services.

## Impact
- **Severity**: Low (annoying but easily worked around)
- **Scope**: All AWS describe commands that use filters (EC2, RDS, ECS, Lambda, etc.)
- **User Impact**: Users must manually correct casing before commands will execute successfully

## Root Cause
The Q Developer CLI appears to have a systematic issue where it generates filter parameters using lowercase JSON keys instead of the Pascal-case format required by AWS CLI.

## Proposed Solution
1. Update the Q Developer CLI filter parameter generation logic to use Pascal-case keys (`Name`, `Values`)
2. Add validation to ensure generated AWS CLI commands conform to AWS CLI parameter specifications
[...]

PR #3093, opened 2025-10-03
refactor: extract truncate_output helper function

sam-fakhreddine:
This pull request introduces improvements to AWS CLI parameter handling in the `UseAws` tool, ensuring better compatibility with AWS CLI expectations. The main changes include recursive conversion of JSON keys to Pascal case for CLI parameters, refactoring output truncation logic, and adding targeted unit tests.

This closes: https://github.com/aws/amazon-q-developer-cli/issues/3092

### AWS CLI Parameter Handling Improvements

* Added a new method `convert_json_keys_to_pascal_case` that recursively converts all JSON object keys to Pascal case, ensuring AWS CLI receives parameters in the expected format. This is applied to non-string parameter values during CLI parameter construction. [[1]](diffhunk://#diff-3fd1fbe5110bd8e3be56caf26d224bae0050677d75ef4d8d296c297176a633efR171-R192) [[2]](diffhunk://#diff-3fd1fbe5110bd8e3be56caf26d224bae0050677d75ef4d8d296c297176a633efL168-R162)

### Output Formatting

* Refactored the output truncation logic into a reusable `truncate_output` function, simplifying how stdout and stderr are truncated and annotated.

### Testing Enhancements

* Added unit tests to verify that filter parameters and nested object keys are correctly converted to Pascal case, ensuring robust handling of AWS CLI requirements.

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/use_aws.rs
+++ b/crates/chat-cli/src/cli/chat/tools/use_aws.rs
@@ -81,25 +81,16 @@ impl UseAws {
         let stdout = output.stdout.to_str_lossy();
         let stderr = output.stderr.to_str_lossy();
 
-        let stdout = format!(
-            "{}{}",
-            &stdout[0..stdout.len().min(MAX_TOOL_RESPONSE_SIZE / 3)],
-            if stdout.len() > MAX_TOOL_RESPONSE_SIZE / 3 {
-                " ... truncated"
+        fn truncate_output(s: &str, max_size: usize) -> String {
+            if s.len() > max_size {
+                format!("{}... truncated", &s[0..max_size])
             } else {
-                ""
+                s.to_string()
             }
-        );
+        }
 
-        let stderr = format!(
-            "{}{}",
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/tools/use_aws.rs
```diff
@@ -174,6 +168,28 @@ impl UseAws {
         }
     }
 
+    /// Recursively converts all keys in a JSON value to Pascal case
+    /// This is needed because AWS CLI expects Pascal case keys (e.g., "Name", "Values")
+    /// in parameters like filters, while serde_json serializes with the original casing
+    fn convert_json_keys_to_pascal_case(value: &serde_json::Value) -> serde_json::Value {
```
brandonskiser, line 174: This doesn't seem true generally - consider `--environment '{"Variables":{"key":"value"}}'` for Lambda etc. so I don't think this is the correct solution

> serde_json serializes with the original casing

So if it's the model hallucinating incorrect parameters, should the tool spec instead be updated to prompt it to prefer pascal case for complex parameters?

PR #3099, opened 2025-10-06
benchmark: terminal bench local testing and instructions

xianwwu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/terminal-bench-test/main_local.py
+++ b/terminal-bench-test/main_local.py
@@ -1,0 +1,104 @@
+from terminal_bench.agents.base_agent import BaseAgent, AgentResult
+from terminal_bench.terminal.tmux_session import TmuxSession
+from pathlib import Path
+import time
+import subprocess
+import threading
+import os
+
+class AmazonQCLILocalAgent(BaseAgent):
+    @staticmethod
+    def name() -> str:
+        return "Amazon Q CLI Local Fixed"
+
+    def perform_task(self, instruction: str, session: TmuxSession, logging_dir: Path | None = None) -> AgentResult:
+        # Kill any existing HTTP server on port 8000
+        subprocess.run(["pkill", "-f", "python3 -m http.server 8000"], stderr=subprocess.DEVNULL)
+        time.sleep(1)
+        
+        # Start HTTP server
+        def start_server():
+            subprocess.run([
+                "python3", "-m", "http.server", "8000", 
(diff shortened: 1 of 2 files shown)


terminal-bench-test/main_local_README.md
```diff
@@ -0,0 +1,54 @@
+# Terminal Bench Test
+
+This repository contains the setup and execution instructions for running the terminal bench test with the Amazon Q CLI Local Agent.
+
+## Prerequisites
+
+- Ubuntu operating system
```
brandonskiser, line 7: Why Ubuntu? Could this instead be a docker container? Otherwise we won't be able to easily run this.

terminal-bench-test/main_local_README.md
```diff
@@ -0,0 +1,54 @@
...
+export AWS_SESSION_TOKEN=your_session_token_here
+```
+
+### 3. Create and Activate Python Virtual Environment
+Create and activate  Python 3.13 virtual environment. 
+
+Then install terminal bench. 
+
+```
+uv tool install terminal-bench
```
brandonskiser, line 33: Can this directory just be a uv managed project? So no need to have to e.g. run `pip install terminal-bench` and deal with virtual env's or whatever. Would be nice to just have some script and let `uv` do the rest

terminal-bench-test/main_local.py
```diff
@@ -0,0 +1,104 @@
...
+    def perform_task(self, instruction: str, session: TmuxSession, logging_dir: Path | None = None) -> AgentResult:
+        # Kill any existing HTTP server on port 8000
+        subprocess.run(["pkill", "-f", "python3 -m http.server 8000"], stderr=subprocess.DEVNULL)
+        time.sleep(1)
+        
+        # Start HTTP server
+        def start_server():
+            subprocess.run([
+                "python3", "-m", "http.server", "8000", 
+                "--directory", "/home/ec2-user/workspace-qcli/amazon-q-developer-cli/target/release"
```
brandonskiser, line 23: what are these directories? feels like this should be parameterized if it's a prerequisite to setup the codebase

I think ideally as part of the script, checking out the codebase to a temp directory can be done automatically, otherwise we can point to some other directory where it's already set up (potentially with docker mounts? So I can run terminal bench on my local development repo)

PR #3112, opened 2025-10-07
Dingfeli/chat cli UI revamp

dingfeli:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/crates/chat-cli-ui/src/conduit.rs
+++ b/crates/chat-cli-ui/src/conduit.rs
@@ -1,0 +1,179 @@
+use crossterm::{
+    execute,
+    style,
+};
+
+use crate::protocol::Event;
+
+#[derive(thiserror::Error, Debug)]
+pub enum ConduitError {
+    #[error(transparent)]
+    Send(#[from] Box<std::sync::mpsc::SendError<Event>>),
+    #[error(transparent)]
+    Utf8(#[from] std::string::FromUtf8Error),
+    #[error("No event set")]
+    NullState,
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+}
+
+/// The view would own this struct.
+/// [ViewEnd] serves two purposes
+/// - To deliver user inputs to the control layer from the view layer
+/// - To deliver state changes from the control layer to the view layer
+pub struct ViewEnd {
+    /// Used by the view to send input to the control
+    // TODO: later on we will need replace this byte array with an actual event type from ACP
+    pub sender: std::sync::mpsc::Sender<Vec<u8>>,
+    /// To receive messages from control about state changes
+    pub receiver: std::sync::mpsc::Receiver<Event>,
+}
+
+impl ViewEnd {
+    /// Method to facilitate in the interim
+    /// It takes possible messages from the old even loop and queues write to the output provided
+    /// This blocks the current thread and consumes the [ViewEnd]
+    pub fn into_legacy_mode(self, mut output: impl std::io::Write) -> Result<(), ConduitError> {
+        while let Ok(event) = self.receiver.recv() {
+            let content = match event {
+                Event::Custom(custom) => custom.value.to_string(),
+                Event::TextMessageContent(content) => content.delta,
+                Event::TextMessageChunk(chunk) => {
+                    if let Some(content) = chunk.delta {
+                        content
+                    } else {
+                        continue;
+                    }
+                },
+                _ => continue,
+            };
+
+            execute!(&mut output, style::Print(content))?;
+        }
+
+        Ok(())
+    }
+}
+
+/// This compliments the [ViewEnd]. It can be thought of as the "other end" of a pipe.
+/// The control would own this.
+pub struct ControlEnd {
+    pub current_event: Option<Event>,
+    /// Used by the control to send state changes to the view
+    pub sender: std::sync::mpsc::Sender<Event>,
+    /// To receive user input from the view
+    // TODO: later on we will need replace this byte array with an actual event type from ACP
+    pub receiver: std::sync::mpsc::Receiver<Vec<u8>>,
+}
+
+impl ControlEnd {
+    /// Primes the [ControlEnd] with the state passed in
+    /// This api is intended to serve as an interim solution to bridge the gap between the current
+    /// code base, which heavily relies on crossterm apis to print directly to the terminal and the
+    /// refactor where the message passing paradigm is the norm
+    pub fn prime(&mut self, event: Event) {
+        self.current_event.replace(event);
+    }
+
+    /// Sends an event to the view layer through the conduit
+    pub fn send(&self, event: Event) -> Result<(), ConduitError> {
+        Ok(self.sender.send(event).map_err(Box::new)?)
+    }
+}
+
+impl std::io::Write for ControlEnd {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        // We'll default to custom event
+        // This hardly matters because in legacy mode we are simply extracting the bytes and
+        // dumping it to output
+        if self.current_event.is_none() {
+            self.current_event.replace(Event::Custom(Default::default()));
+        }
+
+        let current_event = self
+            .current_event
+            .as_mut()
+            .ok_or(std::io::Error::other("No event set"))?;
+
+        current_event
+            .insert_content(buf)
+            .map_err(|_e| std::io::Error::other("Error inserting content"))?;
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let current_state = self.current_event.take().ok_or(std::io::Error::other("No state set"))?;
+
+        self.sender.send(current_state).map_err(std::io::Error::other)
+    }
+}
+
+/// Creates a bidirectional communication channel between view and control layers.
+///
+/// This function establishes a message-passing conduit that enables:
+/// - The view layer to send user input (as bytes) to the control layer
+/// - The control layer to send state changes to the view layer
+///
+/// # Returns
+/// A tuple containing:
+/// - `ViewEnd<S>`: The view-side endpoint for sending input and receiving state updates
+/// - `ControlEnd<S>`: The control-side endpoint for receiving input and sending state updates
+///
+/// # Type Parameters
+/// - `S`: The state type that implements `ViewState` trait
+pub fn get_conduit_pair() -> (ViewEnd, ControlEnd) {
+    let (state_tx, state_rx) = std::sync::mpsc::channel::<Event>();
+    let (byte_tx, byte_rx) = std::sync::mpsc::channel::<Vec<u8>>();
+
+    (
+        ViewEnd {
+            sender: byte_tx,
+            receiver: state_rx,
+        },
+        ControlEnd {
+            current_event: None,
+            sender: state_tx,
+            receiver: byte_rx,
+        },
+    )
+}
+
+pub trait InterimEvent {
+    type Error: std::error::Error;
+    fn insert_content(&mut self, content: &[u8]) -> Result<(), Self::Error>;
+}
+
+// It seems silly to implement a trait we have defined in the crate for a type we have also defined
+// in the same crate. But the plan is to move the Event type definition out of this crate (or use a
+// an external crate once AGUI has a rust crate)
+impl InterimEvent for Event {
+    type Error = ConduitError;
+
+    fn insert_content(&mut self, content: &[u8]) -> Result<(), ConduitError> {
+        debug_assert!(self.is_compatible_with_legacy_event_loop());
+
+        match self {
+            Self::Custom(_custom) => {
+                // custom events are defined in this UI crate
+                // TODO: use an enum as implement AsRef for it
+                // match custom.name.as_str() {
+                //     _ => {},
+                // }
+            },
+            Self::TextMessageContent(msg_content) => {
+                let str = String::from_utf8(content.to_vec())?;
+                msg_content.delta.push_str(&str);
+            },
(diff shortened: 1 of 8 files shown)


crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+    pub fn into_legacy_mode(
+        self,
+        theme_source: impl ThemeSource,
+        mut stderr: std::io::Stderr,
+        mut stdout: std::io::Stdout,
+    ) -> Result<(), ConduitError> {
+        while let Ok(event) = self.receiver.recv() {
+            match event {
+                Event::LegacyPassThrough(content) => match content {
+                    LegacyPassThroughOutput::Stderr(content) => {
```
kensave, line 62: Are all events going to go through this event type?

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+                        is_trusted,
+                        mcp_server_name,
+                        ..
+                    } = tool_call_start;
+
+                    queue!(
+                        stdout,
+                        theme_source.emphasis_fg(),
+                        Print(format!(
+                            "🛠️  Using tool: {}{}",
```
kensave, line 100: whats the trade off here? This implies all tool uses have same UI rendering? I suppose this would be translated by a Renderer or sorts which will compose into specialized components

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+                        theme_source.warning_fg(),
+                        Print(name),
+                        theme_source.error_fg(),
+                        Print(" is rejected because it matches one or more rules on the denied list:"),
+                        Print(reason),
+                        Print("\n"),
+                        theme_source.reset(),
+                    )?;
+                },
+            }
```
kensave, line 173: Do we have a Thinking event or equivalent?

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn flush(&mut self) -> std::io::Result<()> {
+        if let Some(current_state) = self.current_event.take() {
+            self.sender.send(current_state).map_err(std::io::Error::other)
+        } else {
+            Ok(())
+        }
+    }
+}
+
+/// Creates a set of legacy conduits forcommunication between view and control layers.
```
kensave: nit: for communication

crates/chat-cli-ui/src/protocol.rs
```diff
@@ -0,0 +1,499 @@
...
+
+/// Signals the end of a text message
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TextMessageEnd {
+    pub message_id: String,
+}
+
+/// A self-contained text message event that combines start, content, and end
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
kensave, line 115: Are we going to support tool results streaming like : shell commands.

crates/chat-cli-ui/src/protocol.rs
```diff
@@ -0,0 +1,499 @@
...
+                | Event::TextMessageEnd(_)
+                | Event::TextMessageChunk(_)
+        )
+    }
+
+    /// Check if this is a tool call event
+    pub fn is_tool_call_event(&self) -> bool {
+        matches!(
+            self,
+            Event::ToolCallStart(_) | Event::ToolCallArgs(_) | Event::ToolCallEnd(_) | Event::ToolCallResult(_)
```
kensave, line 472: To reply to my on question, at least the event is not classified under ToolCallDelta or similar.

crates/chat-cli-ui/src/ui/components/mod.rs
```diff
@@ -0,0 +1,45 @@
+#![allow(dead_code)]
```
kensave, line 1: Do we want to allow dead code?

crates/chat-cli-ui/src/ui/components/app.rs
```diff
@@ -0,0 +1,46 @@
...
+        let (action_tx, mut _action_rx) = unbounded_channel::<Action>();
+
+        let mut tui = Tui::new(4.0, 60.0)?;
+        // TODO: make a defer routine that restores the terminal on exit
+        tui.enter()?;
+
+        let mut event_receiver = tui.event_rx.take().expect("Missing event receiver");
+
+        // Render Task
+        tokio::spawn(async move {
```
kensave, line 24: I'd be nice to have some screenshots of any visually noticeable change(if any)

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2847,28 +2846,48 @@ impl ChatSession {
...
                 drop(self.spinner.take());
-                queue!(
-                    self.stderr,
-                    terminal::Clear(terminal::ClearType::CurrentLine),
-                    cursor::MoveToColumn(0),
-                    cursor::Show
-                )?;
             }
 
+            info!("## control end: buf: {:?}", buf);
```
brandonskiser, line 2851: make this `debug!` or add more context to this info msg?

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2847,28 +2846,48 @@ impl ChatSession {
...
-                        Some(err) => return Err(ChatError::Custom(err.to_string().into())),
-                        None => break, // Data was incomplete
-                    },
+                if self.stdout.should_send_structured_event {
+                    match interpret_markdown(input, &mut temp_buf, &mut state) {
+                        Ok(parsed) => {
+                            offset += parsed.offset_from(&input);
+                            temp_buf.flush()?;
+
+                            let text_msg_content = TextMessageContent {
```
brandonskiser, line 2864: `temp_buf` would contain ANSI-escaped markdown content, right? Wouldn't we want `TextMessageContent` to just be the raw model response? It should be up to the UI layer consuming these events to interpret the markdown and do the ANSI/markdown parsing.

crates/chat-cli/src/util/ui.rs
```diff
@@ -151,3 +157,18 @@ fn print_with_bold(output: &mut impl Write, segments: &[(String, bool)]) -> Resu
     }
     Ok(())
 }
+
+#[derive(Default, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum UiMode {
```
brandonskiser, line 163: Doc comments on what these different UI modes are for would be helpful

crates/chat-cli-ui/src/protocol.rs
```diff
@@ -0,0 +1,499 @@
...
+}
+
+// ============================================================================
+// Lifecycle Events
+// ============================================================================
+
+/// Signals the start of an agent run
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RunStarted {
```
brandonskiser: What is an "agent run" ? Is this referring to a user turn?

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+/// The control would own this.
+#[derive(Debug)]
+pub struct ControlEnd<T> {
+    pub current_event: Option<Event>,
+    /// Used by the control to send state changes to the view
+    pub sender: std::sync::mpsc::Sender<Event>,
+    /// Flag indicating whether structured events should be sent through the conduit.
+    /// When true, the control end will send structured event data in addition to
+    /// raw pass-through content, enabling richer communication between layers.
+    pub should_send_structured_event: bool,
```
brandonskiser, line 199: This doesn't actually impact any behavior within the control/view end right? We can handle either structured events or byte content regardless of the value of this flag.

This is just a sort of helper state flag for the chat code base which write to the control end so that we don't need to pass some extra flag everywhere about this.

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+        }
+    }
+}
+
+impl<T> ControlEnd<T> {
+    /// Primes the [ControlEnd] with the state passed in
+    /// This api is intended to serve as an interim solution to bridge the gap between the current
+    /// code base, which heavily relies on crossterm apis to print directly to the terminal and the
+    /// refactor where the message passing paradigm is the norm
+    pub fn prime(&mut self, event: Event) {
```
brandonskiser, line 222: Not seeing any references to this function - what is this for exactly?

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+            current_event: self.current_event.clone(),
+            should_send_structured_event: self.should_send_structured_event,
+            sender: self.sender.clone(),
+            pass_through_destination: PhantomData,
+        }
+    }
+}
+
+impl std::io::Write for ControlEnd<DestinationStderr> {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
```
brandonskiser, line 255: So by default `stderr` is not line-buffered like `stdout`, but it seems like now we would need to call `flush` from the chat-cli side in order to actually print content through `stderr`. Is my understanding correct? This seems like stderr should just call flush at the end always

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+        if let Some(current_state) = self.current_event.take() {
+            self.sender.send(current_state).map_err(std::io::Error::other)
+        } else {
+            Ok(())
+        }
+    }
+}
+
+impl std::io::Write for ControlEnd<DestinationStdout> {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
```
brandonskiser, line 288: `stdout` is line-buffered, so I would expect us to call `flush` after seeing a newline, right?

crates/chat-cli-ui/src/conduit.rs
```diff
@@ -0,0 +1,394 @@
...
+        ControlEnd {
+            current_event: None,
+            should_send_structured_event,
+            sender: state_tx,
+            pass_through_destination: PhantomData,
+        },
+    )
+}
+
+pub trait InterimEvent {
```
brandonskiser, line 408: Wondering what this trait is required for

crates/chat-cli-ui/src/ui/components/mod.rs
dingfeli: Yea so these are the scaffolding. My follow up will be using these. 
Also this is in a separate crate so it's well contained. It's not ideal but they won't exist for long.

crates/chat-cli-ui/src/ui/components/app.rs
dingfeli: This is not being used just yet. But yea when we do have something to show I will include it in the next PR then.

crates/chat-cli-ui/src/protocol.rs
dingfeli: My rationale here is that if we do IPC these are already serializable.

crates/chat-cli-ui/src/conduit.rs
dingfeli: Oh yea we do. Let me see if I can incorporate that.

crates/chat-cli-ui/src/conduit.rs
dingfeli: In passthrough mode (stage B) yea.

crates/chat-cli-ui/src/conduit.rs
dingfeli: For legacy mode with structured events (stage C), it is using the current UI but listening for some structured events. The purpose of this is to incrementally work towards a UI that is identical in visual but is capable to sporting another event loop. The reason why the current event loop is instrumented (in some places) to emit these events is to give this mode of the UI a test run.

crates/chat-cli/src/cli/chat/mod.rs
dingfeli: Ahh yea this I forgot to remove.

crates/chat-cli/src/cli/chat/mod.rs
dingfeli: Yea that's the idea for when we do the UI for real. This is a quick and dirty way to do this without changing the current code base too much. When we do it for real (i.e. when we move on from legacy mode) we would need to include the formatting logic in the `ViewEnd`.

crates/chat-cli-ui/src/conduit.rs
dingfeli: yea exactly. It also spares us from having to read the database excessively.

crates/chat-cli-ui/src/conduit.rs
dingfeli: Ahh yea originally I was going to use this prior to writing for structured events. But I had since then changed by mind and forgot to remove this. I shall do that.

crates/chat-cli-ui/src/conduit.rs
dingfeli: Good call out. Will adjust accordingly.

crates/chat-cli-ui/src/conduit.rs
dingfeli: Good call out. Will adjust accordingly.

crates/chat-cli-ui/src/conduit.rs
dingfeli: I have decided to do that in another PR.

PR #3114, opened 2025-10-07
feat: Adds continuation id logic

kensave:
*Issue #, if available:*

This enables the **continuation id** logic on the client side. The goal for this is to separate user turns with separate ids, so each subsequent LLM call executed preserves the same identifier, while user initiated interactions will trigger the id generation for those.

User: Hello
Model: Hi
------> Same interaction id.

User: Do a, b, c
Model: I'll do a
Model: I'll do b
Model: I'll do c
------> Same interaction id

Also when using y, n or t to interact with the model responses, won't cause it to generate a new identifier, while using the rejection with a reason, will trigger cancellation again.

*Description of changes:*


crates/chat-cli/src/api_client/mod.rs
```diff
@@ -400,6 +401,8 @@ impl ApiClient {
                         .map(|v| v.into_iter().map(|i| i.try_into()).collect::<Result<Vec<_>, _>>())
                         .transpose()?,
                 )
+                .set_agent_continuation_id(agent_continuation_id)
+                .agent_task_type(amzn_codewhisperer_streaming_client::types::AgentTaskType::Vibe)
```
kensave, line 405: Keeping Vibe for now.

crates/chat-cli/src/api_client/model.rs
```diff
@@ -870,7 +905,8 @@ impl From<UserInputMessage> for amzn_codewhisperer_streaming_client::types::User
             .set_user_input_message_context(value.user_input_message_context.map(Into::into))
             .set_user_intent(value.user_intent.map(Into::into))
             .set_model_id(value.model_id)
-            .origin(amzn_codewhisperer_streaming_client::types::Origin::Cli)
+            //TODO: Setup new origin.
+            .origin(amzn_codewhisperer_streaming_client::types::Origin::AiEditor)
```
kensave, line 909: AiEditor works, though we need to change to new origin as soon as we have it available.

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -378,7 +383,32 @@ impl ConversationState {
...
+        self.set_next_user_message_with_continuation_policy(input, false).await;
+    }
+
+    /// Set the next user message with control over continuation ID generation
+    ///
+    /// # Arguments
+    /// * `input` - The user input message
+    /// * `preserve_continuation_id` - If true, keeps existing continuation ID; if false, generates
+    ///   new one
+    pub async fn set_next_user_message_with_continuation_policy(
```
kensave: Here we just hung on the existing set_next_user_message for convinience

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2124,14 +2123,24 @@ impl ChatSession {
...
                 // we decide on a better flow.
-                let user_input = if ["n", "N"].contains(&user_input.trim()) {
+                let user_input = if is_reject_response(user_input.trim()) {
                     "I deny this tool request. Ask a follow up question clarifying the expected action".to_string()
                 } else {
                     user_input
                 };
-                self.conversation.abandon_tool_use(&self.tool_uses, user_input);
+
+                // For tool approval responses (y/n/t), preserve continuation ID for billing consistency
```
kensave: Basically any per-selected option keeps the continuation id.

crates/chat-cli/src/cli/chat/parser.rs
```diff
@@ -470,14 +470,24 @@ impl ResponseParser {
...
+                    if let Some(ChatResponseStream::ToolUseEvent { input, stop, .. }) = self.next().await? {
+                        if let Some(i) = input {
+                            tool_string.push_str(&i);
+                        }
+                        if let Some(true) = stop {
+                            break;
+                        }
+                    }
+                },
+                Some(ChatResponseStream::MetadataEvent { .. } | ChatResponseStream::MeteringEvent { .. }) => {
```
kensave, line 485: Metering event was breaking the event loop

crates/chat-cli/src/api_client/model.rs
brandonskiser: My understanding was that this is how routing is controlled within RTS/Maestro so we know we're using the CLI system prompt - does this impact that flow?

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -149,6 +150,9 @@ pub struct ConversationState {
     /// Tangent mode checkpoint - stores main conversation when in tangent mode
     #[serde(default, skip_serializing_if = "Option::is_none")]
     tangent_state: Option<ConversationCheckpoint>,
+    /// Current continuation ID for billing tracking - generated per conversation turn
```
brandonskiser: "billing tracking" - having any influence on customer billing be influenced by the client side (especially one that is open source) is not a good idea. Is this how it will really work? Or is this doc comment referencing something else?

crates/chat-cli/src/cli/chat/conversation.rs
```diff
@@ -149,6 +150,9 @@ pub struct ConversationState {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    current_continuation_id: Option<String>,
```
brandonskiser: Is it possible to combine this with `ChatSession::user_turn_request_metadata` so that the user turn tracking logic and state regarding the current user turn is the same? It sounds like a user turn has the same definition with regards to science/telemetry as it does with continuation id's

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -2124,14 +2123,24 @@ impl ChatSession {
...
+                );
             } else {
-                self.conversation.set_next_user_message(user_input).await;
+                // For regular user messages, always generate new continuation ID
+                self.conversation
+                    .set_next_user_message_with_continuation_policy(user_input, false)
+                    .await;
             }
 
             self.reset_user_turn();
```
brandonskiser: this line actually looks like a bug with the current user turn tracking in our telemetry - anything that requires approvals with "y/Y" will always reset the user turn

crates/chat-cli/src/cli/chat/mod.rs
brandonskiser: Why do we have it so that "n/N" continues the user turn? In our response, we tell the model to ask why we are rejecting it, in which case it would very likely *not* generate a new tool use and instead end the user turn anyways. The only difference here is that one has one more message than the other

In my mind, "n/N" should end the current user turn and thus create a new continuation id, right? Otherwise, maybe we should update the "n/N" flow so that we force the user to provide some reason why they are rejecting it, so we could save on a request-response cycle.

crates/chat-cli/src/api_client/model.rs
kensave: It's functional and enables local testing, but we need to definitely get the back-end work done. It's important because it really changes the behavior of the back-end including the metering events.

crates/chat-cli/src/cli/chat/mod.rs
kensave: That's fair, I think we can consider N and end of turn, but anyway as you said it will stop the flow. It's technically part of previous set of messages. For IDE at least all tool uses are generated upfront.

If n is hit, then next user message will anyways start the new turn. Breaking the last message from previous flow of execution doesn't seem logically correct. Here we keep the tail of the conversation which is the last interaction in the agentic flow with the reason(Client generated) of why to stop, user has NO control over this message hence keeping it.

crates/chat-cli/src/cli/chat/conversation.rs
kensave: Actually I need to strip that word out of the the codebase. It doesn't really influence metering.

crates/chat-cli/src/cli/chat/conversation.rs
kensave: Makes sense.

crates/chat-cli/src/cli/chat/mod.rs
kensave: Interesting, yeah I noticed some anomalies on this user turn reset. At least for the continuation id is nit reset right now.

crates/chat-cli/src/cli/chat/conversation.rs
kensave: It can't be completely centralized as we still need to keep track of latest in ConversationState, but at least it's aligned with all reset_user_turn

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -560,6 +560,60 @@ impl From<parser::RecvError> for ChatError {
...
+            continuation_id: uuid::Uuid::new_v4().to_string(),
+            requests: vec![],
+        }
+    }
+
+    pub fn continuation_id(&self) -> &str {
+        &self.continuation_id
+    }
+
+    pub fn push_request(&mut self, request: RequestMetadata) {
```
brandonskiser: `push_request` and `add_request` are identical

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3663,6 +3741,57 @@ fn does_input_reference_file(input: &str) -> Option<ChatState> {
...
+}
+
+/// Check if input is any simple tool interaction response (y/Y/n/N/t/T)
+/// These responses should preserve continuation ID
+fn is_simple_tool_interaction(input: &str) -> bool {
+    is_trust_response(input) || is_accept_response(input) || is_reject_response(input)
+}
+
+/// Check if input is any approval response (y/Y/t/T)
+fn is_approval_response(input: &str) -> bool {
```
brandonskiser, line 3712: imo it's more readable to inline this function instead, it's only referenced in one other location anyways

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3663,6 +3741,57 @@ fn does_input_reference_file(input: &str) -> Option<ChatState> {
...
+}
+
+/// Check if input is a "reject" response (n/N)
+fn is_reject_response(input: &str) -> bool {
+    ["n", "N"].contains(&input.trim())
+}
+
+/// Check if input is any simple tool interaction response (y/Y/n/N/t/T)
+/// These responses should preserve continuation ID
+fn is_simple_tool_interaction(input: &str) -> bool {
```
brandonskiser: `is_tool_permission_interaction` is a bit clearer imo

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3387,7 +3459,12 @@ impl ChatSession {
     /// that includes tool use rejections.
     fn reset_user_turn(&mut self) {
         info!(?self.user_turn_request_metadata, "Resetting the current user turn");
-        self.user_turn_request_metadata.clear();
+        self.user_turn_request_metadata = UserTurnMetadata::new();
+        // Keeps track of current continuation id.
+        // Maintained in conversation state for convenience
+        // as it needs to be propagated all way till the API.
+        self.conversation
```
brandonskiser: Is there any reason why `user_turn_request_metadata` couldn't be stored on `ConversationState` instead?

crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -560,6 +564,57 @@ impl From<parser::RecvError> for ChatError {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UserTurnMetadata {
```
brandonskiser: nit - I'd probably put this in the conversationstate file instead, chat/mod.rs is large enough as is and this struct is only stored on conversationstate

Issue #3136, opened 2025-10-09
Amazon Q CLI Crashes when truncate checkpoint message


h16rkim:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 26.0.0 (25A354)

### Expected behaviour

Amazon Q Developer CLI should not crash when truncate checkpoint messages written in multibyte characters

### Actual behaviour

Amazon Q Developer CLI crashes when truncate checkpoint messages written in multibyte characters

### Steps to reproduce

1. enable checkpoint features
2. write prompt in multibyte characters (ex. CJK Languages)
3. Amazon Q Developer CLI crashes when truncate checkpoint messages

```
The application panicked (crashed).
Message:  byte index 60 is not a char boundary; it is inside '는' (bytes 59..62) of `CSR 페이지를 렌더링하고 있는데, html, 이미지는 화면에 잘 나타나는데 css, js 파일이 정상적으
로 로드되지 않은 것 같아. 어떻게 코드를 고치면 정상적으로 CSR 페이지가 로드되게 할 수 있
을까 ?`
Location: crates/chat-cli/src/cli/chat/checkpoint.rs:353
```

### Environment

```yaml
[q-details]
version = "1.17.1"
hash = "b0ef39f91625c36c2ed6f065628e7165fae9012c"
date = "2025-09-30T22:01:25.726592Z (8d ago)"

[system-info]
os = "macOS 26.0.0 (25A354)"
chip = "Apple M1 Pro"
total-cores = 8
memory = "16.00 GB"
```

kensave:
Fix merged.

PR #3137, opened 2025-10-09
fix:bug(checkpoint): fix crashes when truncating checkpoint message 

h16rkim:
*Issue #, if available: https://github.com/aws/amazon-q-developer-cli/issues/3136*

*Description of changes:*

When using checkpoint features, the user’s prompt is truncated through the `truncate_message` function. This function trims the prompt with a substring if its length is too long (length: 60).

However, in the case of multibyte characters such as those in CJK languages, **if a character is cut across a UTF-8 boundary, a panic occurs and the program terminates.**

**This PR removes such crashes and fixes the truncate_message** so that it works correctly.

*Screen Capture*

checkpoint-crash

### DIFF
--- a/crates/chat-cli/src/cli/chat/checkpoint.rs
+++ b/crates/chat-cli/src/cli/chat/checkpoint.rs
@@ -28,6 +28,7 @@ use serde::{
 use tracing::debug;
 
 use crate::cli::ConversationState;
+use super::util::truncate_safe;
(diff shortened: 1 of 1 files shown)


h16rkim:
Similar Pull Request(UTF-8 boundary issue): https://github.com/aws/amazon-q-developer-cli/pull/3087

Issue #3106, opened 2025-10-06
feat: Display tokens per second


jci-aws:
Show the tokens-per-second (and possibly more detailed information like eval rate, durations, etc) for each message during a conversation. Could show this when Q is run with `--verbose`

Inspiration from https://www.reddit.com/r/LocalLLaMA/comments/1egxdpt/post_your_tokens_per_second_for_llama3170b/

PR #3149, opened 2025-10-10
feat: display token performance metrics on verbose

marcvernet31:
*Issue #, if available:* https://github.com/aws/amazon-q-developer-cli/issues/3106

*Description of changes:*
Displays detailed token performance metrics (tokens/sec, TTFT, latency) when using verbose flags.
- performance.rs - New module calculating token metrics from request metadata
- mod.rs - Integration logic and styled display output to stderr
- parser.rs - Enhanced metadata support for performance tracking

On this implementation the metrics will be displayed for all levels of verbose, but it can be discussed if it should be displayed only on a specific level or if we can have a specific flag for them (--metrics for example). 

Example output:
```
> hello

> Hello! I'm Amazon Q, an AI assistant built by AWS. I'm here to help you with AWS services, development tasks, 
infrastructure management, and more. 

What can I help you with today?

┌─ Performance Metrics ────────────────────────────────────────┐
│ Performance Metrics:
│   Tokens/sec: 0.0
│   TTFT: 2766ms
│   Total duration: 2.8s
│   Generation time: 0ms
│   Avg inter-token latency: 0.0ms
│   Total tokens: 10 (40 prompt + 0 completion)
│   Prompt processing: 2766ms
```

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -10,6 +10,7 @@ use std::path::MAIN_SEPARATOR;
 pub mod checkpoint;
 mod line_tracker;
 mod parser;
+mod performance;
 mod prompt;
 mod prompt_parser;
 pub mod server_messenger;
@@ -55,8 +56,7 @@ use cli::model::{
 pub use conversation::ConversationState;
 use conversation::TokenWarningLevel;
 use crossterm::style::{
-    Attribute,
-    Stylize,
+    Attribute, Color, Stylize
 };
 use crossterm::{
     cursor,
@@ -2966,6 +2966,18 @@ impl ChatSession {
             self.send_chat_telemetry(os, TelemetryResult::Succeeded, None, None, None, true)
                 .await;
 
+            // Display performance metrics if verbose mode is enabled
+            let should_show_metrics = self.should_show_performance_metrics();
+            if should_show_metrics {
+                // Display metrics for the most recent request
+                if let Some(latest_metadata) = self.user_turn_request_metadata.last().cloned() {
+                    if let Err(e) = self.display_performance_metrics(&latest_metadata).await {
+                        // Log error but don't fail the conversation
+                        error!(?e, "Failed to display performance metrics");
+                    }
+                }
+            }
+
             // Run Stop hooks when the assistant finishes responding
             if let Some(cm) = self.conversation.context_manager.as_mut() {
                 let _ = cm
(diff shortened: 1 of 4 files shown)


crates/chat-cli/src/cli/chat/mod.rs
```diff
@@ -3336,6 +3348,81 @@ impl ChatSession {
         self.conversation.agents.trust_all_tools
     }
 
+    /// Check if performance metrics should be displayed based on verbose flag
+    fn should_show_performance_metrics(&self) -> bool {
+        self.get_verbose_level() > 0
+    }
+
+    /// Get the current verbose level from CLI args or environment variables
+    fn get_verbose_level(&self) -> u8 {
```
brandonskiser, line 3357: Not a fan of doing this type of argument parsing - I would rather we support a new slash command like `/debug` that can be extended more easily, for example here we could have `/debug metrics` that could print out the last request metrics.

Is it necessary to do this printing on every request though? If so, we could have a flag that gets toggled on/off through `/debug metrics toggle` maybe

crates/chat-cli/src/cli/chat/parser.rs
```diff
@@ -641,6 +669,23 @@ impl ResponseParser {
         }
     }
 
+    /// Track tokens for the given content and record timing information
+    fn track_tokens(&mut self, content: &str) {
```
brandonskiser, line 673: This has to be pretty inaccurate - the token estimate alone is not accurate, and if we're doing this on tiny content chunks then the actual error amount is going to be huge.

Should stick to counting chars instead, and then estimate tokens from that

crates/chat-cli/src/cli/chat/parser.rs
```diff
@@ -322,6 +322,14 @@ struct ResponseParser {
...
     time_to_first_chunk: Option<Duration>,
     time_between_chunks: Vec<Duration>,
+
+    // Token tracking fields
+    /// Total tokens generated in the response so far
+    total_tokens: usize,
+    /// Time when first token was received
+    first_token_time: Option<Instant>,
+    /// Incremental token counts with timestamps (duration from request start in ms, token count)
+    token_timestamps: Vec<(u64, usize)>,
```
brandonskiser, line 332: How is this different from time_between_chunks?

crates/chat-cli/src/cli/chat/parser.rs
```diff
@@ -322,6 +322,14 @@ struct ResponseParser {
     received_response_size: usize,
     time_to_first_chunk: Option<Duration>,
     time_between_chunks: Vec<Duration>,
+
+    // Token tracking fields
+    /// Total tokens generated in the response so far
+    total_tokens: usize,
+    /// Time when first token was received
+    first_token_time: Option<Instant>,
```
brandonskiser, line 330: How is this different from time_to_first_chunk?

crates/chat-cli/src/cli/chat/parser.rs
```diff
@@ -685,6 +739,33 @@ pub enum ResponseEvent {
     },
 }
 
+/// Token count metrics for performance calculation
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TokenMetrics {
```
brandonskiser, line 744: Don't define TokenMetrics like this since it's all based on inaccurate estimates rather than concrete token usage.

Why can't this be derived from `RequestMetadata`? This appears to be all duplicated except using a token estimate instead

brandonskiser:
Hey, thanks! Can you also run `cargo +nightly fmt`?

PR #3150, opened 2025-10-10
bskiser/add agent crate

brandonskiser:
*Description of changes:*
- Adding a new crate `agent` that implements the agent config. This crate is based on an actor model where an agent can be interacted with using an `AgentHandle`, from which messages (send prompt, approve tool use, etc.) are sent and events are received.
- A minimal `run` CLI is provided which illustrates the API

This contains substantial code duplication from `chat-cli` for the API client and database since this was developed in isolation from the rest of the code (see *fast follows* below).

A lot of this is still WIP, the goal here is to get this pushed and incrementally develop this crate while working on the `/delegate` tool and overall chat refactor.

**Todos (as of now):**
- Implement remote HTTP MCP support
- Finish built-in tools - `fs_read` and `fs_write`, and the experimental tools we have today.
- Test coverage and validation that the agent spec is fully implemented

There are a lot of details that will be uncovered as more work is done adding this crate to the main `chat-cli` code base.

**Fast Follows:**
- Remove the copied code for the database and API client. `Agent` would instead just have a `Box<dyn Model>` passed to it. All of the `rts` implementation for `Model` should be moved to `chat-cli` instead.

### DIFF
--- a/crates/agent/src/agent/tools/fs_write.rs
+++ b/crates/agent/src/agent/tools/fs_write.rs
@@ -389,10 +389,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_str_replace_single_occurrence() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("test.txt", "hello world"))
-            .await;
+        let test_base = TestBase::new().await.with_file(("test.txt", "hello world")).await;
 
         let tool = FsWrite::StrReplace(StrReplace {
             path: test_base.join("test.txt").to_string_lossy().to_string(),
@@ -409,10 +406,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_str_replace_multiple_occurrences() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("test.txt", "foo bar foo"))
-            .await;
+        let test_base = TestBase::new().await.with_file(("test.txt", "foo bar foo")).await;
 
         let tool = FsWrite::StrReplace(StrReplace {
             path: test_base.join("test.txt").to_string_lossy().to_string(),
@@ -429,10 +423,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_str_replace_no_match() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("test.txt", "hello world"))
-            .await;
+        let test_base = TestBase::new().await.with_file(("test.txt", "hello world")).await;
 
         let tool = FsWrite::StrReplace(StrReplace {
             path: test_base.join("test.txt").to_string_lossy().to_string(),
@@ -465,10 +456,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_insert_append() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("test.txt", "existing"))
-            .await;
+        let test_base = TestBase::new().await.with_file(("test.txt", "existing")).await;
 
         let tool = FsWrite::Insert(Insert {
             path: test_base.join("test.txt").to_string_lossy().to_string(),
--- a/crates/agent/src/agent/tools/image_read.rs
+++ b/crates/agent/src/agent/tools/image_read.rs
@@ -254,10 +254,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_read_valid_image() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("test.png", create_test_png()))
-            .await;
+        let test_base = TestBase::new().await.with_file(("test.png", create_test_png())).await;
 
         let tool = ImageRead {
             paths: vec![test_base.join("test.png").to_string_lossy().to_string()],
@@ -294,10 +291,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_validate_unsupported_format() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("test.txt", "not an image"))
-            .await;
+        let test_base = TestBase::new().await.with_file(("test.txt", "not an image")).await;
 
         let tool = ImageRead {
             paths: vec![test_base.join("test.txt").to_string_lossy().to_string()],
--- a/crates/agent/src/agent/tools/ls.rs
+++ b/crates/agent/src/agent/tools/ls.rs
@@ -450,10 +450,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_ls_validate_file_not_directory() {
-        let test_base = TestBase::new()
-            .await
-            .with_file(("file.txt", "content"))
-            .await;
+        let test_base = TestBase::new().await.with_file(("file.txt", "content")).await;
 
         let tool = Ls {
             path: test_base.join("file.txt").to_string_lossy().to_string(),
--- a/crates/agent/src/agent/util/mod.rs
+++ b/crates/agent/src/agent/util/mod.rs
@@ -142,7 +142,6 @@ pub fn is_integ_test() -> bool {
 mod tests {
     use super::*;
 
-
     #[test]
     fn test_truncate_safe() {
         assert_eq!(truncate_safe("Hello World", 5), "Hello");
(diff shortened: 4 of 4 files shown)


crates/agent/src/cli/run.rs
```diff
@@ -0,0 +1,240 @@
...
+                break;
+            }
+        }
+
+        agent
+            .send_prompt(SendPromptArgs {
+                content: vec![ContentChunk::Text(initial_prompt)],
+                should_continue_turn: None,
+            })
+            .await?;
```
dingfeli, line 125: Do we always assume there is an initial prompt?

crates/agent/src/cli/run.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    async fn handle_output_format_printing(&self, evt: &AgentEvent) -> Result<()> {
+        match self.output_format.unwrap_or(OutputFormat::Text) {
+            OutputFormat::Text => {
+                if let AgentEvent::Update(evt) = &evt {
+                    match &evt {
+                        UpdateEvent::AgentContent(ContentChunk::Text(text)) => {
+                            print!("{}", text);
+                            let _ = std::io::stdout().flush();
```
dingfeli, line 191: Is this subject to change? I see here we are just logging to stdout. Is the plan to instead write to some sort of listener that is in charge of eventually displaying it.

crates/agent/src/agent/protocol.rs
```diff
@@ -0,0 +1,309 @@
...
+    /// The user turn has ended. Metadata about the turn's execution is provided.
+    ///
+    /// This event is emitted in the following scenarios:
+    /// * The user turn has ended successfully
+    /// * The user cancelled the agent's execution
+    /// * The agent encountered an error, and the user sends a new prompt.
+    ///
+    /// Note that a turn can continue even after a [AgentEvent::Stop] for when the agent encounters
+    /// an error, and the next prompt chooses to continue the turn.
+    EndTurn(UserTurnMetadata),
```
dingfeli, line 61: Is a request - response considered a user turn? 
I see this is referenced in [RunArgs::main_loop], and it looks like this is a condition that would end the execution of `main_loop`.

dingfeli: Though I was only following the code path starting from `main.rs`. I am guessing that's only one use case (looks like non-interactive?).

crates/agent/src/agent/mcp/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+impl Default for McpManager {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum McpManagerRequest {
```
dingfeli, line 283: Should we also have a variant for Shutting down server. i.e. during unloading of an agent for example.

crates/agent/src/agent/mcp/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+                        self.handle_initializing_mcp_actor_event(name, evt).await;
+                    }
+                },
+                res = initialized_servers.next(), if !initialized_servers.is_empty() => {
+                    std::mem::drop(initializing_servers);
+                    std::mem::drop(initialized_servers);
+                    if let Some((name, evt)) = res {
+                        self.handle_mcp_actor_event(name, evt).await;
+                    }
+                },
```
dingfeli, line 195: Would the unordered future not be dropped anyway? Not that there is anything wrong with doing that explicitly.

crates/agent/src/cli/run.rs
```diff
@@ -0,0 +1,240 @@
...
+        if let Some(name) = &self.agent {
+            let (configs, _) = load_agents().await?;
+            if let Some(cfg) = configs.into_iter().find(|c| c.name() == name.as_str()) {
+                snapshot.agent_config = cfg.config().clone();
+            } else {
+                bail!("unable to find agent with name: {}", name);
+            }
+        };
+
+        let agent = Agent::new(snapshot, model, McpManager::new().spawn()).await?.spawn();
```
dingfeli, line 105: So the paradigm is that each `Agent` would have its own `McpManager`?

crates/agent/src/agent/mcp/mod.rs
dingfeli: Nvm, as I had discovered in https://github.com/aws/amazon-q-developer-cli/pull/3150/files#r2479586268, each agent is to have their own `McpManager`. There is no concept of unloading agent to `McpManager` since `McpManager` has the same lifetime as the associated agent. 

We wouldn't need a variant for unloading mcp server unless we support hot loading agent config in the future or something.

crates/agent/src/agent/mcp/actor.rs
```diff
@@ -0,0 +1,358 @@
...
+        args: Option<serde_json::Map<String, Value>>,
+    },
+}
+
+#[derive(Debug)]
+enum McpServerActorResponse {
+    Tools(Vec<ToolSpec>),
+    Prompts(Vec<Prompt>),
+    ExecuteTool(oneshot::Receiver<ExecuteToolResult>),
+}
```
dingfeli, line 126: I think missing from this is the prompt "hydration" result. `Prompt` here is just metadata about the prompt. The actual filled in prompt would need its own variant as well.

crates/agent/src/agent/agent_config/definitions.rs
```diff
@@ -0,0 +1,395 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct McpServers {
+    pub mcp_servers: HashMap<String, McpServerConfig>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(untagged)]
+pub enum McpServerConfig {
+    Local(LocalMcpServerConfig),
+    StreamableHTTP(StreamableHTTPMcpServerConfig),
```
dingfeli, line 221: I think we can just name this variant remote (as well as StreamableHTTPMcpServerConfig to just RemoteMcpServerConfig). We won't be differentiating bewteen SSE and Streamable HTTP from the config.

crates/agent/src/agent/agent_config/definitions.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// The URL endpoint for HTTP-based MCP servers
+    pub url: String,
+    /// HTTP headers to include when communicating with HTTP-based MCP servers
+    #[serde(default)]
+    pub headers: HashMap<String, String>,
+    /// Timeout for each mcp request in ms
+    #[serde(alias = "timeout")]
+    #[serde(default = "default_timeout")]
+    pub timeout_ms: u64,
+}
```
dingfeli, line 254: Missing here is the auth related attributes. I'll add that in once this merges.

crates/agent/src/agent/protocol.rs
brandonskiser: A user turn consists of all requests up until either a non-retryable error, or the model responds with no tool uses

crates/agent/src/agent/mcp/mod.rs
brandonskiser: There were borrow checker errors without the explicit drop for some reason - this is an area of the code that could definitely use some testing to ensure no messages are dropped (`FuturesUnordered` should be cancel safe so long as all of the futures it contains are cancel safe, which they are since `mpsc` is cancel safe)

I imagine there has to be a cleaner way to do this

crates/agent/src/cli/run.rs
brandonskiser: An agent has a handle to an MCP manager, meaning that multiple agents could potentially share the same MCP manager

crates/agent/src/agent/mcp/actor.rs
brandonskiser: What is prompt hydration? If you mean the request that actually expands the prompt, then yeah that wasn't included in this enum yet. It would be the same as the execute tool flow though

crates/agent/src/agent/agent_config/definitions.rs
brandonskiser: SSE + Streamable have the same exact config structure? Then yeah sounds good.

PR #3167, opened 2025-10-14
fix: improve settings list display

konippi:
*Issue #, if available:*
`N/A` (But I often hear users say they don't know how to view the full list of settings.)

*Description of changes:*
I think it would improve the UX if the q settings all command displayed all settings and only highlighted the ones that were changed as shown in the attached image.

image

### DIFF
--- a/crates/chat-cli/src/cli/settings.rs
+++ b/crates/chat-cli/src/cli/settings.rs
@@ -25,7 +25,20 @@ use crate::util::directories;
 pub enum SettingsSubcommands {
     /// Open the settings file
     Open,
-    /// List all available settings with their descriptions and current values
+    /// List settings
+    List {
+        /// Show all available settings
+        #[arg(long)]
+        all: bool,
+        /// Format of the output
+        #[arg(long, short, value_enum, default_value_t)]
+        format: OutputFormat,
+        /// Whether or not we want to modify state instead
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/settings.rs
```diff
@@ -24,7 +25,7 @@ use crate::util::directories;
 pub enum SettingsSubcommands {
     /// Open the settings file
     Open,
-    /// List all the settings
+    /// List all available settings with their descriptions and current values
```
dingfeli: I think the biggest change here is the fact that it lists all the settings rather than the formatting. 
Let me ask the team to see how people feel about that.

kkashilk: I'd be interested in knowing all the available settings, but would want a way to see just the ones that I have modified/enabled
```
q settings list
q settings list --all
```

where current `q settings all` = `q settings list`
and what the CR proposes would go under `q settings list --all`

konippi: Thank you for your feedback. I modified the command to display both all available settings and the settings users have enabled. 

```bash
# configured settings
$ q settings list

# all available settings
$ q settings list --all

# backward compatible (q settings all = q settings list)
$ q settings all
```

PR #3172, opened 2025-10-15
Update README.md to include instructions for installing via Homebrew

ke4qqq:
*Description of changes:* Homebrew is a much more user friendly way to install Q CLI 

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 
 - **macOS**:
   - **DMG**: [Download now](https://desktop-release.q.us-east-1.amazonaws.com/latest/Amazon%20Q.dmg)
+  - **HomeBrew**: ```brew install --cask amazon-q ```
 - **Linux**:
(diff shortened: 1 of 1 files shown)


dingfeli:
Hi, thanks for your contribution! 
Unfortunately we don't actually own the brew installation (AFAIK nobody on our team set this up). 
So while users are free to install it however they see fit, we don't want to officially sanction this way of installing since it can just be changed without our knowledge.

ke4qqq:
It's in the official documentation

https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-installing.html#command-line-installing-homebrew

dingfeli:
After some discussion with the team we have found that the installation set up was actually done by a previous team member. Thanks for bringing this to our attention!

dingfeli (approved):

brandonskiser (approved):

PR #3180, opened 2025-10-16
fix(mcp): correct environment variable precedence in stdio transport

abaw:
- Fix issue where parent process env vars overwrite custom MCP server env vars
- Ensure custom env vars from agent config take precedence over parent env
- Resolves bug introduced in RMCP migration (commit 1813bb77)

Fixes environment variable configuration not taking effect for MCP servers when specified in agent configuration files.

*Issue #, if available:*

*Description of changes:*
I found custom environment variable AWS_PROFILE set for MCP server is not applied to my MCP server after some update of Q CLI. After checking the code, I thought it's a bug introduced during RMCP migration. The change should work the same way as before the RMCP migration.


mschrage (approved):

mschrage:
Thanks for this PR!

Issue #3185, opened 2025-10-17
ctrl+s shows "/hooks help" while it does not exist


d-chechelashvili:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.7.1 (24G231)

### Expected behaviour

/hooks help should not be part of the list of commands or it should be a valid command.

### Actual behaviour

`Ctrl(^) + s` shows list of commands in which "/hooks help" appears. When I run it I get `error: unexpected argument 'help' found`

### Steps to reproduce

1. Run `Ctrl(^) + s`
2. See that `/hooks help` is there
3. Run `/hooks help`
4. See the error: 
```

error: unexpected argument 'help' found

Usage: /hooks

For more information, try '--help'.

Options:
  -h, --help  Print help (see more with '--help')
```

### Environment

```yaml
<This will be visible to anyone. Do not include personal or sensitive information>

[q-details]
version = "1.18.1"
hash = "5610605baef17b76ef41448f2e943f021c2b2273"
date = "2025-10-13T22:31:07.064465Z (3d ago)"

[system-info]
os = "macOS 15.7.1 (24G231)"
chip = "Apple M3 Pro"
total-cores = 12
memory = "36.00 GB"

[environment]
cwd = "/Users/USER"
cli-path = "/Users/USER"
install-method = "unknown"

[env-vars]
[...]

PR #3206, opened 2025-10-18
refactor: dynamically generate command list

konippi:
*Issue #, if available:*
#3185 

*Description of changes:*
Replace static `COMMANDS` array with dynamic generation to automatically sync available commands with SlashCommand
enum definition. This prevents inconsistencies like #3185 where unimplemented commands (e.g., `/hooks help`) appeared in the command list.

### DIFF
--- a/crates/chat-cli/src/cli/chat/prompt.rs
+++ b/crates/chat-cli/src/cli/chat/prompt.rs
@@ -98,11 +98,13 @@ impl PasteState {
     }
 }
 
-/// Cache for dynamically generated commands
+/// Cache for static slash commands from SlashCommand enum
 static COMMANDS_CACHE: OnceLock<Vec<String>> = OnceLock::new();
 
-/// Generate command list from SlashCommand enum using clap introspection
-fn generate_commands() -> &'static Vec<String> {
+/// Returns a cached list of static slash commands from SlashCommand enum.
+/// This list is generated once at first call and never changes during runtime.
+/// Does NOT include experiment-based commands.
+fn get_static_slash_commands() -> &'static Vec<String> {
     COMMANDS_CACHE.get_or_init(|| {
         use clap::CommandFactory;
 
@@ -149,10 +151,16 @@ fn generate_commands() -> &'static Vec<String> {
     })
 }
 
-/// Generate dynamic command list including experiment-based commands when enabled
+/// Returns the complete list of available commands including:
+/// - Static slash commands (cached from SlashCommand enum)
+/// - Experiment-based commands (dynamically evaluated based on current experiment status)
 pub fn get_available_commands(os: &Os) -> Vec<&'static str> {
-    let mut commands: Vec<&'static str> = generate_commands().iter().map(|s| s.as_str()).collect();
+    // Get static slash commands (cached)
+    let mut commands: Vec<&'static str> = get_static_slash_commands().iter().map(|s| s.as_str()).collect();
+
+    // Add experiment-based commands (dynamically evaluated)
     commands.extend(ExperimentManager::get_commands(os));
+
     commands.sort();
     commands
 }
--- a/crates/chat-cli/src/cli/chat/cli/mod.rs
+++ b/crates/chat-cli/src/cli/chat/cli/mod.rs
@@ -123,7 +123,7 @@ pub enum SlashCommand {
     )]
     Checkpoint(CheckpointSubcommand),
     /// View, manage, and resume to-do lists
-    #[command(subcommand)]
+    #[command(subcommand, hide = true)]
     Todos(TodoSubcommand),
     /// Paste an image from clipboard
     Paste(PasteArgs),


crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -54,58 +55,60 @@ use crate::database::settings::Setting;
...
+            // Add the main command
+            commands.push(format!("/{}", name));
+
+            // Add aliases
+            for alias in subcommand.get_all_aliases() {
+                commands.push(format!("/{}", alias));
+            }
+
+            // If this command has subcommands, add them too
+            if subcommand.has_subcommands() {
```
kkashilk, line 133: Does this cover the case `/context show --expand`

crates/chat-cli/src/cli/chat/prompt.rs
```diff
@@ -115,13 +118,13 @@ pub type PromptQuerySender = tokio::sync::broadcast::Sender<PromptQuery>;
...
 /// Complete commands that start with a slash
-fn complete_command(commands: Vec<&'static str>, word: &str, start: usize) -> (usize, Vec<String>) {
+fn complete_command(commands: Vec<&str>, word: &str, start: usize) -> (usize, Vec<String>) {
     (
         start,
         commands
             .iter()
             .filter(|p| p.starts_with(word))
-            .map(|s| (*s).to_owned())
+            .map(|s| s.to_string())
```
kkashilk: Also, seeing this clippy error - 
```
error: calling `to_string` on `&&str`
   --> crates/chat-cli/src/cli/chat/prompt.rs:127:22
    |
127 |             .map(|s| s.to_string())
    |                      ^^^^^^^^^^^^^ help: try dereferencing the receiver: `(*s).to_string()`
    |
    = help: `&str` implements `ToString` through a slower blanket impl, but `str` has a fast specialization of `ToString`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string
    = note: `-D clippy::inefficient-to-string` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::inefficient_to_string)]`

error: could not compile `chat_cli` (lib) due to 1 previous error
```

Can you run - 

```
cargo clippy --locked --workspace --color always -- -D warnings
cargo +nightly fmt
```

crates/chat-cli/src/cli/chat/prompt.rs
kkashilk: Looks like we're also missing these now - 
```
"/hooks add",
"/hooks rm",
"/hooks enable",
"/hooks disable",
"/hooks enable-all",
"/hooks disable-all",
```

kkashilk: I'd recommend adding the existing commands as tests here - https://github.com/aws/amazon-q-developer-cli/blob/b3fa2264bb2010d7c7e6f3cbe7be67a5c5b3c168/crates/chat-cli/src/cli/chat/prompt.rs#L836

Don't need to commit all of those in, only for local testing

konippi: > Does this cover the case /context show --expand

Why does it only support the “/context show --expand” option? It should only support slash commands and their subcommands.

> Looks like we're also missing these now -
  ```bash
  "/hooks add",
  "/hooks rm",
  "/hooks enable",
  "/hooks disable",
  "/hooks enable-all",
  "/hooks disable-all",
  ```
These subcommands should not exist at present.

konippi: @kkashilk
Any updates?

kensave: Can we revisit this with latest experiment manager logic which enables or disables command based on experiment status?
```
/// Generate dynamic command list including experiment-based commands when enabled
pub fn get_available_commands(os: &Os) -> Vec<&'static str> {
    let mut commands = COMMANDS.to_vec();
    commands.extend(ExperimentManager::get_commands(os));
    commands.sort();
    commands
}
```

kensave:
Please make sure to run:
`cargo clippy --workspace --color always -- -D warning` 

Then fix the clippy errors.

crates/chat-cli/src/cli/chat/prompt.rs
konippi: I confirmed that the display can be changed according to whether the experimental feature is enabled/disabled.

https://github.com/user-attachments/assets/3c522c07-c264-47c6-9b08-8800ab926fdd

konippi: Incidentally, since there was no `hide = true` setting in `/todo`, it was always displayed. Therefore, this PR adds `hide = true` to it so that when disabled via the experimental feature, it won't appear when pressing `Ctrl+S`.

Issue #2912, opened 2025-09-17
/save <path>  results in "Session paths are not serialized"


rahulpyd:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

Linux 6.1.0-39-amd64 - Debian GNU/Linux 12 (bookworm)

### Expected behaviour

> /context add <some/path>
> /save <path>  -- should save serialized conversation to <path>

### Actual behaviour

> /context add <some/path>
> /save <path> -- Failed to export to chatservice: Session paths are not serialized

### Steps to reproduce

q chat
> /context add <some/path>
> /save <path>

### Environment

```yaml
<This will be visible to anyone. Do not include personal or sensitive information>

[q-details]
version = "1.15.0"
hash = "a8a0426eaff32ac9b1599ca1b1f1881ffd16054d"
date = "2025-09-02T17:17:19.458949Z (15d ago)"

[system-info]
chip = "Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz"
total-cores = 2
memory = "15.53 GB"

[system-info.os.linux]
kernel_version = "6.1.0-39-amd64"
id = "debian"
name = "Debian GNU/Linux"
pretty_name = "Debian GNU/Linux 12 (bookworm)"
version_id = "12"
version = "12 (bookworm)"

[environment]
cwd = "/home/USER"
cli-path = "/home/USER"
install-method = "unknown"

[env-vars]
DISPLAY = ":0"
[...]

Issue #2652, opened 2025-08-20
Unable to save chat


944750720:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

macOS 15.5.0 (24F74)

### Expected behaviour

Be able to save chat by `/save .` Does it mean my chat history conclude something can't be serialized in json file?

### Actual behaviour

```
[default-agent] > /save .


Failed to export to .: Session paths are not serialized
```

### Environment

```yaml
<This will be visible to anyone. Do not include personal or sensitive information>

[q-details]
version = "1.13.3"
hash = "0c82fc60be46f6a51eff3da52861611e534e39a6"
date = "2025-08-11T21:20:15.725305Z (9d ago)"

[system-info]
os = "macOS 15.5.0 (24F74)"
chip = "Apple M1"
total-cores = 8
memory = "16.00 GB"

[environment]
cwd = "/Users/USER/Code/gemini-cli"
cli-path = "/Users/USER/Code/gemini-cli"
install-method = "unknown"

[env-vars]
DISPLAY = "/private/tmp/com.apple.launchd.HHcUHSKAaf/org.xquartz:0"
[...]

PR #3217, opened 2025-10-20
fix: save convo with context

dingfeli:
*Issue #, if available:*
https://github.com/aws/amazon-q-developer-cli/issues/2652
https://github.com/aws/amazon-q-developer-cli/issues/2912

*Description of changes:*
As titled. 
Note that when loading convo with saved context, we are going to do the following:
- Examine to see if file path is in the current agent's permanent scope
- If it is NOT present, move it to the current agent's session paths 

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/persist.rs
+++ b/crates/chat-cli/src/cli/chat/cli/persist.rs
@@ -5,6 +5,7 @@ use crossterm::style::{
 };
 
 use crate::cli::ConversationState;
+use crate::cli::chat::context::ContextFilePath;
 use crate::cli::chat::{
     ChatError,
     ChatSession,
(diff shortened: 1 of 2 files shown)


crates/chat-cli/src/cli/chat/cli/persist.rs
```diff
@@ -102,6 +103,21 @@ impl PersistSubcommand {
...
+                // And we'll place them as temporary context
+                // Note that we are NOT doing the same with hooks because hooks are more
+                // instrinsically linked to agent and it affects the behavior of an agent
+                if let Some(cm) = &new_state.context_manager {
+                    if let Some(existing_cm) = &mut session.conversation.context_manager {
+                        let existing_paths = &mut existing_cm.paths;
+                        for incoming_path in &cm.paths {
+                            if !existing_paths.contains(incoming_path) {
+                                existing_paths
+                                    .push(ContextFilePath::Session(incoming_path.get_path_as_str().to_string()));
```
kkashilk, line 116: we should also check if incoming_path was a session-path?

dingfeli: They are all deserialized as Agent path actually. So it is included in here.

evanliu048 (approved):

PR #3268, opened 2025-10-25
feat: Path based loading for agent

mekza:
New version of #2859 after fixing the tests:

This PR transforms how users organize and reference their Q CLI agents by introducing hierarchical directory support. Previously, all agents lived in a flat directory structure, forcing users to rely solely on unique names to avoid collisions. Now, agents can be organized in subdirectories and referenced using intuitive path notation.

```
cli-agents/
  team/
    backend-expert.json
    frontend-specialist.json
  personal/
    code-reviewer.json
  shared/
    documentation-helper.json
```

It enables users to reference agents as `q chat --agent team/backend-expert` in addition to `q chat --agent backend-expert`. Users can now organize agents hierarchically in the Q CLI directory structure, avoiding naming collisions and improving discoverability through logical grouping (e.g., team-specific, project-specific, or role-specific agent directories).

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -10,10 +10,7 @@ use std::collections::{
     HashSet,
 };
 use std::ffi::OsStr;
-use std::io::{
-    self,
-    Write,
-};
+use std::io::Write;
 use std::path::{
     Path,
     PathBuf,
@@ -37,7 +34,6 @@ use serde::{
     Serialize,
 };
 use thiserror::Error;
-use tokio::fs::ReadDir;
 use tracing::{
     error,
     info,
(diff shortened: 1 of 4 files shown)


adamalrefai:
Does this PR also add support for loading an agent with a path to the config file? So for example `q chat --agent /path/to/agent.json`

mekza:
@adamalrefai not full path only path relative to `$HOME/.aws/amazonq/cli-agents`

adamalrefai:
@mekza Alright, got it :) This is really good and really needed, but was hoping that this will also add support for global path picking.. I think that should be brought up, the /agent swap should also support that in my opinion, like /agent swap --path /path/to/agent.json

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -476,14 +488,22 @@ impl Agents {
         self.agents.get_mut(&self.active_idx)
     }
 
-    pub fn switch(&mut self, name: &str) -> eyre::Result<&Agent> {
-        if !self.agents.contains_key(name) {
+    pub fn switch(&mut self, name: &str, os: &Os) -> eyre::Result<&Agent> {
```
dingfeli: Can we switch the args position around have os to be the first thing (after self)?

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -476,14 +488,22 @@ impl Agents {
...
+        // Find agent by either JSON name or path identifier
+        let matching_key = self
+            .agents
+            .iter()
+            .find(|(_, agent)| agent.name.as_str() == name || agent.path_identifier(os).as_deref() == Some(name))
```
dingfeli:
```rust
.find(|(_, agent)| agent.name.as_str() == name || agent.path_identifier(os).as_deref().is_some_and(|n| n == name))
```

crowecawcaw:
I'm interested in this feature. I saw that it was previously merged and reverted. Is this change still intended to be added?

mekza:
@crowecawcaw I will resume work on that pretty soon.

PR #3305, opened 2025-10-29
add auth login failure telemetry

abhraina-aws:
We didnt have good telemetry for auth login failures. Added it here. 

### DIFF
--- a/crates/chat-cli/src/auth/builder_id.rs
+++ b/crates/chat-cli/src/auth/builder_id.rs
@@ -302,7 +302,10 @@ impl BuilderIdToken {
     }
 
     /// Load the token from the keychain, refresh the token if it is expired and return it
-    pub async fn load(database: &Database) -> Result<Option<Self>, AuthError> {
+    pub async fn load(
+        database: &Database,
+        telemetry: Option<&crate::telemetry::TelemetryThread>,
+    ) -> Result<Option<Self>, AuthError> {
         // Can't use #[cfg(test)] without breaking lints, and we don't want to require
         // authentication in order to run ChatSession tests. Hence, adding this here with cfg!(test)
         if cfg!(test) {
@@ -328,7 +331,7 @@ impl BuilderIdToken {
 
                         if token.is_expired() {
                             trace!("token is expired, refreshing");
-                            token.refresh_token(&client, database, &region).await
+                            token.refresh_token(&client, database, &region, telemetry).await
                         } else {
                             trace!(?token, "found a valid token");
                             Ok(Some(token))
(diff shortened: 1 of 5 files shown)


crates/chat-cli/src/auth/builder_id.rs
```diff
@@ -357,6 +360,7 @@ impl BuilderIdToken {
         client: &Client,
         database: &Database,
         region: &Region,
+        telemetry: Option<&crate::telemetry::TelemetryThread>,
```
dingfeli, line 363: Does this need to be an Option type? Do we not want to always emit something when token refresh fails?

crates/chat-cli/src/auth/builder_id.rs
```diff
@@ -486,6 +505,14 @@ impl BuilderIdToken {
     }
 }
 
+pub fn token_type_from_start_url(start_url: Option<&str>) -> TokenType {
```
dingfeli, line 508: Maybe just implement an `From<Option<&str>>` for `TokenType` here. We own the definition of `TokenType` in this crate so you don't have to worry about orphan rule.

crates/chat-cli/src/auth/builder_id.rs
abhraina-aws: Its because of the load function. There are cases when we can pass None to it for telemetry because its not a login failure case. And hence that gets passed down. So we just implemented it to be able to handle a None for telemetry.

dingfeli: Oh alright. I'll let the second PR chime in here. Not familiar enough with the auth flow to have more insights here.

abhraina-aws: This is more like a code arrangement thing now. Either I handle the None in one place vs the other.

crates/chat-cli/src/auth/builder_id.rs
abhraina-aws: You are right. Will take care of this.

crates/chat-cli/src/auth/builder_id.rs
kkashilk: imo, instead of making this an Option, you can just check if there's an error and only then publish the auth-failed?

Also, do you wanna publish an auth-succeeded metric?

abhraina-aws: So the issue here is that load is being called in a-lot of places which may not necessarily need telemetry publish.  So instead of adding the code everywhere to have telemetry to pass, I am allowing a None. We are already only posting auth_failed only when we have an error. 

We already have a login successful metric. So that covers it.

abhraina-aws:
https://github.com/aws/amazon-q-developer-cli/pull/3317/files

PR #3306, opened 2025-10-29
fix: Restrict profile command to idc users only

evanliu048:
*Issue #, if available:*

*Description of changes:*
Updated profile command to check user type before calling service
Non-IDC users now receive clear error message: "This command is only available for IAM Identity Center users"
image


crates/chat-cli/src/auth/builder_id.rs
dingfeli: I guess this function really just does not need to return a `Result` anymore right.

Issue #2834, opened 2025-09-10
Command chaining bypasses allowedCommands restrictions


en0:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

Linux 6.14.0-29-generic - Ubuntu 24.04.3 LTS

### Expected behaviour

When allowedCommands is configured to only allow specific commands like notify-send, command chaining with semicolons should be blocked or restricted to prevent execution of unauthorized commands.

### Actual behaviour

Commands chained with semicolons (e.g., "notify-send 'hello' 'world'; touch ~/did-it-work") execute successfully, allowing unauthorized commands to run alongside allowed ones. Both the allowed notify-send and the unauthorized touch command executed.

```
[chat-transcript]
> use notify send to tell me a joke.


[Tool uses: execute_bash]

There you go! A classic programming joke delivered via desktop notification.
[Tool uses: none]

> ok so the allowedCommands worked. Can you try one more thing for me. im curious what happens. When you run the execute_bash tool, can you do so with this command?  notify-send "hello" "world"; echo this shouldnt be allowed.


[Tool uses: execute_bash]

Interesting! Both commands executed successfully. The notification was sent and the echo command ran as well, outputting "this shouldnt be allowed". It appears the command chaining with semicolons worked as expected in a normal bash environment.
[Tool uses: none]

> i din't see the echo. lets use this instead.... notify-send "hello" "world"; touch ~/did-it-work


[Tool uses: execute_bash]

Now let's check if the file was created:
[Tool uses: fs_read]

Perfect! I can see the file `did-it-work` was created at `/home/USER/did-it-work` with a timestamp of Sep 10 20:24. Both commands in the chained sequence executed successfully - the notification was sent and the file was created.
[Tool uses: none]

> yikes... lets file a bug


[Tool uses: report_issue]
```

### Steps to reproduce

<This will be visible to anyone. Do not include personal or sensitive information>

1. Configure allowedCommands to restrict execute_bash to specific commands like notify-send
2. Execute a command with semicolon chaining: "notify-send 'hello' 'world'; touch ~/did-it-work"
3. Observe that both commands execute successfully, bypassing the restriction

### Environment

```yaml
[q-details]
version = "1.15.0"
[...]

PR #3313, opened 2025-10-29
fix: block dangerous patterns for tool calls

ekang7:
*Issue #, if available:*
https://github.com/aws/amazon-q-developer-cli/issues/2834

*Description of changes:*
Prioritized blocking dangerous patterns for tool calls above allowed commands regex. Removed test for the case where we formerly prioritized allowed commands over blocking dangerous patterns. Added unit tests for blocking dangerous patterns. 

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
+++ b/crates/chat-cli/src/cli/chat/tools/execute/mod.rs
@@ -89,7 +89,6 @@ impl ExecuteCommand {
         if !current_cmd.is_empty() {
             all_commands.push(current_cmd);
         }
-        
 
         // Check if each command in the pipe chain starts with a safe command
         for cmd_args in &all_commands {


crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -100,6 +89,17 @@ impl ExecuteCommand {
...
+
+        let has_regex_match = allowed_commands
+            .iter()
+            .map(|cmd| Regex::new(&format!(r"\A{}\z", cmd)))
+            .filter(Result::is_ok)
+            .flatten()
+            .any(|regex| regex.is_match(&self.command));
+        if has_regex_match {
+            return false;
+        }
```
evanliu048: Should we move the regex check after the special validations for `find` and `grep` commands?

ekang7: I concur with this choice, moving the regex check for allowed commands after the special validations for find and grep commands would maximize security because then only if we have deemed that the special malicious cases for find and grep have not occurred do we then check for the allowed commands. I'll go ahead and make the edit and push again.

ekang7: Fixed!

crates/chat-cli/src/cli/chat/tools/execute/mod.rs
```diff
@@ -129,12 +118,29 @@ impl ExecuteCommand {
                     {
                         return true;
                     }
-                    let is_cmd_read_only = READONLY_COMMANDS.contains(&cmd.as_str());
-                    if !allow_read_only || !is_cmd_read_only {
-                        return true;
-                    }
                 },
-                None => return true,
+                None => {},
```
evanliu048, line 122: why we remove `None => return true,` here?

ekang7: I think we need to do this because otherwise, it will incorrectly return true early if any of the commands in the pipe chain do not start with a safe/special command.

evanliu048: The None => branch handles cases where the command list is empty — isn't that potentially suspicious?

evanliu048: But I guess it's ok here since we only add vec in `all_commands` when vec is not empty
https://github.com/aws/amazon-q-developer-cli/pull/3313/files#diff-7625c9845cf1c9abd5360e26c1c62f195e0efbd5054a1d36b54f20441734a127L100-R90

evanliu048 (approved):

PR #3315, opened 2025-10-29
refactor: centralize environment variable access

dk19y:
Centralize environment variable access through new module with semantic wrapper functions.

### DIFF
--- a/crates/chat-cli/src/util/env_var.rs
+++ b/crates/chat-cli/src/util/env_var.rs
@@ -1,97 +1,92 @@
-use std::ffi::OsString;
-
 use crate::os::Env;
+use crate::util::consts::env_var::*;
 
-/// Get environment variable as String
-pub fn get_var(key: &str) -> Result<String, std::env::VarError> {
-    std::env::var(key)
+/// Get log level from environment
+pub fn get_log_level(env: &Env) -> Result<String, std::env::VarError> {
+    env.get(Q_LOG_LEVEL)
 }
 
-/// Get environment variable as OsString
-pub fn get_var_os(key: &str) -> Option<OsString> {
-    std::env::var_os(key)
-}
-
-/// Get environment variable with default value
-pub fn get_var_or(key: &str, default: &str) -> String {
-    std::env::var(key).unwrap_or_else(|_| default.to_string())
-}
-
-// Semantic helpers that abstract away the actual variable names
-pub fn get_log_level() -> Result<String, std::env::VarError> {
-    get_var("Q_LOG_LEVEL")
-}
-
+/// Get chat shell with default fallback
 #[cfg(unix)]
 pub fn get_chat_shell() -> String {
-    get_var_or("AMAZON_Q_CHAT_SHELL", "bash")
+    Env::new()
+        .get(AMAZON_Q_CHAT_SHELL)
+        .unwrap_or_else(|_| "bash".to_string())
 }
 
+/// Check if stdout logging is enabled
 pub fn is_log_stdout_enabled() -> bool {
-    get_var_os("Q_LOG_STDOUT").is_some()
+    Env::new().get_os(Q_LOG_STDOUT).is_some()
 }
 
+/// Check if telemetry is disabled
 pub fn is_telemetry_disabled() -> bool {
-    get_var_os("Q_DISABLE_TELEMETRY").is_some()
+    Env::new().get_os(Q_DISABLE_TELEMETRY).is_some()
 }
 
-pub fn get_mock_chat_response(env: Option<&Env>) -> Option<String> {
-    match env {
-        Some(e) => e.get("Q_MOCK_CHAT_RESPONSE").ok(),
-        None => get_var("Q_MOCK_CHAT_RESPONSE").ok(),
-    }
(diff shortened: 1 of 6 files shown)


crates/chat-cli/src/util/env_var.rs
```diff
@@ -0,0 +1,97 @@
+use std::ffi::OsString;
+
+use crate::os::Env;
+
+/// Get environment variable as String
+pub fn get_var(key: &str) -> Result<String, std::env::VarError> {
```
brandonskiser: Environment variables should only be retrieved from `Env`. These functions don't really accomplish anything other than wrapping `std` for no reason

crates/chat-cli/src/util/env_var.rs
```diff
@@ -0,0 +1,97 @@
...
+
+pub fn is_log_stdout_enabled() -> bool {
+    get_var_os("Q_LOG_STDOUT").is_some()
+}
+
+pub fn is_telemetry_disabled() -> bool {
+    get_var_os("Q_DISABLE_TELEMETRY").is_some()
+}
+
+pub fn get_mock_chat_response(env: Option<&Env>) -> Option<String> {
```
brandonskiser: I wouldn't parameterize this as an `Option`, instead the caller can just do `get_mock_chat_response(Env::new())`

crates/chat-cli/src/util/env_var.rs
```diff
@@ -0,0 +1,97 @@
...
+    get_var("Q_LOG_LEVEL")
+}
+
+#[cfg(unix)]
+pub fn get_chat_shell() -> String {
+    get_var_or("AMAZON_Q_CHAT_SHELL", "bash")
+}
+
+pub fn is_log_stdout_enabled() -> bool {
+    get_var_os("Q_LOG_STDOUT").is_some()
```
brandonskiser: I think we should define the env vars as consts at least - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/util/consts.rs#L32

crates/chat-cli/src/util/env_var.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// Get environment variable with default value
+pub fn get_var_or(key: &str, default: &str) -> String {
+    std::env::var(key).unwrap_or_else(|_| default.to_string())
+}
+
+// Semantic helpers that abstract away the actual variable names
+pub fn get_log_level() -> Result<String, std::env::VarError> {
+    get_var("Q_LOG_LEVEL")
```
brandonskiser: This is already defined here - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/util/consts.rs#L49

crates/chat-cli/src/util/env_var.rs
```diff
@@ -0,0 +1,92 @@
...
+    Env::new().get(AWS_REGION)
+}
+
+/// Check if SigV4 authentication is enabled
+pub fn is_sigv4_enabled(env: &Env) -> bool {
+    env.get(AMAZON_Q_SIGV4).is_ok_and(|v| !v.is_empty())
+}
+
+/// Get all environment variables
+pub fn get_all_env_vars() -> std::env::Vars {
```
brandonskiser, line 85: This doesn't seem necessary if all we're doing is wrapping `std`

crates/chat-cli/src/cli/settings.rs
```diff
@@ -190,12 +189,10 @@ impl SettingsArgs {
         match self.cmd {
             Some(SettingsSubcommands::Open) => {
                 let file = GlobalPaths::settings_path().context("Could not get settings path")?;
-                if let Ok(editor) = os.env.get("EDITOR") {
-                    tokio::process::Command::new(editor).arg(file).spawn()?.wait().await?;
-                    Ok(ExitCode::SUCCESS)
-                } else {
-                    bail!("The EDITOR environment variable is not set")
-                }
+                let editor =
```
brandonskiser, line 192: I get this is unrelated but I think we could easily fix a bug here by using this util function for launching the editor I think - https://github.com/aws/amazon-q-developer-cli/blob/main/crates/chat-cli/src/util/editor.rs#L15

crates/chat-cli/src/cli/chat/tools/mod.rs
```diff
@@ -444,9 +444,9 @@ fn format_path(cwd: impl AsRef<Path>, path: impl AsRef<Path>) -> String {
         .unwrap_or(path.as_ref().to_string_lossy().to_string())
 }
 
-fn supports_truecolor(os: &Os) -> bool {
+fn supports_truecolor() -> bool {
```
brandonskiser, line 447: Why is `os` being removed from this?

dk19y: `os` is being removed as its we rely on env vars util method now

crates/chat-cli/src/cli/settings.rs
dk19y: Will have a followup for this when i wrap up initial stuff

PR #3318, opened 2025-10-30
feat: UX updates

zixlin7:
*Issue #, if available:*

*Description of changes:*
- UX updates to colors, strings, etc.

### DIFF
--- a/crates/chat-cli/src/theme/colors.rs
+++ b/crates/chat-cli/src/theme/colors.rs
@@ -2,6 +2,16 @@
 
 use crossterm::style::Color;
 
+// ANSI color value constants
+/// Purple brand color
+pub const BRAND_PURPLE: u8 = 141;
+/// Light grey for primary text
+pub const LIGHT_GREY: u8 = 252;
+/// Medium-light grey for secondary text
+pub const MEDIUM_LIGHT_GREY: u8 = 248;
+/// Bright magenta for emphasis and current items
+pub const BRIGHT_MAGENTA: u8 = 13;
+
 /// Colors for status messages and feedback
 #[derive(Debug, Clone)]
 pub struct StatusColors {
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/chat/cli/checkpoint.rs
```diff
@@ -91,7 +91,9 @@ impl CheckpointSubcommand {
             execute!(
                 session.stderr,
                 StyledText::error_fg(),
-                style::Print("\nCheckpoint is disabled. Enable it with: kiro-cli settings chat.enableCheckpoint true\n"),
+                style::Print(
+                    "\nCheckpoint is disabled. Enable it with: kiro-cli settings chat.enableCheckpoint true\n"
```
brandonskiser: The binary name should be formatted using a const - https://github.com/aws/amazon-q-developer-cli/blob/qv2/crates/chat-cli/src/util/consts.rs#L2

crates/chat-cli/src/cli/chat/cli/persist.rs
```diff
@@ -14,7 +14,7 @@ use crate::cli::chat::{
 use crate::os::Os;
 use crate::theme::StyledText;
 
-/// Commands for persisting and loading conversation state
+/// Use any of these commands to manage your Kiro session. All commands start with '/'.
```
brandonskiser: Why is this being changed?

> All commands start with '/'.

This is true for all slash commands right? No need to call it out for save/load specifically I think

crates/chat-cli/src/constants.rs
```diff
@@ -133,15 +106,28 @@ pub mod ui_text {
         ));
         help.push_str(&format!(
             "\n                    {}",
-            StyledText::secondary("Change using: kiro-cli settings chat.skimCommandKey x")
+            StyledText::secondary("Change these keybinding at anytime using: kiro-cli settings chat.skimCommandKey x")
```
brandonskiser: Also use const for binary name here?

crates/chat-cli/src/cli/chat/cli/tangent.rs
```diff
@@ -53,7 +53,9 @@ impl TangentArgs {
             execute!(
                 session.stderr,
                 StyledText::error_fg(),
-                style::Print("\nTangent mode is disabled. Enable it with: kiro-cli settings chat.enableTangentMode true\n"),
+                style::Print(
+                    "\nTangent mode is disabled. Enable it with: kiro-cli settings chat.enableTangentMode true\n"
```
brandonskiser: binary name here

crates/chat-cli/src/constants.rs
```diff
@@ -28,7 +28,7 @@ pub mod ui_text {
 
     /// Changelog header text
     pub fn changelog_header() -> String {
-        format!("{}\n\n", StyledText::emphasis("What's New in Kiro CLI"))
+        format!("{}\n\n", StyledText::primary("✨ What's New in Kiro CLI"))
```
brandonskiser: I see we also interpolate the brand name using a const in other places as well, figure we want to be consistent and also use a const here?

crates/chat-cli/src/theme/colors.rs
```diff
@@ -61,21 +63,23 @@ impl Default for StatusColors {
 impl Default for UiColors {
     fn default() -> Self {
         Self {
-            primary_brand: Color::Cyan,
-            primary_text: Color::White,
-            secondary_text: Color::DarkGrey,
-            emphasis: Color::Magenta,
-            command_highlight: Color::Green,
+            primary_brand: Color::AnsiValue(141),
```
brandonskiser: Shouldn't these be constants? instead of just numbers

PR #3324, opened 2025-10-30
Refactor /Usage into smaller modules

akhileshb:
As we plan to expand functionality of Usage, splitting the single file into three parts1
1. Data Provider to fetch the data from Chat Session or other systems
2. Rendering to print the UI
3. Main/Orchestrator for co-ordination

Note: This is purely a refactoring PR with NO functional changes made (Code has been cut/paste as-is)


crates/chat-cli/src/cli/chat/cli/usage/usage_renderer.rs
```diff
@@ -0,0 +1,182 @@
+use crossterm::style::Attribute;
+use crossterm::{execute, queue, style};
+
+use crate::cli::chat::token_counter::TokenCount;
+use crate::cli::chat::{ChatError, ChatSession};
+use crate::theme::StyledText;
+
+/// Calculate usage percentage from token counts (private utility)
+fn calculate_usage_percentage(tokens: TokenCount, context_window_size: usize) -> f32 {
```
abhraina-aws, line 9: If we are separating this. Would it make sense to move this calculation to the data provider and just have functions there to call? I think all the data for these calculations is anyways coming from there.

akhileshb: Thats a fare point Abhay. My first version of code indeed had that :) 

However, I wanted to avoid dependency between provider and renderer directly. Communication should just be in the form of models. The right way to do this would be to populate even percentage in the "DetailedUsageData" model.

Didn't want to introduce that reduce the changes being made

abhraina-aws: Yes I meant calculate_usage_percentage -> usagedata.calculate_tool_token_usage_percentage()
We just call that like we are calling everything else in the renderer file. All the data is already in the provider.

abhraina-aws: Had an offline discussion. Aligned future approach on further refactor. Basically might create a util to hold calculation to separate them.

Issue #3271, opened 2025-10-26
Amazon Q freezes on triple backticks input


sjhung:
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

Linux 6.14.0-33-generic - Ubuntu 24.04.3 LTS

### Expected behaviour

<This will be visible to anyone. Do not include personal or sensitive information>

Should handle triple backticks normally and provide a response

### Actual behaviour

<This will be visible to anyone. Do not include personal or sensitive information>

When user sends only ``` (three backticks), Amazon Q stalls and doesn't respond

```
[chat-transcript]
> you can;t readn "`"

I can read backticks (`) just fine. Could you clarify what specific issue you're experiencing with backticks? Are you having trouble with:

- Code formatting in my responses?
- Reading code blocks I've provided?
- Something else related to backtick characters?

Let me know what you need help with and I'll address it directly.
[Tool uses: none]

> you stall with three "`"

I understand - you're saying I stall or have issues when encountering three backticks (\```). 

I can process triple backticks normally - they're used for code blocks in markdown. If you're seeing me stall or have problems with them, that might be a technical issue.

What specifically happens when I encounter \```? Do I:
- Stop responding mid-message?
- Take too long to respond?
- Give incomplete responses?

This could be worth reporting as a technical issue if it's consistently happening.
[Tool uses: none]

> I just did . you got stuck

You're right - I did get stuck when you sent just the three backticks. That's a clear technical issue where I'm not handling that input properly.

Would you like me to report this as a bug? I can use the issue reporting tool to document this behavior where sending just "\```" causes me to stall or not respond correctly.
[Tool uses: none]

> yes it's a major bug


[Tool uses: report_issue]

Issue reported. This is definitely a significant bug that affects basic functionality - triple backticks are common in developer workflows for code blocks.
[Tool uses: none]

> you just opened a web page

Yes, the report_issue tool opens a GitHub issue template in your browser with the bug details pre-filled. That's how the issue reporting system works - it creates a web form for you to submit the bug report to the development team.

You can review the details I filled in and submit it, or modify anything before submitting.
[Tool uses: none]

> no.. it shows login screen

That makes sense - GitHub requires you to be logged in to create issues. You'll need to:

[...]

PR #3327, opened 2025-10-30
added hint for multiline triple backtick code block detector

ekang7:
*Issue #, if available:*
https://github.com/aws/amazon-q-developer-cli/issues/3271

*Description of changes:*
Since the triple backtick detector results in unexpected hangs, I added a multiline hint when an unclosed triple backtick is present in the user prompt. I also made it so that the hint will not be incorrectly used for autocompletion.

### DIFF
--- a/crates/chat-cli/src/cli/chat/prompt.rs
+++ b/crates/chat-cli/src/cli/chat/prompt.rs
@@ -358,6 +358,7 @@ impl ChatHinter {
             let triple_backtick_count = line.matches("```").count();
             if triple_backtick_count % 2 == 1 {
                 // We have an odd number of ```, meaning we're in multiline mode
+                // Show status hint (right arrow key is overridden to not complete this)
                 return Some("in multiline mode, waiting for closing backticks ```".to_string());
             }
         }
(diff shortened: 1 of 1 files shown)


evanliu048 (commented):
This is expected behavior, not a bug. When the user types ``` and presses Enter, the validator correctly treats it as an unclosed code block and returns Incomplete, moving the cursor to the next line to wait for input. I don’t currently see a strong reason to remove this check

ekang7:
I think we may then need to update the documentation because right now
https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-chat-commands.html
doesn't include anything about triple backticks. We could also continue with the removal as users have other ways of doing multilines and sometimes if a user copies and pastes something with a single ```, this could result in unexpected behavior

evanliu048:
> I think we may then need to update the documentation because right now https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line-chat-commands.html doesn't include anything about `. We could also continue with the removal as users have other ways of doing multilines and sometimes if a user copies and pastes something with a single `, this could result in unexpected behavior

The documentation you pointed to covers commands, while ``` isn’t a command — it’s just a syntax check. So it’s expected that it doesn’t appear there.

That said, from a usability perspective, I agree we could improve the experience. Instead of silently waiting for the closing code fence, we might consider showing a hint when the user opens a code block — something like a clear UI indicator that  Q is now in multiline mode and waiting for the closing backticks. That would avoid confusion when users forget to close the block

ekang7:
Sounds good! I'll use the greyed out version of the hint text like in the attached screenshot to say something like "in multiline mode, waiting for closing backticks ```" when there are opening backticks but not closing backticks
Screenshot 2025-11-04 at 4 31 25 PM

ekang7:
I have successfully added the multiline hint

evanliu048:
> I have successfully added the multiline hint

I like the UX improvement! One concern though: pressing → (right arrow) currently fills in "in multiline mode, waiting for closing backticks ```" as input text. This seems unintended since it's a status hint, not a completion suggestion

ekang7:
Thank you! Good point, I made it so that the hint will not be used as part of an autocompletion suggestion any longer.

evanliu048:
The current solution uses a hint to notify users when they have an unclosed triple backtick code block. @ekang7 will continue working on making the experience smoother and more comfortable — for example, finding a way to avoid always appending the hint after user input.

image

PR #3334, opened 2025-10-31
Modify /subscribe to redirect to kiro.dev

akhileshb:
Changing /subscribe to support new usecases/
1. Social/Free BID - Redirect to kiro.dev
2. Enterprise - Keep the current flow
3. Paid BID - Keep the current flow

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/subscribe.rs
+++ b/crates/chat-cli/src/cli/chat/cli/subscribe.rs
@@ -25,7 +25,7 @@ const DEFAULT_SUBSCRIPTION_DOMAIN: &str = "app.kiro.dev";
 
 fn get_subscription_url() -> String {
     let domain = env::var("KIRO_SUBSCRIPTION_DOMAIN").unwrap_or_else(|_| DEFAULT_SUBSCRIPTION_DOMAIN.to_string());
-    format!("https://{}/account/usage", domain)
+    format!("https://{domain}/account/usage")
 }
 
 /// Arguments for the subscribe command to manage Developer Pro subscriptions
--- a/crates/chat-cli/src/constants.rs
+++ b/crates/chat-cli/src/constants.rs
@@ -172,15 +172,7 @@ pub mod ui_text {
 pub mod subscription_text {
     use super::PRODUCT_NAME;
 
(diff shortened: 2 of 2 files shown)


crates/chat-cli/src/cli/chat/cli/subscribe.rs
```diff
@@ -94,93 +70,71 @@ impl SubscribeArgs {
...
+async fn redirect_to_kiro_dev(session: &mut ChatSession) -> Result<(), ChatError> {
+    execute!(
+        session.stderr,
+        style::Print("\n"),
+        StyledText::primary_fg(),
+        style::Print("Visit kiro.dev to manage your subscription.\n\n"),
+        StyledText::reset(),
+    )?;
+
+    let url = "https://app.kiro.dev/account/usage";
```
evanliu048: Would it make sense to extract this URL as a constant and allow overriding it via an environment variable for gamma/beta testing?

akhileshb: Good point. Updated in the least revision

crates/chat-cli/src/cli/chat/cli/subscribe.rs
```diff
@@ -1,26 +1,33 @@
...
 };
 use crate::constants::subscription_text;
 use crate::os::Os;
 use crate::theme::StyledText;
 use crate::util::system_info::is_remote;
 
+const DEFAULT_SUBSCRIPTION_DOMAIN: &str = "app.kiro.dev";
+
+fn get_subscription_url() -> String {
+    let domain = env::var("KIRO_SUBSCRIPTION_DOMAIN").unwrap_or_else(|_| DEFAULT_SUBSCRIPTION_DOMAIN.to_string());
```
kkashilk, line 27: add this to the env_var util - https://github.com/aws/amazon-q-developer-cli/blob/846669fb86a940ee255ea63a2c07f059d3023410/crates/chat-cli/src/util/env_var.rs#L15

akhileshb: This is temporary env added to support testing. We will remove them early next week hence might be better to not add in env_var

PR #3335, opened 2025-10-31
feat: warn when duplicate agents are found during load

brandonskiser:
*Description of changes:*
- Add a warning for when we load duplicate agents in the global and workspace directories. Currently, we silently drop duplicate agents.

Example with duplicates in both global and workspace, and loading an overriden agent:

Screenshot 2025-10-31 at 10 29 15 AM

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -563,7 +563,7 @@ impl Agents {
             };
 
             let mut agents = Vec::<Agent>::new();
-            let results = load_agents_from_entries(files, os, &mut global_mcp_config, mcp_enabled, output).await;
+            let results = load_agents_from_entries(files, os, &mut global_mcp_config, mcp_enabled, false, output).await;
             for result in results {
                 match result {
                     Ok(agent) => agents.push(agent),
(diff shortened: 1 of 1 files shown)


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -563,7 +563,7 @@ impl Agents {
             };
 
             let mut agents = Vec::<Agent>::new();
-            let results = load_agents_from_entries(files, os, &mut global_mcp_config, mcp_enabled, output).await;
+            let results = load_agents_from_entries(files, os, &mut global_mcp_config, mcp_enabled, false, output).await;
```
abhraina-aws, line 566: Duplicate agent between workspace and global or something else?

brandonskiser: Duplicates found in either the workspace or global directories. We only warn when there is a global agent being overridden by a workspace agent, but we don't warn anything when there's more than one agent with the same name found in the workspace/global directories (meaning we silently drop agents since we dedup on agent name)

abhraina-aws: Why is it warning for the hi agent above then?

PR #3337, opened 2025-10-31
Upgrade delegate with better UX for notifications and change the file…

abhraina-aws:
Enhance the delegate UX and mechanics. 
Change the notification system. 
And remove the dependence on the agent execution file for notifications.
Make the summaries static and read them into context automatically. 

image

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/delegate.rs
+++ b/crates/chat-cli/src/cli/chat/tools/delegate.rs
@@ -12,10 +12,7 @@ use crossterm::{
     queue,
     style,
 };
-use eyre::{
-    Result,
-    bail,
-};
+use eyre::Result;
 use schemars::JsonSchema;
 use serde::{
     Deserialize,
@@ -109,7 +106,14 @@ impl Delegate {
             Operation::Status => match &self.agent {
                 Some(agent_name) => status_agent(os, agent_name).await?,
                 None => match status_all_agents(os).await {
-                    Ok(execution) => execution,
+                    Ok(executions) => {
+                        if executions.is_empty() {
+                            "No new completed delegate tasks".to_string()
+                        } else {
+                            let agent_names: Vec<String> = executions.iter().map(|e| e.agent.clone()).collect();
+                            format!("The following delegate tasks are ready: {}", agent_names.join(", "))
+                        }
+                    },
                     Err(msg) => msg.to_string(),
                 },
             },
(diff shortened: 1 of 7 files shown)


crates/chat-cli/src/cli/chat/tools/delegate.rs
```diff
@@ -416,47 +483,47 @@ pub async fn status_agent(os: &Os, agent: &str) -> Result<String> {
                 save_agent_execution(os, &execution).await?;
             }
 
-            if execution.status == AgentStatus::Completed {
-                let _ = os.fs.remove_file(path).await;
-            }
```
dingfeli, line 421: So we are saving the agent execution but we are no longer removing them?

abhraina-aws: Yes. We just overwrite. No need to delete. Only if its running we error out saying we are busy.

crates/chat-cli/src/cli/chat/tools/delegate.rs
```diff
@@ -416,47 +483,47 @@ pub async fn status_agent(os: &Os, agent: &str) -> Result<String> {
...
             Ok(execution.format_status())
         },
         None => Ok(format!("No execution found for agent '{}'", agent)),
     }
 }
 
-pub async fn status_all_agents(os: &Os) -> Result<String> {
-    // Because we would delete completed execution that has been read, everything that remains is
-    // assumed to not be stale
+pub async fn status_all_agents(os: &Os) -> Result<Vec<AgentExecution>> {
```
brandonskiser, line 492: So this function returns "only completed/failed tasks that haven't been shown to user" - I think we need a better name other than `status_all_agents`. Maybe instead `status_all_stopped_and_unnotified_agents`

brandonskiser (approved):

crates/chat-cli/src/cli/chat/tools/delegate.rs
dingfeli: Okay but this directory would just grow. Not sure how I feel about that.

crates/chat-cli/src/cli/chat/tools/delegate.rs
abhraina-aws: The function is actually checking status of all but returning only 2 thats why I think name wise its ok.

crates/chat-cli/src/cli/chat/tools/delegate.rs
abhraina-aws: Its will be equal to the number agents that we have worst case. Thats not much. 1 file per agent shouldn't be that big a deal. Plus people can just delete these files on the fly if they want to. 
We were ok with having this number.

PR #3340, opened 2025-10-31
chore: move RTS code from agent crate to chat-cli

brandonskiser:
*Description of changes:* 
- Cleaning up the API client error type and error classification logic to be shared across the cw and sigv4 API's.
- Moving over copied code from the `agent` API client crate to `chat-cli`.
- Commenting out Windows GitHub actions since we don't have any Windows machines and it's not officially supported.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -75,46 +75,46 @@ jobs:
       - name: Run tests
         run: cargo test --locked --workspace --lib --bins --test '*' --exclude fig_desktop-fuzz
 
-  cargo-clippy-windows-chat-cli:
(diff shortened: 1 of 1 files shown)


crates/agent/Cargo.toml
```diff
@@ -69,7 +69,7 @@ tempfile.workspace = true
 thiserror.workspace = true
```
dingfeli, line 69: It looks like this file is the only change in the agent crate.
Do we not also need to delete stuff from it (the stuff that you had copied over from chat-cli crate).

dingfeli (approved):

PR #3352, opened 2025-11-03
fix: make command/url conditionally required based on MCP server type

### LINKED ISSUE #3346: bug: mcpServers schema requires "command" even for HTTP-based MCP servers
### Checks

- [x] I have searched [github.com/aws/amazon-q-developer-cli/issues](https://github.com/aws/amazon-q-developer-cli/issues?q=) and there are no duplicates of my issue
- [x] I have run `q doctor` in the affected terminal session
- [x] I have run `q restart` and replicated the issue again

### Operating system

WSL

### Expected behaviour

In the agent-v1.json schema, each entry under mcpServers requires the "command" field. 
This requirement does not make sense for MCP servers that use "type": "http", since HTTP servers connect via a URL instead of a local command.

https://github.com/aws/amazon-q-developer-cli/blob/main/schemas/agent-v1.json

### Actual behaviour

Image

### Steps to reproduce

* Define an MCP server with "type": "http" and "url": "https://knowledge-mcp.global.api.aws".
* Validate it against the schema.
* Validation fails because "command" is missing.

### Environment

```yaml

```

mzkmnk:
### Summary
Fixes #3346. Improves the MCP server configuration schema to conditionally require fields based on the transport type.

### Problem
The previous schema required the command field for all MCP servers. However, http type MCP servers need a url field instead of command. This caused 
incorrect validation for HTTP-based MCP server configurations.

### Changes
• Introduced JSON Schema if-then-else syntax in schemas/agent-v1.json
• When type: "http", the url field is required
• Otherwise (for stdio), the command field is required

### Impact
• Only affects agent configuration schema validation logic
• No impact on existing valid configurations
• Invalid configurations will now be detected more accurately

#### Before

スクリーンショット 2025-11-03 午後10 43 23

#### After

スクリーンショット 2025-11-03 午後9 48 21

[Tool used](https://www.jsonschemavalidator.net/)

### DIFF
--- a/schemas/agent-v1.json
+++ b/schemas/agent-v1.json
@@ -146,9 +146,20 @@
             "default": false
           }
         },
-        "required": [
-          "command"
-        ]
+        "if": {
+          "properties": {
+            "type": {
+              "const": "http"
+            }
+          },
+          "required": ["type"]
+        },
+        "then": {
+          "required": ["url"]
+        },
+        "else": {
+          "required": ["command"]
+        }
       },
       "default": {}
     },


mzkmnk:
@mschrage 

Will this PR be merged once the qv2 project is completed?

PR #3353, opened 2025-11-03
feat(diff): Add custom diff tool support for fs_write via Q_DIFF_TOOL

avinashsivaraman:
*Description of changes:*

Add support for configuring external diff tools to view file changes in preferred editor/IDE instead of terminal.

Features:
- Q_DIFF_TOOL environment variable for tool configuration
- Support for file write approval diffs only
- Automatic fallback to inline diff when not configured

Usage:
  export Q_DIFF_TOOL="code --diff"  # VSCode
  export Q_DIFF_TOOL="vimdiff"    # vimdiff

Integration points:
- File write tool: approval flow diffs

Implementation:
- New diff_tool module with launch and content handling

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/diff_tool.rs
+++ b/crates/chat-cli/src/cli/chat/cli/diff_tool.rs
@@ -28,16 +28,11 @@ pub fn launch_diff_tool(before_path: &Path, after_path: &Path) -> Result<(), Cha
         cmd.arg(arg);
     }
 
-    // Wait for the process to complete so files aren't deleted while tool is reading them
-    let status = cmd
+    cmd
         .arg(before_path)
         .arg(after_path)
         .status()
         .map_err(|e| ChatError::Custom(format!("Failed to launch diff tool: {}", e).into()))?;
-
-    if !status.success() {
-        return Err(ChatError::Custom("Diff tool exited with non-zero status".into()));
-    }
 
     Ok(())
 }


evanliu048 (commented):
Hey @avinashsivaraman ! I really like this feature — it makes using the CLI with the IDE super convenient while coding.
One quick question: if a user accidentally closes the diff file, how can they view it again?
Should we also consider printing the diff in the terminal?
Curious to hear your thoughts!

avinashsivaraman:
> Hey @avinashsivaraman ! I really like this feature — it makes using the CLI with the IDE super convenient while coding. One quick question: if a user accidentally closes the diff file, how can they view it again? Should we also consider printing the diff in the terminal? Curious to hear your thoughts!

I can think of two options here 
1) They can ask the Q to generate the diff again. 
2) We provide the option to show diff again while getting the approval from user.  Today, we provide trust, yes & no. Probably we can add diff to show the diff again.

evanliu048:
> I can think of two options here
> 
> 1. They can ask the Q to generate the diff again.
> 2. We provide the option to show diff again while getting the approval from user.  Today, we provide trust, yes & no. Probably we can add diff to show the diff again.

Or we can just keep printing the diff in the terminal, so we don’t waste another model call.

avinashsivaraman:
If the external diff tool is cli tool like [`delta`](https://github.com/dandavison/delta) which prints the diff inline, then we will be showing two diffs all the time. That's why I don't us to print two diff all the time.

DIFF_TOOL can be two types. 
1. Opens the diff in separate process like ide, vimdiff etc.
2. prints the diff inline like `delta`.

The proposed option 2 doesn't do another model call. It will re-invoke the approval request once again.

PR #3354, opened 2025-11-04
feat(agent crate): adds remote mcp support in agent crate

dingfeli:
*Issue #, if available:*

*Description of changes:*
- Adds support to http transport 
- Slight architectural changes where the instantiation of `McpServerActor` is given a `mpsc::Sender<McpServerActorEvent>` as opposed to each `McpServerActor` having to spawn their own pair. This is because this allows us to poll a single `mpsc::Receiver<McpServerActorEvent>`, as compared to before, where we have to call `recv` on all of the `McpServerActor` and putting them all a collection `impl Future`. 

Changes since review 1:
- adds method to see if mcp config is disabled
- changes mpsc to broadcast for mcp_main_loop_to_handle_server_event_rx 
- expands path using provided util function
- uses event buffer for mcp events
- relocates server event sender to main loop
- restores use of oneshot in server launch 
- propagates appropriate messages to prevent leaky abstractions

*Tests*
Test case: cached credentials, making tool calls (this was ran before the decoupling of Model from the agent crate):
```
➜  amazon-q-developer-cli jj:(dingfeli/adds-http-support-agent-crate) ✗ RUST_LOG=trace cargo run -p agent -- run --agent=test_test --dangerously-trust-all-tools "what notion docs do i have"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.13s
     Running `target/debug/agent run --agent=test_test --dangerously-trust-all-tools 'what notion docs do i have'`
I'll search your Notion workspace to see what documents you have available.
{
  "toolUseId": "tooluse_e6OugG3yTeWBD4ZDc_peSw",
  "name": "notion-search",
  "input": {
    "__tool_use_purpose": "Search the user's Notion workspace to find all available documents",
    "query": "*",
    "query_type": "internal"
  }
}
Based on the search results, here are the Notion documents you have:

**Task Management:**
- Weekly To-do List
- My tasks

**Getting Started:**
- Welcome to Notion! (onboarding page)

**Other:**
- Home views

You have a mix of creative content, financial tracking documents, and productivity tools. Most were created about a month ago, with the Steam Deck poem being more recent (11 days ago).%

```

### DIFF
--- a/crates/agent/src/agent/mcp/mod.rs
+++ b/crates/agent/src/agent/mcp/mod.rs
@@ -110,6 +110,7 @@ pub mod types;
 
 use std::collections::HashMap;
 use std::path::PathBuf;
+use std::time::Duration;
 
 use actor::{
     McpServerActor,
@@ -254,8 +255,11 @@ impl McpManagerHandle {
         }
     }
 
-    pub async fn recv(&mut self) -> Result<McpServerActorEvent, RecvError> {
-        self.mcp_main_loop_to_handle_server_event_rx.recv().await
+    pub async fn recv(&mut self) -> Result<McpServerEvent, RecvError> {
+        self.mcp_main_loop_to_handle_server_event_rx
+            .recv()
+            .await
+            .map(|evt| evt.into())
     }
 }
 
@@ -463,4 +467,50 @@ pub enum McpManagerError {
     Channel,
     #[error("{}", .0)]
     Custom(String),
+}
+
+/// MCP events relevant to agent operations.
+/// Provides abstraction over [McpServerActorEvent] to avoid leaking implementation details.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum McpServerEvent {
+    /// The MCP server has launched successfully
+    Initialized {
+        server_name: String,
+        /// Time taken to launch the server
+        serve_duration: Duration,
+        /// Time taken to list all tools.
+        ///
+        /// None if the server does not support tools, or there was an error fetching tools.
+        list_tools_duration: Option<Duration>,
+        /// Time taken to list all prompts
+        ///
+        /// None if the server does not support prompts, or there was an error fetching prompts.
+        list_prompts_duration: Option<Duration>,
+    },
+    /// The MCP server failed to initialize successfully
+    InitializeError { server_name: String, error: String },
+    /// An OAuth authentication request from the MCP server
+    OauthRequest { server_name: String, oauth_url: String },
+}
+
+impl From<McpServerActorEvent> for McpServerEvent {
+    fn from(value: McpServerActorEvent) -> Self {
+        match value {
+            McpServerActorEvent::Initialized {
+                server_name,
+                serve_duration,
+                list_tools_duration,
+                list_prompts_duration,
+            } => Self::Initialized {
+                server_name,
+                serve_duration,
+                list_tools_duration,
+                list_prompts_duration,
+            },
+            McpServerActorEvent::InitializeError { server_name, error } => Self::InitializeError { server_name, error },
+            McpServerActorEvent::OauthRequest { server_name, oauth_url } => {
+                Self::OauthRequest { server_name, oauth_url }
+            },
+        }
+    }
 }
--- a/crates/agent/src/agent/mod.rs
+++ b/crates/agent/src/agent/mod.rs
@@ -58,7 +58,7 @@ use agent_loop::{
 use chrono::Utc;
 use consts::MAX_RESOURCE_FILE_LENGTH;
 use futures::stream::FuturesUnordered;
-use mcp::actor::McpServerActorEvent;
+use mcp::McpServerEvent;
 use permissions::evaluate_tool_permission;
 use protocol::{
     AgentError,
@@ -494,7 +494,7 @@ impl Agent {
                 evt = self.mcp_manager_handle.recv() => {
                     match evt {
                         Ok(evt) => {
-                            self.handle_mcp_server_actor_events(evt).await;
+                            self.handle_mcp_events(evt).await;
                         },
                         Err(e) => {
                             error!(?e, "mcp manager handle closed");
@@ -1698,11 +1698,9 @@ impl Agent {
         Ok(())
     }
 
-    async fn handle_mcp_server_actor_events(&self, evt: McpServerActorEvent) {
+    async fn handle_mcp_events(&mut self, evt: McpServerEvent) {
         let converted_evt = AgentEvent::Mcp(evt);
-        if let Err(e) = self.agent_event_tx.send(converted_evt) {
-            error!(?e, "failed to emit agent event");
-        }
+        self.agent_event_buf.push(converted_evt);
     }
 }
 
--- a/crates/agent/src/agent/protocol.rs
+++ b/crates/agent/src/agent/protocol.rs
@@ -17,9 +17,11 @@ use super::agent_loop::types::{
     ImageBlock,
     ToolUseBlock,
 };
-use super::mcp::McpManagerError;
-use super::mcp::actor::McpServerActorEvent;
 use super::mcp::types::Prompt;
+use super::mcp::{
+    McpManagerError,
+    McpServerEvent,
+};
 use super::task_executor::TaskExecutorEvent;
 use super::tools::{
     Tool,
@@ -76,7 +78,7 @@ pub enum AgentEvent {
     Internal(InternalEvent),
 
     /// Events from MCP (Model Context Protocol) servers
-    Mcp(McpServerActorEvent),
+    Mcp(McpServerEvent),
 }
 
 impl From<TaskExecutorEvent> for AgentEvent {


crates/agent/src/agent/mod.rs
```diff
@@ -346,72 +344,59 @@ impl Agent {
...
-            let mut results = FuturesUnordered::new();
-            for config in &self.cached_mcp_configs.configs {
-                let Ok(rx) = self
+            let mut total_servers_to_be_loaded = 0_usize;
+
+            for config in self
+                .cached_mcp_configs
+                .configs
+                .iter()
+                .filter(|config| match &config.config {
```
brandonskiser: nit - add a `McpServerConfig::disabled` method instead

crates/agent/src/agent/mod.rs
```diff
@@ -346,72 +344,59 @@ impl Agent {
...
+                            error!("mcp manager handle channel closed");
                             break;
                         };
-                        debug!(?name, "MCP server successfully initialized");
-                        launched_servers.push(name);
-                    },
-                    name = failed_rx.recv() => {
-                        let Some(name) = name else {
+
+                        if matches!(evt, McpServerActorEvent::Initialized{ .. } | McpServerActorEvent::InitializeError { .. }) {
```
brandonskiser: Is there a reason why you removed the `rx` for handling the result? With this paradigm, we now have to call `recv` on the handle in order to get the result, meaning we're going to be inter-mixing unrelated MCP events with the result we actually want (ie, in this case launching a server).


Just thinking about launching new servers at runtime post-initialization, seems more intuitive to reason about having an `rx` that the client would simply call `recv` on to get the result.

crates/agent/src/agent/mod.rs
```diff
@@ -346,72 +344,59 @@ impl Agent {
...
+                             total_servers_to_be_loaded = total_servers_to_be_loaded.saturating_sub(1);
+                        }
+                        self.handle_mcp_server_actor_events(evt).await;
+
+                        if total_servers_to_be_loaded == 0 {
+                            info!("all mcp servers loaded before timeout");
```
brandonskiser: We're losing info about what servers launched successfully versus failed, would highly prefer to keep this available.

crates/agent/src/agent/mod.rs
```diff
@@ -1679,6 +1672,15 @@ impl Agent {
         self.set_active_state(ActiveState::ExecutingRequest).await;
         Ok(())
     }
+
+    async fn handle_mcp_server_actor_events(&self, evt: McpServerActorEvent) {
+        tracing::info!(?evt, "received mcp actor event");
```
brandonskiser: minor nit - just `info!`, and this method name shouldn't be plural

crates/agent/src/agent/protocol.rs
```diff
@@ -73,6 +74,9 @@ pub enum AgentEvent {
     /// Lower-level events associated with the agent's execution. Generally only useful for
     /// debugging or telemetry purposes.
     Internal(InternalEvent),
+
+    /// Events from MCP (Model Context Protocol) servers
+    Mcp(McpServerActorEvent),
```
brandonskiser: I'd wrap this in a dedicated `MCP` event that populates context about the actor event (namely the server name), actor events are moreso the concern of just the MCP manager itself

crates/agent/src/agent/mcp/mod.rs
```diff
@@ -40,39 +151,46 @@ use crate::agent::util::request_channel::{
 #[derive(Debug, Clone)]
 pub struct McpManagerHandle {
     /// Sender for sending requests to the tool manager task
-    sender: RequestSender<McpManagerRequest, McpManagerResponse, McpManagerError>,
+    request_tx: RequestSender<McpManagerRequest, McpManagerResponse, McpManagerError>,
+    server_to_handle_server_event_tx: mpsc::Sender<McpServerActorEvent>,
+    mcp_main_loop_to_handle_server_event_rx: Arc<Mutex<mpsc::Receiver<McpServerActorEvent>>>,
```
brandonskiser: Use a `broadcast::Receiver` instead, my intention behind handles being `Clone` is that multiple consumers need to consume all emitted events.

crates/agent/src/agent/mcp/mod.rs
```diff
@@ -40,39 +151,46 @@ use crate::agent::util::request_channel::{
 #[derive(Debug, Clone)]
 pub struct McpManagerHandle {
     /// Sender for sending requests to the tool manager task
-    sender: RequestSender<McpManagerRequest, McpManagerResponse, McpManagerError>,
+    request_tx: RequestSender<McpManagerRequest, McpManagerResponse, McpManagerError>,
+    server_to_handle_server_event_tx: mpsc::Sender<McpServerActorEvent>,
```
brandonskiser: What is the intention behind `server_to_handle_server_event_tx`? The handle should theoretically just be a `tx` to send requests, and a receiver for consuming events. It seems like this is purely to get the result of a `LaunchServer`? Would highly prefer to remove this if that's the case

Also - would prefer not to expose actor events as the API behind the MCP manager, should wrap these in manager events. I learned that the hard way by trying to use `AgentLoopEvent` as the API behind an `Agent`, it's a leaky abstraction that makes updating internals more difficult.

crates/agent/src/agent/mcp/mod.rs
```diff
@@ -238,44 +355,46 @@ impl McpManager {
...
+                self.initializing_servers.remove(server_name);
             },
-            McpServerActorEvent::InitializeError(msg) => {
-                let _ = tx.send(Err(McpManagerError::Custom(msg)));
-                self.initializing_servers.remove(&server_name);
+            McpServerActorEvent::OauthRequest { server_name, oauth_url } => {
+                tracing::info!(?server_name, ?oauth_url, "received oauth request");
             },
         }
+        let _ = mcp_main_loop_to_handle_server_event_tx.send(evt).await;
```
brandonskiser: When it comes to emitting events, would prefer to follow the same paradigm currently in place with `Agent` - we keep a field `mcp_event_buf: Vec<McpEvent>` that we push to in handlers, and at the beginning of every loop in `main_loop` we drain the buf sending all of the events.

It helps with decoupling updating internal state from emitting events that clients consume - essentially, we want to finish updating all of our internal state when handling an event before emitting anything to any consumers.

crates/agent/src/agent/mcp/mod.rs
```diff
@@ -238,44 +355,46 @@ impl McpManager {
...
     }
 }
 
 impl Default for McpManager {
     fn default() -> Self {
-        Self::new()
+        let expanded_path =
```
brandonskiser, line 415: For handling paths, I have a sort of testing system in place (see `src/agent/util/path.rs` `expand_path`). Essentially, we want to include a `SystemProvider` for `McpManager` that states what env, cwd, etc. is in place. This is critical for running integ tests in temp directories

crates/agent/src/agent/mcp/mod.rs
dingfeli: > What is the intention behind server_to_handle_server_event_tx?

It's to equip McpServerActor a sender to emit events. It goes beyond just servers being launched. It's to enable all server driven events to be emitted. These are things like oauth request and sampling (where the server is asking the agent to make an llm call. Though this is a feature we currently do not support). 

> Also - would prefer not to expose actor events as the API behind the MCP manager, should wrap these in manager events

Yea that makes sense. I'll wait until I get all the feedback for the first draft to make this change along with all the other changes.

crates/agent/src/agent/mod.rs
dingfeli: > Is there a reason why you removed the rx for handling the result?

I removed it because with the `recv` method on the handle would already have a way to gain insights to the server loading events. Fundamentally I think the server launch status is not in a different categories as other Mcp related events. 

> With this paradigm, we now have to call recv on the handle in order to get the result, meaning we're going to be inter-mixing unrelated MCP events with the result we actually want (ie, in this case launching a server).

I do see your point here. I don't think calling `recv` on the handle is inherently bad. But I do think perhaps putting the onus on the agent layer to filter out relevant info that it wants from all the other mcp related events is a leaky abstraction that we wouldn't want to commit ourselves to. 

There is also the issue of prematurely consuming events that are not relevant at the time, thus starving the relevant actor (which may or may not have been spawned). But I think that would be resolved by changing the channel to a `broadcast` like you had suggested in a [previous comment](https://github.com/aws/amazon-q-developer-cli/pull/3354/files#r2495674793). 

> Just thinking about launching new servers at runtime post-initialization, seems more intuitive to reason about having an rx that the client would simply call recv on to get the result.

Not sure if I am on the same page with this one. I think it would be reasonable to assume that post init the relevant actors would already be in place, polling on the their respective receivers. We should just utilize these established channel of communication to deliver the launch status. Otherwise we would be asking the consumer of these events to create another async scope to wait on these results in order to not block on waiting for them. 

Though we could always take a hybrid approach and preserve these oneshot channels. When a server is launched we would be delivering this event via both channels. That way we leave it up to the consumer of these events how they want to do this. It does mean that they would need to be doing the deduping of these events on their own if they already have an actor set up to poll on recv of `McpManagerHandle`.

crates/agent/src/agent/mod.rs
dingfeli: Sounds good. I'll restore this.

crates/agent/src/agent/protocol.rs
dingfeli: That makes sense. Though right now the events of interest to an agent consumer have a large overlap with what `McpServerActorEvent` currently covers: 

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub enum McpServerActorEvent {
    /// The MCP server has launched successfully
    Initialized {
        server_name: String,
        /// Time taken to launch the server
        serve_duration: Duration,
        /// Time taken to list all tools.
        ///
        /// None if the server does not support tools, or there was an error fetching tools.
        list_tools_duration: Option<Duration>,
        /// Time taken to list all prompts
        ///
        /// None if the server does not support prompts, or there was an error fetching prompts.
        list_prompts_duration: Option<Duration>,
    },
    /// The MCP server failed to initialize successfully
    InitializeError { server_name: String, error: String },
    /// An OAuth authentication request from the MCP server
    OauthRequest { server_name: String, oauth_url: String },
}
```

crates/agent/src/agent/mod.rs
brandonskiser: > Though we could always take a hybrid approach and preserve these oneshot channels

This is the thought behind requests that return an `rx` - we would emit two events, one to the rx channel, and one to the "global" handle channel.

Still not sure if we can just not have the `rx` here - consider now during initialization, we could be inter-mixing OAuth events during server initialization. Would highly prefer to keep the `rx` here at this point in time I think unless I'm missing something.

PR #3359, opened 2025-11-04
feat: Add automatic migration from amazon-q to kiro-cli

kkashilk:
*Issue #, if available:*
Provide a migration strategy for settings.json and the database

*Description of changes:*

- Migrate database from {data_local_dir}/amazon-q to {data_local_dir}/kiro-cli
- Migrate settings from {data_local_dir}/amazon-q to ~/.aws/kiro-cli
- Transform api.q.service → api.kiro.service in settings
- Add migration.kiro.completed flag to track completion
- Implement file locking with rustix to prevent race conditions
- Run migration automatically on CLI startup (silent, no user interaction)
- Add hidden 'migrate' subcommand for manual migration
- Update paths.rs to use new locations with fallback to old

### DIFF
--- a/crates/chat-cli/src/util/paths.rs
+++ b/crates/chat-cli/src/util/paths.rs
@@ -11,6 +11,11 @@ use globset::{
     GlobSetBuilder,
 };
 use thiserror::Error;
+use tracing::{
+    debug,
+    info,
+    warn,
+};
 
 use crate::os::Os;
 
@@ -43,12 +48,9 @@ pub enum DirectoryError {
 
 pub mod workspace {
     //! Project-level paths (relative to current working directory)
-    pub const AGENTS_DIR: &str = ".amazonq/cli-agents";
-    pub const PROMPTS_DIR: &str = ".amazonq/prompts";
-    pub const MCP_CONFIG: &str = ".amazonq/mcp.json";
     pub const TODO_LISTS_DIR: &str = ".amazonq/cli-todo-lists";
     pub const SUBAGENTS_DIR: &str = ".amazonq/.subagents";
-    pub const RULES_PATTERN: &str = ".amazonq/rules/**/*.md";
+    pub const RULES_PATTERN: &str = "file://{}/**/*.md";
 
     // Default documentation files for agent resources
     pub const DEFAULT_AGENT_RESOURCES: &[&str] = &[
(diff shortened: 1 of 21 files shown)


crates/chat-cli/src/util/paths.rs
```diff
@@ -309,30 +311,93 @@ impl<'a> GlobalPaths<'a> {
         Ok(dir)
     }
 
+    /// Get settings path (new location with fallback to old)
```
brandonskiser, line 408: Why do we fallback? Migrate should always run at the start, so in theory we should always have the new settings created under the kiro path.

Reasoning - if I delete my kiro settings, suddenly I still have old settings being applied from q, which is not very intuitive in my mind. Same reasoning behind the database.

crates/chat-cli/src/cli/migrate.rs
```diff
@@ -0,0 +1,284 @@
...
+    pub yes: bool,
+}
+
+impl MigrateArgs {
+    pub async fn execute(self, _os: &mut Os) -> Result<ExitCode> {
+        // Try to acquire migration lock
+        let _lock = match acquire_migration_lock()? {
+            Some(lock) => lock,
+            None => {
+                // Another process is migrating, skip silently
```
brandonskiser: What if there is an error and the lock file exists - how would the user clean it up? We should probably print that a migration is currently in progress.

We could probably check the file creation time and if it's longer than e.g. a minute, then something must've went wrong.

crates/chat-cli/src/cli/migrate.rs
```diff
@@ -0,0 +1,284 @@
...
+                new_path.display()
+            ),
+            bytes_copied: 0,
+        });
+    }
+
+    if let Some(parent) = new_path.parent() {
+        fs::create_dir_all(parent)
+            .await
+            .context("Failed to create target directory")?;
```
brandonskiser: Should print the path that failed to be created

crates/chat-cli/src/cli/migrate.rs
```diff
@@ -0,0 +1,284 @@
...
+    let mut transformations = Vec::new();
+
+    // Transform api.q.service → api.kiro.service
+    if let Some(q_service) = transformed.remove("api.q.service") {
+        transformed.insert("api.kiro.service".to_string(), q_service);
+        transformations.push("api.q.service → api.kiro.service".to_string());
+    }
+
+    // Add migration completed flag
+    transformed.insert("migration.kiro.completed".to_string(), Value::Bool(true));
```
brandonskiser, line 237: This should be in the database, not the settings file

crates/chat-cli/src/cli/migrate.rs
```diff
@@ -0,0 +1,284 @@
...
+            .context("Failed to create target directory")?;
+    }
+
+    let json = serde_json::to_string_pretty(&transformed).context("Failed to serialize settings")?;
+    fs::write(new_path, json)
+        .await
+        .context("Failed to write settings file")?;
+
+    Ok(SettingsMigrationResult {
+        message: format!("Settings migrated successfully ({} settings)", transformed.len()),
```
brandonskiser: Do we really need to print how many settings are included? Probably just print what was transformed and that's it

crates/chat-cli/src/util/paths.rs
```diff
@@ -309,30 +311,93 @@ impl<'a> GlobalPaths<'a> {
...
+
+    /// Get new settings path (kiro-cli location in home)
+    pub fn new_settings_path() -> Result<PathBuf> {
+        Ok(dirs::home_dir()
+            .ok_or(DirectoryError::NoHomeDirectory)?
+            .join(global::SETTINGS_FILE))
+    }
+
+    /// Check if migration completed by reading settings file
+    pub fn is_migration_completed_static() -> Result<bool> {
```
brandonskiser: Should be from the db - also, something like this shouldn't be in `paths.rs`

crates/chat-cli/src/cli/migrate.rs
```diff
@@ -0,0 +1,284 @@
...
+    Needed {
+        old_db: PathBuf,
+        old_settings: PathBuf,
+        new_db: PathBuf,
+        new_settings: PathBuf,
+    },
+    Completed,
+}
+
+fn detect_migration() -> Result<MigrationStatus> {
```
brandonskiser: What if the new paths exist, but migration was not completed?

crates/chat-cli/src/cli/migrate.rs
kkashilk: these won't be shown to the user, only for debugging purposes

crates/chat-cli/src/cli/migrate.rs
kkashilk: will update

crates/chat-cli/src/util/paths.rs
kkashilk: updated

PR #3360, opened 2025-11-04
Fallback behavior for Paths & Env Variables

dk19y:
Implements fall backs for paths / env variables.

### DIFF
--- a/crates/chat-cli/src/cli/agent/mod.rs
+++ b/crates/chat-cli/src/cli/agent/mod.rs
@@ -742,7 +742,7 @@ impl Agents {
 
                 // Add rules pattern using dynamic path resolution
                 if let Ok(rules_dir) = resolver.workspace().rules_dir() {
-                    let rules_pattern = format!(paths::workspace::RULES_PATTERN, rules_dir.display());
+                    let rules_pattern = paths::workspace::RULES_PATTERN.replace("{}", &rules_dir.display().to_string());
                     agent.resources.push(rules_pattern.into());
                 }
 


crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -197,7 +197,8 @@ impl Default for Agent {
             resources: {
                 let mut resources = Vec::new();
                 resources.extend(paths::workspace::DEFAULT_AGENT_RESOURCES.iter().map(|&s| s.into()));
-                resources.push(format!("file://{}", paths::workspace::RULES_PATTERN).into());
+                // Default fallback pattern - will be updated dynamically in load() function
+                resources.push("file://.amazonq/rules/**/*.md".into());
```
dk19y: Should this remain in a constant ? - RULES_PATTERN ?

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -740,6 +741,20 @@ impl Agents {
 
             all_agents.push({
                 let mut agent = Agent::default();
+
+                // Update rules pattern to use dynamic path resolution
+                if let Ok(rules_dir) = resolver.workspace().rules_dir() {
+                    let rules_pattern = format!("{}/**/*.md", rules_dir.display());
+                    // Replace the hardcoded rules pattern with the dynamic one
```
dk19y: Move rules pattern to a var

crates/chat-cli/src/cli/agent/mod.rs
```diff
@@ -740,6 +741,20 @@ impl Agents {
...
+                    if let Some(pos) = agent
+                        .resources
+                        .iter()
+                        .position(|r| r.as_str().contains(".amazonq/rules/"))
+                    {
+                        agent.resources[pos] = format!("file://{rules_pattern}").into();
+                    }
```
dk19y: These needs to be constants / Should be addressed by migration logic ?

kensave (approved):

crates/agent/src/agent/util/directories.rs
```diff
@@ -57,28 +56,50 @@ pub fn settings_schema_path(base: impl AsRef<Path>) -> PathBuf {
     base.as_ref().join("settings_schema.json")
 }
 
+fn resolve_migrated_path(is_global: bool, subpath: &str) -> Result<PathBuf> {
+    let (kiro_base, amazonq_base) = if is_global {
+        let home = home_dir()?;
```
kensave, line 61: `nit`: It seems this is actually two methods, `resolve_global_migrated_path` and `resolve_local_migrated_path` for the public facing methods. Using boolean is error prone and not very self explanatory.

crates/agent/src/agent/agent_config/mod.rs
brandonskiser, line 1: Any changes under `crates/agent` should be removed, this is dead code

crates/chat-cli/src/util/paths.rs
```diff
@@ -61,18 +64,124 @@ pub mod workspace {
...
+            warn!(
+                "Migration notice: Using .amazonq {} configs at: {}",
+                scope,
+                amazonq_base.display()
+            );
+            amazonq_base.join(subpath)
+        },
+        (true, true) => {
+            warn!(
+                "Config conflict: Both .amazonq and .kiro {} configs exist, using .amazonq at: {}",
```
brandonskiser: We should always prioritize using kiro if both exist, no?

crates/chat-cli/src/util/paths.rs
```diff
@@ -61,18 +64,124 @@ pub mod workspace {
...
+struct RealFileSystem;
+
+impl FileSystemChecker for RealFileSystem {
+    fn exists(&self, path: &std::path::Path) -> bool {
+        path.exists()
+    }
+}
+
+/// Test filesystem implementation
+#[cfg(test)]
```
brandonskiser: I would just move this test code under the `mod tests`

crates/chat-cli/src/cli/agent/mod.rs
brandonskiser: What is this logic doing exactly? By default we use amazonq, but it updates to kiro using this resolver?

I would say the `impl Default` should only contain the kiro rules, and as part of loading all agents and providing the built-in default, we can include `.amazonq` rules if it exists in the current dir

crates/chat-cli/src/util/paths.rs
dk19y: Fixed

PR #3333, opened 2025-10-31
Enhancing /usage to support credits information

akhileshb:
Adding two subcommands to expand support for context and credits. By default /usage would now show both of these.

Screenshot 2025-10-31 at 9 23 40 AM

PR #3374, opened 2025-11-05
fix: Split `/usage` and `/context` commands for better UX

evanliu048:
*Issue #, if available:*

*Description of changes:*
Based on https://github.com/aws/amazon-q-developer-cli/pull/3333

`/usage` command:
Now shows only billing and credits information
Removed --context and --credits flags (no longer needed)

`/context` command:
Without subcommands: displays context window usage (tokens breakdown, dropped files, etc.)
With subcommands: retains all existing functionality (show, add, remove, clear)

Before:
image

After:
Screenshot 2025-11-05 at 2 31 58 PM


crates/chat-cli/src/cli/chat/cli/usage/usage_data_provider.rs
```diff
@@ -19,7 +19,7 @@ use crate::cli::chat::{
 use crate::os::Os;
 
 /// Get detailed usage data for context window analysis
-pub(super) async fn get_detailed_usage_data(
+pub async fn get_detailed_usage_data(
```
kkashilk: nit: move this out of usage_data_provider if we no longer consider this `usage`

kkashilk (approved):
one comment about refactoring, otherwise LGTM

crates/chat-cli/src/cli/chat/cli/usage/usage_data_provider.rs
evanliu048: Good catch! I've refactored this into a separate `context_data_provider.rs` module under `cli/context/` since it's really about context window analysis rather than billing usage. The `usage` module now only handles billing/credits data.

PR #3376, opened 2025-11-05
chore: separates system prompts for subagent

dingfeli:
*Issue #, if available:*

*Description of changes:*
As titled. 

This change is done to facilitate the iteration on the delegate feature (aka subagent) where separate system prompts are necessitated. The system prompt for a main agent is to be different than that of the subagent. Because of that, a portion of the system prompt is to be inserted from the client. The portion that is applicable to both main agent and sub agent shall remain on the backend. 


dingfeli:
NOTE: DO NOT MERGE THIS JUST YET. LET ME DO IT. 
The applicability of this change depends on APPSEC approval.

PR #3382, opened 2025-11-06
Show credit usage upon user turn completion

akhileshb:
We want to show credit consumed after the end of every user turn. 

Note: Metering events are still not being emitted for Kiro CLI hence switching back to AI_EDITOR for now

Tested this across various scenarios
1. Single request in a turn
2. Multiple requests in a turn

Here is the UX
Screenshot 2025-11-06 at 1 31 25 AM


crates/chat-cli/src/api_client/mod.rs
```diff
@@ -365,7 +365,7 @@ impl ApiClient {
         let request = self
             .client
             .list_available_models()
-            .set_origin(Some(Origin::KiroCli))
+            .set_origin(Some(Origin::AiEditor))
```
kensave, line 368: Why changing back? This should be KiroCli

akhileshb: As discussed offline since RTS changes are pending for Kiro CLI to support metering. Changing it back to testing the credits feature

PR #3414, opened 2025-11-09
Fix ANSI Escape Sequence Corruption in Clear Command

mo-vic:
## Problem
The `/clear` command suffered from a race condition where user input could start before the confirmation prompt was fully rendered. This caused corrupted ANSI escape sequences to appear in the terminal output.

## Root Cause
The `execute!` macro flushes display to stderr, but user input reading (`read_user_input`) could start before the flush completed, causing:
- Interleaved ANSI escape sequences 
- Malformed terminal output
- Poor user experience

## Solution
Implemented a **bandaid synchronization fix** using the existing conduit system (similar to the fix already present in `mod.rs`):

1. **Send MetaEvent**: After `execute!()`, send timing event to signal display completion
2. **Wait for Acknowledgment**: UI layer processes the event and sends acknowledgment  
3. **Safe Input Reading**: Only start user input after confirmation display is complete

### Development Workflow Compliance
✅ **All required checks passed:**
- `cargo clippy` - No linting warnings for `clear.rs`
- `cargo +nightly fmt` - Code properly formatted
- `typos` - No typos found

## Testing
- `cargo test` - All tests pass
- manual testing confirms clean prompt display

## Screenshot Comparison
screenshot

**Left Panel (Before)**: Shows corrupted ANSI sequences appearing in the prompt due to race condition between display rendering and user input.

**Right Panel (After)**: Clean, properly formatted confirmation prompt with complete styling and clear text layout.

## Files Changed
- `crates/chat-cli/src/cli/chat/cli/clear.rs`: Added bandaid synchronization logic with detailed comments explaining the temporary fix

### DIFF
--- a/crates/chat-cli/src/cli/chat/cli/clear.rs
+++ b/crates/chat-cli/src/cli/chat/cli/clear.rs
@@ -1,3 +1,4 @@
+use chat_cli_ui::protocol::Event;
 use clap::Args;
 use crossterm::style::{
     self,
@@ -7,6 +8,7 @@ use crossterm::{
     cursor,
     execute,
 };
+use tracing::error;
 
 use crate::cli::chat::{
     ChatError,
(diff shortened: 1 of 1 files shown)


dingfeli:
Thanks for doing this!

PR #3716, opened 2026-04-04
fix(chat): prevent panic on UTF-8 boundary when rendering response

### LINKED ISSUE #3715: panic: byte index out of bounds in chat response rendering when response contains triple backticks
## Bug Report

**Version:** `kiro-cli 1.25.1`
**OS:** Debian GNU/Linux 12 (bookworm), kernel 6.1.0-44-amd64, x86_64
**Install method:** Official install script from kiro.dev (`curl`)
**Binary location:** `/home/dev/.local/bin/kiro-cli`

## Description

The application panics with a `byte index out of bounds` error when the AI response contains triple backticks (```````) in the rendered output.

## Steps to Reproduce

1. Start a chat session: `kiro-cli chat`
2. Ask a question that causes the AI to respond with a message containing triple backticks (e.g., a code block or a message referencing backtick syntax)
3. The application crashes mid-response

## Panic Output

```
The application panicked (crashed).
Message:  byte index 536 is out of bounds of `...`[...]
Location: crates/chat-cli/src/cli/chat/mod.rs:3849
```

## Root Cause Analysis

In `crates/chat-cli/src/cli/chat/mod.rs`, the response rendering loop uses a byte `offset` to track parsing progress:

```rust
let input = Partial::new(&buf[offset..]);
// ...
offset += parsed.offset_from(&input);
```

`parsed.offset_from(&input)` returns a byte count relative to the `input` slice. This is accumulated into `offset`, which is used as a byte index into `buf`. When `buf` contains multi-byte UTF-8 characters (e.g., non-ASCII text, emoji, or characters adjacent to triple backticks), the accumulated `offset` can land in the middle of a multi-byte character boundary, causing the panic on the next iteration when `&buf[offset..]` is evaluated.

## Suggested Fix

Validate that `offset` falls on a valid UTF-8 character boundary before slicing, or use `buf.get(offset..)` which returns `None` instead of panicking:

```rust
// Instead of:
let input = Partial::new(&buf[offset..]);

// Use:
let Some(slice) = buf.get(offset..) else { break };
let input = Partial::new(slice);
```

Alternatively, ensure `offset` is always advanced using `ceil_char_boundary` or equivalent to stay on valid UTF-8 boundaries.

## Additional Context

This panic is reproducible when the AI response includes triple backticks in its text content (not just code blocks), particularly when the surrounding text contains non-ASCII characters.

sandikodev:
## Issue

Closes #3715

## Problem

When the AI response contains multi-byte UTF-8 characters (e.g. non-ASCII text adjacent to triple backticks), the byte `offset` accumulated via `parsed.offset_from()` can land in the middle of a character boundary. The subsequent `&buf[offset..]` slice then panics at runtime:

```
The application panicked (crashed).
Message:  byte index 536 is out of bounds of `...`[...]
Location: crates/chat-cli/src/cli/chat/mod.rs:3849
```

The same pattern existed in **two places**:
- `crates/chat-cli/src/cli/chat/mod.rs` — streaming response rendering loop
- `crates/chat-cli/src/cli/chat/parse.rs` — `validate!` test macro loop (line 675)

## Fix

Replace the direct slice with `.get(offset..)` which returns `None` instead of panicking when the index is out of bounds or misaligned, and break the loop gracefully.

```rust
// Before (both locations)
let input = Partial::new(&buf[offset..]);

// After
let Some(slice) = buf.get(offset..) else { break };
let input = Partial::new(slice);
```

## Testing

```bash
# Run the regression test
cargo test -p chat_cli multibyte_utf8

# Run the full test suite to check for regressions
cargo test -p chat_cli
```

The regression test covers three real-world inputs that triggered the panic:
- Indonesian text adjacent to triple backticks
- Chinese characters adjacent to triple backticks
- Emoji adjacent to triple backticks

**Before this fix:** the test panics with `byte index N is out of bounds`
**After this fix:** all inputs parse without panic

### DIFF
--- a/crates/chat-cli/src/cli/chat/mod.rs
+++ b/crates/chat-cli/src/cli/chat/mod.rs
@@ -3099,7 +3099,11 @@ impl ChatSession {
 
             // Print the response for normal cases
             loop {
-                let input = Partial::new(&buf[offset..]);
+                // Use `get` to avoid panicking if `offset` lands on a non-UTF-8 boundary,
+                // which can happen when the response contains multi-byte characters (e.g.
+                // non-ASCII text adjacent to triple backticks). See: #3715
+                let Some(slice) = buf.get(offset..) else { break };
+                let input = Partial::new(slice);
                 if self.stdout.should_send_structured_event {
                     match interpret_markdown(input, &mut temp_buf, &mut state) {
                         Ok(parsed) => {
--- a/crates/chat-cli/src/cli/chat/parse.rs
+++ b/crates/chat-cli/src/cli/chat/parse.rs
@@ -672,7 +672,8 @@ mod tests {
                 let mut offset = 0;
 
                 loop {
-                    let input = Partial::new(&input[offset..]);
+                    let Some(slice) = input.get(offset..) else { break };
+                    let input = Partial::new(slice);
                     match interpret_markdown(input, &mut presult, &mut state) {
                         Ok(parsed) => {
                             offset += parsed.offset_from(&input);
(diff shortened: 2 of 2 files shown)


sandikodev:
**Update (amended commit):** The fix was originally applied only to `mod.rs`. On closer inspection, the identical `&buf[offset..]` pattern existed in `parse.rs` as well (inside the `validate!` test macro loop at line 675). Both locations are now fixed.

A regression test was also added in `parse.rs` with real-world inputs that triggered the panic in the wild — Indonesian text, Chinese characters, and emoji adjacent to triple backticks — to confirm the fix prevents the panic and catch any future regression.

PR #3717, opened 2026-04-04
fix(fs_write): suppress spurious trailing-newline diff lines

### LINKED ISSUE #923: bug: diffs show trailing newline differences
> There's an annoying bug where Q chat shows diffs for files it modifies, but the last line of the diff always has (what i'm guessing is) a newline difference, so it clutters the context, and also I have to scroll up to actually see what the real change was.

Example:
```
 73, 73:     local role_name="$1"
  74, 74:     
  75, 75:     # Validate environment
  76, 76:     if ! validate_environment; then
  77, 77:         echo "Error: Invalid environment configuration" >&2
  78, 78:         return 1
  79, 79:     fi
  80, 80:     
  81, 81:     # Export role-specific variables
  82, 82:     export QDT_ROLE="$role_name"
  83, 83:     export QDT_ROLE_SCRIPT_DIR="$QDT_SCRIPT_DIR"
  84, 84:     
  85, 85:     return 0
  86, 86: }
  87, 87: 
  88, 88: # Main initialization
  89, 89: if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
  90, 90:     echo "This script should be sourced, not executed directly" >&2
  91, 91:     exit 1
- 92    : fi
+     92: fi
```

sandikodev:
## Issue

Closes #923

## Problem

When `fs_write` modifies a file, the diff shown to the user always has a spurious last line:

```diff
- 92    : fi
+     92: fi
```

The content is identical — the only difference is a trailing newline. This is a cosmetic artefact caused by `write_to_file` always appending `\n` to file content, while the diff compares the raw strings before normalization.

## Fix

Strip at most **one** trailing newline from both sides before passing to `TextDiff::from_lines`:

```rust
// Before
let diff = similar::TextDiff::from_lines(&old_str.content, &new_str.content);

// After — strip_suffix removes at most one \n, preserving intentional multi-newline differences
let old_content = old_str.content.strip_suffix('\n').unwrap_or(&old_str.content);
let new_content = new_str.content.strip_suffix('\n').unwrap_or(&new_str.content);
let diff = similar::TextDiff::from_lines(old_content, new_content);
```

`strip_suffix` (not `trim_end_matches`) is used deliberately — it removes exactly one trailing newline so that intentional multi-newline differences (e.g. old has 1 trailing newline, new has 3) are still shown.

## Testing

```bash
# Run the three targeted tests
cargo test -p chat_cli test_print_diff

# Run the full test suite to check for regressions
cargo test -p chat_cli
```

Three tests cover all scenarios:
1. `test_print_diff_trailing_newline_only_change_is_not_shown` — the bug case: trailing-newline-only diff produces zero deletions and zero insertions
2. `test_print_diff_real_change_is_shown` — regression guard: a genuine content change still appears (uses `insta::assert_snapshot!` for the exact formatted output)
3. `test_print_diff_multiple_trailing_newlines_difference_is_shown` — edge case: intentional extra newlines are still shown as insertions

Tests use `strip_ansi_escapes` (already in the workspace) to strip terminal color codes before asserting.

### DIFF
--- a/crates/chat-cli/src/cli/chat/tools/fs_write.rs
+++ b/crates/chat-cli/src/cli/chat/tools/fs_write.rs
@@ -641,7 +641,14 @@ fn print_diff(
     new_str: &StylizedFile,
     start_line: usize,
 ) -> Result<()> {
-    let diff = similar::TextDiff::from_lines(&old_str.content, &new_str.content);
+    // Strip at most one trailing newline from both sides before diffing. This prevents a
+    // spurious Delete/Insert pair on the last line when the only difference is whether the
+    // file ends with a newline — a cosmetic artefact caused by write_to_file always appending
+    // one. We strip only a single '\n' so that intentional multi-newline differences are still
+    // shown. See: #923
+    let old_content = old_str.content.strip_suffix('\n').unwrap_or(&old_str.content);
+    let new_content = new_str.content.strip_suffix('\n').unwrap_or(&new_str.content);
+    let diff = similar::TextDiff::from_lines(old_content, new_content);
 
     // First, get the gutter width required for both the old and new lines.
     let (mut max_old_i, mut max_new_i) = (1, 1);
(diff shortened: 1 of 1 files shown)


sandikodev:
**Update (amended commit):** The fix was originally using `trim_end_matches('\n')` which strips *all* trailing newlines, silently hiding intentional multi-newline differences. Changed to `strip_suffix('\n')` which removes at most one — enough to eliminate the `write_to_file` artefact without swallowing real content changes.

Three tests were added covering all scenarios:
1. Trailing-newline-only difference → not shown (the bug case)
2. Real content change → still shown (regression guard)
3. Multiple trailing newlines difference → still shown (edge case guard)

Tests use `strip_ansi_escapes` (already in the workspace) to strip terminal color codes before asserting, and `insta::assert_snapshot!` for the formatted diff output.

